PR #149, opened 2015-02-18
Make the framed page look like it is on a smaller page

fzzzy:
This is to reinforce the fact that we are looking at the page from
pageshot, not the original site.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@ Profile
 .DS_Store
 pageshot.xpi
 pageshot.update.rdf
-  
+*.pyc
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -2,4 +2,4 @@ Profile
 .DS_Store
 pageshot.xpi
 pageshot.update.rdf
-  
\ No newline at end of file
+*.pyc
```
ianb: Note you can use `~/.gitconfig` to point to a global gitignore file, and ignore these files forever.

Also I think you might not have your editor configured to always include a trailing newline.

PR #159, opened 2015-02-20
r? ianbicking Add "summary" to view selector.

### LINKED ISSUE #156: Add overview to mode selector
The current selector goes between [normal | readable].  We want to add a "summary" option.

To do the summary option, `frame.html` should include the card rending copied from `collection.html`.  Then when you hit the selector it will hide the iframe and include the rendered card in its place.

We'll have to do some additional work to make this work on new-shot creation, as the current logic loads `frame.html` with dummy data, and then gets updated in place (by `js/newframe.js`).

fzzzy:
Fixes #156.

### DIFF
--- a/appengine/pageshotpages/collection.html
+++ b/appengine/pageshotpages/collection.html
@@ -24,18 +24,8 @@
     </div>
     {{endif}}
 
-    <div id="info">
-      <div id="login-widget"></div>
-      <div id="clip">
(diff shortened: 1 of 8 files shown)


ianb:
Generally looks good.  There's a comment field (looks like a small white box) to the left of the card view that doesn't appear operational, and should get removed.  The transitions look really funky, not sure why?  Like a transparent iframe appears over the card view for a bit.

PR #177, opened 2015-03-11
Selection reimplementation

ianb:
Mostly this puts back the inline selection in the addon (selection in content doesn't work yet)
Also some regressions from my rearranging of everything earlier
And minor cleanups


lib/shooter.js
```diff
@@ -377,6 +385,7 @@ const Shot = Class({
         for (let attr in attrs) {
           body[attr] = attrs[attr];
         }
+        console.log("PUT", url);
```
fzzzy: Is this supposed to be here, or is it a debug print that didn't get removed?

PR #227, opened 2015-03-26
r? ianbicking Fixes #218 Use Promise.catch instead of synchronous try/catch around the main rende...

### LINKED ISSUE #218: Use Promise.catch instead of a synchronous catch around the main render call

fzzzy:
...r call

### DIFF
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -151,6 +151,7 @@ let server = http.createServer(function (req, res) {
         footer);
     }).catch(function (e) {
       res.setHeader(content_type, "text/plain; charset=utf-8");
+      res.writeHead(500);
       res.end(e.stack);
     });
   });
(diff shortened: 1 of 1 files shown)


server/src/server.js
```diff
@@ -137,23 +137,21 @@ let server = http.createServer(function (req, res) {
...
+      res.setHeader(content_type, "text/html; charset=utf-8");
+      res.end(
+        doctype +
+        header +
+        "<script>var cached_data = " + JSON.stringify(data) + ";" +
+        script +
+        "</script>" +
+        footer);
+    }).catch(function (e) {
+      res.setHeader(content_type, "text/plain; charset=utf-8");
```
ianb: Should set the status to 500 too, right?

fzzzy: Ah, yes. I will add that.

PR #258, opened 2015-04-03
r? ianbicking Fixes #257 Cache busting query strings

### LINKED ISSUE #257: Cache static files and put cache-busting query string in links to them

fzzzy:


server/src/server.js
```diff
@@ -110,6 +114,7 @@ let server = http.createServer(function (req, res) {
         params: state.params,
         query: state.query}
     ).then(function (data) {
+      data.git_revision = git_revision;
```
ianb: camelCaseIt!

fzzzy: Whoops!

server/src/views/shell.js
```diff
@@ -9,9 +9,9 @@ exports.Shell = React.createClass({
         <title>
           PageShot
         </title>
-        <script src="/js/bundle.js" />
-        <link rel="stylesheet" href="/css/styles.css" />
-        <link rel="stylesheet" href="/css/login.css" />
+        <script src={"/js/bundle.js?git_revision=" + this.props.git_revision } />
+        <link rel="stylesheet" href={"/css/styles.css?git_revision=" + this.props.git_revision } />
+        <link rel="stylesheet" href={"/css/login.css?git_revision=" + this.props.git_revision }/>
```
ianb: Maybe we can pass something like `props.linkify=function (link) {return link +(link.contains("?") ? "?" : "&") + "bust=" + gitRevision}` and then avoid the manual link construction here.

ianb:
Pushed, assuming the nits can get picked up later

server/src/views/shell.js
fzzzy: Sounds good

PR #262, opened 2015-04-06
r? ianbicking Fixes #260 Use this.props.linkify to add the git revision to links.

### LINKED ISSUE #260: Create a props.linkify function that generates links.

fzzzy:


server/src/server.js
```diff
@@ -23,6 +23,14 @@ const jspath = "/js/",
...
   script = `
 function gotData(Handler, data) {
+  data.linkify = function (url) {
+    if (url.indexOf("?") !== -1) {
+      url += "&gitRevision=" + gitRevision;
+    } else {
+      url += "?gitRevision=" + gitRevision;
+    }
+    return url;
+  }
```
ianb: Missing semicolon, to be pedantic.  Also, why the same function twice?

fzzzy: The function needs to be defined both in the client and the server rendering paths. I'll see if I can move it somewhere that it is not duplicated.

ianb: oh, I missed that context from the diff.  That seems reasonable to me to have two copies then.

PR #275, opened 2015-04-08
r? ianbicking Fixes #269 Run the 6to5 translation against the addon code

### LINKED ISSUE #269: Use es6 in the addon panel

fzzzy:
Note that you have to generate the xpi inside the addon/dist directory,
now. I will start generating the xpi from the gulpfile next.


gulpfile.js
```diff
@@ -0,0 +1,76 @@
...
+
+gulp.task("lib-addon", function () {
+  return gulp.src("addon/lib/**/*.{js,jsx}").pipe(to5()).pipe(gulp.dest("addon/dist/lib"));
+});
+
+gulp.task("test-addon", function () {
+  return gulp.src("addon/test/**/*.{js,jsx}").pipe(to5()).pipe(gulp.dest("addon/dist/test"));
+});
+
+gulp.task("static-addon", function () {
```
ianb: We'll have some SVG shortly (Bram has made some SVG icons I believe)

gulpfile.js
```diff
@@ -0,0 +1,76 @@
...
+});
+
+gulp.task("javascript-addon", ["data-addon", "lib-addon", "test-addon", "static-addon"], function () {
+  var bundler = browserify({
+    entries: ["./addon/dist/addon-main.js"],
+    debug: true
+  });
+
+  return (function () {
+    return bundler.bundle().pipe(source("bundle.js")).pipe(gulp.dest("addon/dist"));
```
ianb: Maybe we should name bundles differently, i.e., addon-bundle.js vs website-bundle.js

ianb:
Looks like this still needs a change against the run script too, to run gulp before each cfx invocation and use `--pkgdir`?

fzzzy:
Eventually gulp is going to run the run script, instead of the run script running gulp.

gulpfile.js
fzzzy: Added svg.

fzzzy:
I added code to the gulp script that generates the addon.

PR #291, opened 2015-04-13
r? ianbicking Fixes #185 Implement the skin of the new panel. Some elements are not hooked up yet.

### LINKED ISSUE #185: Design control to add more clips
I think it would be nice to optionally add more than one clip.  The UI we have now always replaces the clip, which I actually think is an okay approach.  But sometimes I might want more than one clip.  Also text clips, so I imagine maybe "add clip [crop]" and "add clip [text select]" to both add things and switch between those two modes.

fzzzy:


addon/data/panel.css
```diff
@@ -57,28 +58,80 @@ input[type=text] {
...
+
+img.snippet {
+  position: absolute;
+  top: 4.5em;
+  left: 5%;
+  width: 90%;
 }
 
 .text {
+  display: none;
```
ianb: Is this just a form of commenting these out?

fzzzy: Yes, just for now. Do you think I should just remove them instead?

ianb: Or just include a comment indicating why they are hidden.  I wasn't sure if they would be toggled on somehow elsewhere.

fzzzy: Ok. I added such a comment.

PR #297, opened 2015-04-17
Render panel with new model

fzzzy:


addon/data/shoot-panel.js
```diff
@@ -59,20 +65,24 @@ let ShootPanel = React.createClass({
...
       <div className="link-row">
-        <a className="link" target="_blank" href={ this.props.viewUrl } onClick={ this.onLinkClick }>{ this.props.viewUrl }</a>
+        <a className="link" target="_blank" href={ this.props.shot.viewUrl } onClick={ this.onLinkClick }>{ this.props.shot.viewUrl }</a>
         <button className="copy" ref="copy" type="button" data-normal-text="Copy Link" data-copied-text="Copied!" onClick={ this.onCopyClick }>Copy Link</button>
       </div>
 
 
-      <div className="text">{ this.props.textSelection }</div>
-      <div className="comment">{ this.props.comment}</div>
+      <div className="text">{ this.props.shot.textSelection }</div>
```
ianb: There's no textSelection property, what is this supposed to be?  Maybe the text selection stuff that was in there before?  That was never on the shot properly, just a kind of little test to surface that selection functionality.  Text selections will be clips in the future (and the model supports that).

fzzzy: This is from the old view html, should I just remove it?

addon/data/shoot-panel.js
```diff
@@ -59,20 +65,24 @@ let ShootPanel = React.createClass({
...
+      <div className="comment">{ this.props.shot.comment}</div>
       <input className="comment-input" ref="input" type="text" placeholder="Say something" onKeyup={ this.onKeyup }/>
     </div>;
   }
 });
 
 self.port.on("shotData", err.watchFunction(function (data) {
-  React.render(React.createElement(ShootPanel, data), document.body);
+  console.log("shotData", Object.getOwnPropertyNames(data));
+  let as = new shot.AbstractShot(data.backend, data.id, data.shot);
```
ianb: Should just be called a shot – I'm guessing we'll want to subclass AbstractShot for the panel eventually, but either way the instance isn't abstract

addon/data/shoot-panel.js
ianb: Yeah

addon/lib/shot.js
```diff
@@ -198,11 +198,12 @@ class AbstractShot {
...
       for (let clipId in attrs.clips) {
         let clip = attrs.clips[clipId];
-        this.clips[clipId] = new this.Clip(this, clipId, clip);
+        this._clips[clipId] = new this.Clip(this, clipId, clip);
       }
     }
+
     for (let attr in attrs) {
-      if (attr != "clips" && this.REGULAR_ATTRS.indexOf(attr) == -1) {
+      if (attr !== "clips" && attr !== "id" && this.REGULAR_ATTRS.indexOf(attr) == -1) {
```
ianb: The id shouldn't show up here, as it's a separate argument to the constructor and isn't mutable.  Is it getting through somewhere?

I guess maybe it could be okay if it does tend to arise naturally, though then we should assert that `attrs.id === this.id`

fzzzy: Yes, it is getting there somehow, not sure how.

ianb: Let's do an assert of equality plus warning for now

fzzzy: Done.

PR #302, opened 2015-04-17
Update server enough to just barely work with the new model

fzzzy:


server/src/models.js
```diff
@@ -10,6 +10,8 @@ let modelMap = null,
 exports.modelMap = modelMap;
 exports.metaMap = metaMap;
 
+let shot_model = require("../../addon/dist/lib/shot.js");
```
ianb: underscores!  Also since it only really exports AbstractShot, better `const {AbstractShot} = require("...")` ?

server/src/models.js
```diff
@@ -156,10 +158,22 @@ exports.shot = function shot(state) {
     [modelMap.get(key), metaMap.get(key)]
   ).then(
     ([data, meta]) => {
-      if (! (data && meta)) {
+      if (! data) {
         return Promise.reject(new Error("No data or returned from model"));
       }
-      return Promise.resolve({data: JSON.parse(data), meta: JSON.parse(meta), identifier: key});
+
+      // FIXME we need some way to configure the url
```
ianb: I guess an environmental variable if it's too awkward to pull it out of the request.

server/src/routes.js
```diff
@@ -8,7 +8,9 @@ let React = require("react"),
   shell = require("./views/shell.js"),
   main = require("./views/main.js"),
   frame = require("./views/frame.js"),
-  content = require("./views/content.js");
+  content = require("./views/content.js"),
+  shot_model = require("../../addon/dist/lib/shot.js");
```
ianb: another one

PR #314, opened 2015-04-20
Client bugfixes

fzzzy:


server/src/views/frame.js
```diff
@@ -6,21 +7,23 @@ exports.Frame = React.createClass({
...
-      snippet = this.props.shot.getClip(clipNames[0]).image.url;
+      snippet = shot.getClip(clipNames[0]).image.url;
     }
 
     let linkTextShort = "";
-    if (this.props.shot.url) {
-      let txt = this.props.shot.url;
+    if (shot.url) {
+      let txt = shot.url;
       if (txt.length > 50) {
```
ianb: Unrelated to this commit, but we can do this ellipsis with CSS (I believe the panel does this), and truncate just what we need to keep it under a certain size.

fzzzy: That's correct. At the time I did this, I didn't want to use css because I didn't know how wide the window was going to be and the link was centered in an element that was as wide as the window.

PR #322, opened 2015-04-21
Visible comment area

fzzzy:


addon/data/shoot-panel.js
```diff
@@ -126,8 +145,10 @@ let ShootPanel = React.createClass({
         <button className="copy" ref="copy" type="button" data-normal-text="Copy Link" data-copied-text="Copied!" onClick={ this.onCopyClick }>Copy Link</button>
       </div>
 
-      <div className="comment">{this.props.shot.comment}</div>
-      <input className="comment-input" ref="input" type="text" placeholder="Say something" onKeyup={ this.onKeyup }/>
+      <div className="comment-area">
+        <div className="comment">{this.props.shot.comment}</div>
```
ianb: We should look at this later.  I'm not sure we want to both show the comment and have an input for the comment.  Maybe show the comment, then edit on click?  (Unless there's no comment, then maybe an "add comment" link, or the input area always shown)

fzzzy: Yes, I think the comments should be hidden behind a button or something, instead of inline. Otherwise they may cause the panel to overflow.

PR #323, opened 2015-04-21
Adding a .travis.yml for travis-ci.org

erikvold:


ianb:
Thanks!  This is a good incentive to write some actual tests.

I think we'll need to put gulp in there somewhere too

PR #326, opened 2015-04-21
Making some changes required to switch from cfx to jpm

erikvold:


ianb:
Merged, though somehow through bad command-fu I merged it in a way that doesn't show this commit as closed, but it is merged.

PR #327, opened 2015-04-21
Using relative paths to require local modules

erikvold:


ianb:
It looks like this still fails with an import error?  `Message: Module `shooter.js` is not found at  resource://gre/modules/commonjs/shooter.js`

erikvold:
Ah checkout https://github.com/mozilla-services/pageshot/pull/328 it's finally green :)

PR #332, opened 2015-04-23
r? ianbicking Detail ui refresh

fzzzy:


server/src/views/frame.js
```diff
@@ -40,31 +84,24 @@ exports.Frame = React.createClass({
...
-              </div>
-              <div className="if-logged-out">
-                Log in:
-                <input id="log-in-email" placeholder="email" type="text" />
-              </div>
-            </div>
-          </div>
-        </div>
         <a className="main-link" href={ shot.url }>
+          { shot.docTitle }
```
ianb: Just FYI, eventually we'll probably use shot.title which will pick from the best title (maybe shot.userTitle || shot.ogTitle || shot.docTitle)

fzzzy: That sounds great.

PR #336, opened 2015-04-24
Scroll to clip

fzzzy:


server/src/js/content-helper.js
```diff
@@ -8,4 +8,16 @@ window.addEventListener(
     window.location.origin)
 );
 
+window.addEventListener(
+  "message",
+  (m) => {
+    let node = document.getElementById(m.data.show),
```
ianb: We should be doing some assertion here, I think that `m.origin == location.origin` (i.e., the message comes from our site)

server/src/views/frame.js
```diff
@@ -20,8 +23,9 @@ let Snippet = React.createClass({
         FIXME: Support text clips
       </p>;
     }
-    return <div class="snippet-container">
-      <a href="#" onClick={ this.onClickFullPage }>
+
+    return <div className="snippet-container">
+      <a href={ "#?clip=" + clip.id } onClick={ this.onClickFullPage }>
```
ianb: should probably do `encodeURIComponent(clip.id)`, on principle.

ianb:
Just those two notes, if you want to push after them

PR #344, opened 2015-04-27
Scroll to clip

fzzzy:


server/src/js/parent-helper.js
```diff
@@ -17,10 +17,28 @@ window.onmessage = function(e) {
     if (loaded) {
       doResize();
     }
-  } else if (e.data.scrollX) {
-    let frameOffset = document.getElementById("frame").getBoundingClientRect().top;
-    let toolbarHeight = document.getElementById("toolbar").clientHeight;
-    window.scrollTo(0, frameOffset + e.data.scrollY - toolbarHeight);
+  } else if (e.data.clipWidth) {
+    window.scrollTo(0, 0);
```
ianb: What's this for?  Note later you'll need to take into account window.scrollY for some things like the rect top

fzzzy: By using scrollTo(0,0) I don't have to take into account window.scrollY...

ianb: That's what I expected, but that's an awfully scrappy approach ;)

server/src/server.js
```diff
@@ -33,7 +33,40 @@ Router.run(routes.routes, Router.HistoryLocation, function (Handler, state) {
...
+    }
+  }
+  if (!appnames.length) {
+    console.error("Error: No app was routed");
+    return;
+  }
+
+  console.log("Route to app", appnames[0]);
+
+  if (appnames[0] === "shot") {
```
ianb: It looks like it's doing a full refresh of the data here, what's the reason for that?

fzzzy: I don't think it's safe to assume the cachedData from the full page load is applicable to the partial page refresh, although right now it's safe because we don't link to any other pages.

server/src/js/parent-helper.js
```diff
@@ -17,10 +17,27 @@ window.onmessage = function(e) {
     if (loaded) {
       doResize();
     }
-  } else if (e.data.scrollX) {
-    let frameOffset = document.getElementById("frame").getBoundingClientRect().top;
-    let toolbarHeight = document.getElementById("toolbar").clientHeight;
-    window.scrollTo(0, frameOffset + e.data.scrollY - toolbarHeight);
+  } else if (e.data.clipWidth) {
+    let frameOffset = document.getElementById("frame").getBoundingClientRect().top + window.scrollY,
```
fzzzy: I switched to taking into account window.scrollY instead of using scrollTo(0,0).

PR #360, opened 2015-04-30
r?fzzzy Add users

ianb:
This has a bunch of flaws, but it does work.

Specifically I think models.js should be refactored generally – probably using a couple functions for insert/update/get, and then making specific classes for individual objects.  When we change up comments it would probably be a good time to do this.

Also the actual security is very lax.  It would be easy to take over ownership of someone else's shot, for instance.  This will be easier to fix with a refactored models.js.  But I thought I'd get the code out there first.


package.json
```diff
@@ -7,24 +7,26 @@
   },
   "dependencies": {
     "babel": "5.1.5",
-    "node-sass": "3.0.0-alpha.0",
-    "react": "0.13.1",
-    "react-router": "0.13.2",
+    "browserify": "9.0.8",
+    "cookies": "^0.5.0",
```
fzzzy: When I edit this file manually, I intentionally leave out the ^ so that any upgrades are explicit.

package.json
```diff
@@ -7,24 +7,26 @@
...
+    "gulp-jshint": "1.10.0",
+    "gulp-nodemon": "2.0.2",
     "gulp-react": "3.0.1",
+    "gulp-run": "1.6.7",
     "gulp-sass": "1.3.3",
-    "vinyl-source-stream": "1.1.0",
-    "gulp-nodemon": "2.0.2",
-    "browserify": "9.0.8",
+    "gulp-sourcemaps": "1.5.1",
+    "keygrip": "^1.0.1",
```
fzzzy: Same here

PR #406, opened 2015-05-14
Add a "Add Text Selection" button. Click it and drag a selection in the page to add a text selection.

fzzzy:


addon/data/shooter-interactive-worker.js
```diff
@@ -241,6 +251,10 @@ function restore(state, pos, scrollIntoView) {
       window.scrollTo(0, newScroll);
     }
   }
+
+  if (state === "text") {
+    addHandlers();
```
ianb: Why addHandlers() for text?  These handlers are for the rectangle selection, they shouldn't be necessary for text.

addon/data/shooter-interactive-worker.js
```diff
@@ -873,16 +887,11 @@ if (window.getSelection().rangeCount && ! window.getSelection().isCollapsed) {
...
 
-let textSelectButton;
-
 window.addEventListener("mouseup", watchFunction(function (event) {
-  if (textSelectButton) {
-    textSelectButton.parentNode.removeChild(textSelectButton);
-    textSelectButton = null;
-  }
-  if (event.target.className == "pageshot-textbutton") {
+  if (currentState !== "text") {
```
ianb: I'd like to leave the button in.  So if currentState is text, then do captureSelection() unconditionally, otherwise still add the button.

addon/lib/shooter.js
```diff
@@ -198,10 +198,15 @@ const ShotContext = Class({
       }).bind(this)));
     }, this));
     this.interactiveWorker.port.on("textSelection", watchFunction(function (textSelection) {
-      this.activeClipName = this.shot.addClip({
-        createdDate: Date.now(),
-        text: textSelection
-      });
+      if (this.activeClipName){
+        let c = this.shot.getClip(this.activeClipName);
+        c.text = textSelection;
```
ianb: This can't happen currently, can it?  Since text selection is only on the add-clip menu.  This code wouldn't be correct if, e.g., the active clip was an image clip.

Probably there's a bug currently if you do a text selection then hit the "visible" or something that produces an image clip.

fzzzy: Yes, it does happen currently. If you are viewing a text selection on the panel and you make a new text selection in the page, it happens.

fzzzy: It won't ever happen when the clip in the panel was not a text clip, though.

addon/data/shooter-interactive-worker.js
fzzzy: Sure

addon/lib/shooter.js
ianb: OK – if it did happen there's code in shot.js that would catch it too

addon/data/shooter-interactive-worker.js
fzzzy: You're right. I thought it was required, but I just tested without it and it's not.

PR #430, opened 2015-05-22
r?fzzzy r?kitcambridge Use express

### LINKED ISSUE #361: Refactor models.js
The models.js file should have proper models for our core objects, and some separate helper functions for the database access.  I think we don't really need upsert either, we can code around that in the server and with some changes to the API.

### LINKED ISSUE #410: Consider Express
I'm looking at Express, and it looks okay.  We can keep it fairly simple, using the [routing](http://expressjs.com/guide/routing.html) and the [middleware pattern](http://expressjs.com/guide/using-middleware.html) (we shouldn't jump into using any middleware libraries IMHO).  [Error handling](http://expressjs.com/guide/error-handling.html) is probably useful.  [Convenient static files](http://expressjs.com/starter/static-files.html).

No templating.  We keep it simple in a `server.js` file, and don't use the [layout they suggest](http://expressjs.com/starter/generator.html).  No configuration.

ianb:
Fixes #410, gets rid of models.js and so fixes #361, and touches a bunch of the server here and there.


server/src/server.js
```diff
@@ -1,337 +1,189 @@
...
+const path = require('path');
+const Cookies = require("cookies");
 
 const { Shot } = require("./servershot");
 const { checkLogin, registerLogin } = require("./users");
 const dbschema = require("./dbschema");
+const express = require("express");
+const bodyParser = require('body-parser');
+const morgan = require("morgan");
+const linker = require("./linker");
```
fzzzy: Why repeat const over and over instead of using the comma form?

server/src/views/frame.js
```diff
@@ -217,36 +309,56 @@ exports.Frame = React.createClass({
...
+    staticLink: req.staticLink,
+    backend: req.backend,
+    shot: req.shot,
+    id: req.shot.id,
+    shotDomain: req.url, // FIXME: should be a property of the shot
+    query: req.query,
+    params: req.params
+  });
+  let body = React.renderToString(frame);
+  body = '<!DOCTYPE html>\n' + body;
```
fzzzy: So we're not hooking up react on the client side at all any more?

gulpfile.js
```diff
@@ -35,11 +35,11 @@ gulp.task("6to5", function () {
 });
 
 gulp.task("sass", function () {
-  return gulp.src("server/src/**/*.scss").pipe(sass()).pipe(gulp.dest("server/dist/css"));
+  return gulp.src("server/src/**/*.scss").pipe(sass()).pipe(gulp.dest("server/dist/static/css"));
```
fzzzy: I don't like this extra level of folder nesting. I'd rather see all the static files move up to the top level /shared/ directory. But, we have another issue for that, so I guess we'll change the structure again later.

server/src/views/frame.js
```diff
@@ -125,13 +225,21 @@ exports.Frame = React.createClass({
...
+      let previousClip = null;
+      if (i > 0) {
+        previousClip = shot.getClip(clipNames[i-1]);
+      }
 
-      snippets.push(<Snippet key={ name } clip={ clip } linkify={ this.props.linkify } shotId={ shotId } shotDomain={ shotDomain } previousClip={ previousClip } nextClip={ nextClip } />);
+      snippets.push(<Snippet staticLink={this.props.staticLink} key={ clipId } clip={ clip }  shotId={ shotId } shotDomain={ shotDomain } previousClip={ previousClip } nextClip={ nextClip } />);
 
-      if (query.clip === name) {
+      if (query.clip === clipId) {
```
fzzzy: Does this actually work any more? The anchor tags are putting the clip id in the hash of the url, not the query string.

ianb: I realized I forgot to generally restore the functionality of the loaded page generally, this included.

gulpfile.js
ianb: Yeah, src/ is becoming less necessary – adding static/ made the routing simpler and clearer.

fzzzy: Upon thinking about it further, I think we want a new static/ directory at the top level, next to addon/, server/, and shared/, but we can do that later in another issue

server/src/server.js
ianb: I don't like the comma form, it's harder to edit and rearrange, each line doesn't stand on its own.

fzzzy: I disagree. The fact that each line doesn't stand on it's own is the advantage.

server/src/server.js
```diff
@@ -1,337 +1,189 @@
...
-    };
+app.use(morgan("dev"));
 
-    xhr.open("GET", "/data/" + key);
-    xhr.send();
-  }
+app.use(function (req, res, next) {
+  let cookies = new Cookies(req, res, dbschema.getKeygrip());
+  req.userId = cookies.get("user", {signed: true});
+  let proto = req.headers["x-forwarded-proto"] || "http";
```
kitcambridge: Alternatively, you can use `app.set('trust proxy', true)`, and then this becomes `req.protocol`. It'll also set `req.ip` based on the `X-Forwarded-For` header.

server/src/server.js
```diff
@@ -1,337 +1,189 @@
...
-    });
+app.get("/content/:id/:domain", function (req, res) {
+  let shotId = req.params.id + "/" + req.params.domain;
+  Shot.get(req.backend, shotId).then((shot) => {
+    if (! shot) {
+      simpleResponse(res, "Not found", 404);
+      return;
+    }
+    res.writeHead(200);
+    res.end(shot.staticHtml({
```
kitcambridge: `res.send()` is a bit more compact...but `writeHead + end` works just as well.

ianb: Leftover stuff from before Express – but probably I should search for all .end()s and update them

server/src/server.js
```diff
@@ -1,337 +1,189 @@
...
+  });
+});
 
-  if (pth.startsWith(modelspath)) {
-    modelname = pth.slice(modelspath.length);
-    storeMap = models.modelMap;
+app.put("/data/:id/:domain", function (req, res) {
+  let bodyObj = req.body;
+  if (typeof bodyObj == "string") {
+    bodyObj = JSON.parse(bodyObj);
```
kitcambridge: Does `body-parser` not parse this correctly?

ianb: At some point no?  Or the addon isn't setting Content-Type?  This might no longer be necessary due to changes that happened elsewhere.

PR #439, opened 2015-05-28
Fix hooking up react on the client side in the express-wip branch

fzzzy:


server/src/server.js
```diff
@@ -46,7 +44,7 @@ app.use(function (err, req, res, next) {
 });
 
 app.post("/api/register", function (req, res) {
-  let vars = req.params;
+  let vars = req.body;
```
ianb: Oops!  Maybe this was the second bug causing login errors?

server/src/views/frame.js
```diff
@@ -115,14 +64,14 @@ class Snippet extends React.Component {
...
-      <a href={'#clip=' + encodeURIComponent(clip.id)}>
+      <a href={'?clip=' + encodeURIComponent(clip.id)}>
         { node }
       </a>
       <p>
         <img ref="commentBubble" className="comment-bubble"
           src={ closed ? this.props.staticLink("img/comment-bubble.png") : this.props.staticLink("img/comment-bubble-open.png") }
           onClick={ this.onClickComment.bind(this) } />
-        <a href={'#clip=' + encodeURIComponent(clip.id)}>
+        <a href={'?clip=' + encodeURIComponent(clip.id)}>
```
ianb: Does this mean we're doing a page reload every time you view a clip?

server/src/server.js
fzzzy: Yep. This was preventing login

server/src/views/frame.js
fzzzy: Yes. We can't use the hash for the clip id, because react requires rendering the same thing on the server and the client, and the hash doesn't get sent to the server.

Before, react-router was handling partial page refreshes and pushState, but we have to do it ourselves if we want that now.

ianb: React doesn't *require* rendering the same thing, does it?  That is, on the server it'll always render "clip 1/5" or something, and if the hash is `#clipId=secondClipId` then when we make React do its thing on the client it'll just update that in place.

fzzzy: Yes, it requires the same thing. If it doesn't see the same thing, checksums don't match, it complains, and the entire page is re-rendered on the client instead of just hooking up the client side event handlers and leaving the dom as it was from the server.

ianb: Then, weird though it is, we can double-render: first render with the server variables, then read location.hash and change the variables and then render again

fzzzy: Why?

ianb: Because it's not a new resource; the clip that you are focused on is a classic use of a fragment.  We're doing something slightly fancier than scrolling, but conceptually it's the same.

fzzzy: I don't agree. The state of the previous/next buttons is different when each clip is highlighted. You can't cache the html from the page where no clip is selected to use with a page where a clip is selected. It's different html from the server. It's a different resource.

ianb: We can't render the content highlighting on the server anyway, it is contained in an iframe.

You can cache the HTML, it just isn't a completely accurate representation of the clip number indicator.  The server HTML doesn't even contain the scroll position, nor can the server calculate where the highlight should appear.  So the only thing the server can add is a little indicator.  The active clip is primarily a client-side mechanism.

fzzzy: I really think we should draw the highlight on the server, by capturing the rect of the selection on the client and the width of the window on the client. It would make the highlight a lot more accurate.

fzzzy: Another advantage of storing the clip state in the query string is that the server can do analytics on which clips are being shared. Storing it in the hash makes that needlessly complicated.

ianb: It'll never be more accurate than doing the rendering on the client, where we can use DOM introspection to actually locate the elements.

We need to do client-side analytics anyway, we can never capture enough on the server alone.  Plus if we use pushstate to suppress a reload (which we'd have to do – reloading on clip selection is way too slow) we wouldn't get the server analytics either.

PR #444, opened 2015-05-29
r? ianbicking Fixes #394 Move all dist directories to one toplevel dist dir, and also move server static files to a toplevel static dir.

### LINKED ISSUE #394: move dist directories to one dist directory at pageshot/dist
Instead of having addon/dist shared/dist and server/dist, there should be a single pageshot/dist directory with addon, shared, and server subdirectories.

fzzzy:


ianb:
Looks like there's a leftover reference to addon/dist somewhere that is causing the build to fail?

fzzzy:
Fixed.

kitcambridge:
This occasionally causes nodemon to become very confused. I think it detects a change to a source file, rebuilds `dist`, sees that `dist` has changed, and then rebuilds again. Oddly, `dist` is in the ignore list, so it shouldn't do that—but the rebuild cycle went away after I reverted this patch. Maybe we can dig into it more tomorrow?

fzzzy:
I wish nodemon would actually print the path of the file that changed. Maybe it can be configured to do that.

ianb:
I saw the same behavior at Kit – I tried putting in `dist/**` just in case, which maybe improved things?  But I was paying attention to something else so I didn't really check.

kitcambridge:
Whoa, looks like I have 54 nested `dist` directories. Going to clobber those and rebuild from scratch.

PR #445, opened 2015-05-29
r?fzzzy Clip hash

ianb:


server/src/static/js/parent-helper.js
```diff
@@ -8,39 +8,40 @@ function doResize() {
...
-    document.body.appendChild(d);
-
-    window.scrollTo(0, scrollY);
+  let message = m.data;
+  let type = message.type;
+  if (! type) {
+    console.warn("Parent iframe received message with no type:", message);
+    return;
+  }
+  if (type == "setHeight") {
```
fzzzy: Please use ===

server/src/views/frame.js
```diff
@@ -1,16 +1,10 @@
...
 const { getGitRevision } = require("../linker");
 
-let IS_BROWSER = true;
-
-try {
-  window;
-} catch (e) {
-  IS_BROWSER = false;
-}
+let IS_BROWSER = typeof window != "undefined";
```
fzzzy: Please use !==

server/src/clientglue.js
```diff
@@ -1,30 +1,88 @@
...
-    params: data.params
-  });
+let model;
+
+exports.setModel = function (data) {
+  let firstSet = ! model;
+  model = data;
+  model.shot = new AbstractShot(data.backend, data.id, data.shot);
+  render();
+  console.log("firstSet", firstSet);
```
fzzzy: I'm not sure this debug print should stick around

server/src/clientglue.js
```diff
@@ -1,30 +1,88 @@
...
-  window.url = url;
-  window.clientGlue = clientGlue;
-} catch (e) {
+function refreshHash() {
+  let clipId = null;
+  let match = (/clip=([^&]+)/).exec(location.hash);
+  if (match) {
+    clipId = decodeURIComponent(match[1]);
+  }
+  console.log("refresh hash", clipId, model.activeClipId);
```
fzzzy: Probably don't want this debug print either?

server/src/clientglue.js
```diff
@@ -1,30 +1,92 @@
...
+      console.error("Message:", message);
+      throw e;
+    }
+    _pendingMessage = null;
+  }
+  if (frame.contentDocument.readyState == "complete") {
+    post(postMessage);
+  } else {
+    _pendingMessage = postMessage;
+    frame.contentWindow.onload = function () {post();};
```
fzzzy: Why is _pendingMessage required? Why can't this just say post(postMessage) here?

ianb: I didn't want the messages to stack up, though I guess by using onload they couldn't anyway

fzzzy: In that case _pendingMessage is probably required, but it should be an array and _pendingMessages.push(postMessage) instead

ianb: I actually specifically didn't want them to stack, so no queue.  I've added a commit that removes _pendingMessage, it was superfluous

server/src/views/frame.js
```diff
@@ -149,7 +128,7 @@ class Frame extends React.Component {
     }
 
     let shot = this.props.shot;
-    let query = this.props.query;
+    let activeClipId = this.props.activeClipId;
```
fzzzy: Is anything ever passing activeClipId in props? I don't see where

ianb: It's part of the model object, and all those attributes are passed in

server/src/clientglue.js
```diff
@@ -1,30 +1,87 @@
...
+};
+
+function render() {
+  setGitRevision(model.gitRevision);
+  let attrs = {
+    staticLink: staticLink
+  };
+  for (let attr in model) {
+    attrs[attr] = model[attr];
+  }
```
fzzzy: You can write this like {staticLink: staticLink, ...model}

kitcambridge: Neat, I didn't know you could do `{...model}`. Is that a JSX extension?

`{staticLink, ...model}` is even more compact. :wink:

fzzzy: No, it's just part of es6 as far as I can tell.

kitcambridge: This syntactic sugar is rather sweet...

PR #456, opened 2015-06-03
r? ianbicking Fixes #438 Put an alpha badge on the panel

### LINKED ISSUE #438: Put "alpha" onto the panel
I'd like a little something to scare people off, similar to what Hello has, but with the scarier "alpha":

screen shot 2015-05-27 at 2 07 06 pm

fzzzy:


addon/data/panel.css
```diff
@@ -79,6 +79,12 @@ input[type=text]:focus {
   overflow: hidden;
 }
 
+.alpha-badge {
+  position: absolute;
+  top: 0px;
+  left: 0px;
+}
```
ianb: I think this should have `pointer-events: none` too, so it can't overlap any functional element.  Or it could be a background image.

fzzzy: Added.

PR #458, opened 2015-06-03
r? ianbicking Fixes #309 If a shot has no clips, 404.

### LINKED ISSUE #309: Shot should get deleted when last clip is deleted
After #308 is implemented:

In `lib/shooter.js` the ShotContext should notice when it is destroying that its shot has no clips, and delete the shot on the server at that point.

I don't think we need to be too aggressive in deleting the shot early.  The server could also check if a shot has no clips, and refuse to render it in that case.  (Encourages a dirty database though.)

fzzzy:


server/src/server.js
```diff
@@ -175,7 +175,7 @@ app.get("/", function (req, res) {
 app.get("/:id/:domain", function (req, res) {
   let shotId = req.params.id + "/" + req.params.domain;
   Shot.get(req.backend, shotId).then((shot) => {
-    if (! shot) {
+    if (! shot || Object.getOwnPropertyNames(shot._clips).length === 0) {
```
ianb: `shot.clipNames().length === 0` would be a better way to say this

fzzzy: We only have the json at this point, don't we?

ianb: No, it's a real instance (created by `Shot.get()`).  Also if it was JSON then it would be `shot.clips` and not `shot._clips`

fzzzy: Ok cool. I misunderstood while inspecting the object

fzzzy: Fixed.

PR #459, opened 2015-06-03
Fixes #443 Give an os native notification when you press the copy lin…

### LINKED ISSUE #443: Copy link doesn't show any indicator it worked
When you click the icon it does copy the link, but doesn't give any indication that it happened.  Maybe a popup/notification.

fzzzy:
…k button

### DIFF
--- a/addon/lib/shooter.js
+++ b/addon/lib/shooter.js
@@ -289,10 +289,8 @@ const ShotContext = Class({
       clipboard.set(this.shot.viewUrl, "text");
       notifications.notify({
         title: "Link Copied",
-        text: "Link to the shot detail copied to the clipboard.",
-        data: "",
-        onClick: function (data) {
-        }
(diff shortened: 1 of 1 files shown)


addon/lib/shooter.js
```diff
@@ -286,6 +287,13 @@ const ShotContext = Class({
     },
     copyLink: function () {
       clipboard.set(this.shot.viewUrl, "text");
+      notifications.notify({
+        title: "Link Copied",
+        text: "Link to the shot detail copied to the clipboard.",
+        data: "",
```
ianb: data and onClick are optional.

Maybe the text could be "The link to your shot has been copied to the clipboard"?  Also, can we use the link icon here? addon/data/copy.png

fzzzy: Removed optional attributes, changed text, added icon. Looks good with the icon.

Issue #446, opened 2015-06-01
Schema migrations


kitcambridge:
I added a few secondary indexes in kitcambridge@c3b01c6542dd71297136b9416ae47f650b3b3858, but Postgres won't support MySQL's `CREATE INDEX IF NOT EXIST` until [version 9.5](http://www.postgresql.org/docs/devel/static/sql-createindex.html). (Incidentally, 9.5 adds [upserts](http://www.postgresql.org/docs/devel/static/sql-insert.html) in the form of `ON CONFLICT UPDATE`, too).

In the meantime, though, it looks like we'll need to roll our own migration code. The FxA folks use https://github.com/mozilla/mysql-patcher; maybe we can adapt it for Postgres?

fzzzy:
https://www.npmjs.com/package/pg-patcher

kitcambridge:
Sweet! I was tired and didn't research further. :wink:

ianb:
When migrating after latest change, we need to be sure we can have body='' and then try to pick up the body from the json

fzzzy:
Add the "head" and "body" columns to the data table.

PR #461, opened 2015-06-04
r?fzzzy ianbicking FxA integration

kitcambridge:
Whew, sorry this took so long! This is a first cut at adding FxA integration to PageShot. The "sign in" and "create account" buttons are wired up, and you can change your nickname. The styling could use some love. I'd love feedback on the React views, and the general approach.

Dev client credentials: `CLIENT_ID=209ee70cb3cdc851 CLIENT_SECRET=efde5874713abf2a8ec9d8af8488fae7c7e0a933dba08208f644fa93b5622dab` (We can check these into the repo; the FxA folks can provide us with real production credentials later).


addon/lib/main.js
```diff
@@ -51,6 +52,15 @@ var shootPanel = panels.Panel({
 
 watchWorker(shootPanel);
 
+shootPanel.port.on("signIn", function () {
+  let oAuthHandler = new user.OAuthHandler('http://localhost:10080');
```
fzzzy: Don't forget to get this out of the environment before landing.

addon/lib/main.js
```diff
@@ -16,6 +16,7 @@ const helperworker = require("./helperworker");
 const { ToggleButton } = require('sdk/ui/button/toggle');
 const panels = require("sdk/panel");
 const { watchFunction, watchWorker } = require("./errors");
+const user = require("user");
```
ianb: Later on we'll want to rename this deviceid or something.  Otherwise we're all going to get really confused later.

addon/lib/main.js
ianb: This can just be replaced with `backend`

server/src/server.js
```diff
@@ -186,6 +196,132 @@ app.get("/:id/:domain", function (req, res) {
...
+          res.send({
+            access_token: accessTokenHex
+          });
+        });
+      });
+    });
+  }).catch(next);
+});
+
+function setState(deviceId, state) {
```
ianb: I'd like to keep SQL itself out of server.js.  Probably a new module to handle these?

server/src/users.js
```diff
@@ -2,7 +2,7 @@ const db = require("./db");
 
 exports.checkLogin = function (userId, secret) {
   return db.select(
-    `SELECT secret FROM users WHERE id = $1`,
+    `SELECT secret FROM devices WHERE id = $1`,
```
ianb: oh, and you renamed the table, cool ;)

package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "babel": "5.2.17",
     "body-parser": "1.12.4",
+    "boom": "^2.7.0",
```
ianb: We've been using strict versions in package.json (i.e., `"2.7.0"`)

server/src/dbschema.js
```diff
@@ -2,16 +2,25 @@ const db = require("./db");
 const Keygrip = require('keygrip');
 
 const createSQL = `
-CREATE TABLE IF NOT EXISTS users (
+CREATE TABLE IF NOT EXISTS accounts (
+  id BYTEA PRIMARY KEY,
+  token BYTEA,
+  email varchar(256) NOT NULL
```
ianb: Generally I try to use `TEXT` as often as possible

server/src/dbschema.js
```diff
@@ -2,16 +2,25 @@ const db = require("./db");
 const Keygrip = require('keygrip');
 
 const createSQL = `
-CREATE TABLE IF NOT EXISTS users (
+CREATE TABLE IF NOT EXISTS accounts (
+  id BYTEA PRIMARY KEY,
+  token BYTEA,
```
ianb: These are bytes?  That's kind of surprising I guess.

server/src/dbschema.js
```diff
@@ -20,6 +29,13 @@ CREATE TABLE IF NOT EXISTS signing_keys (
   created TIMESTAMP DEFAULT NOW(),
   key TEXT
 );
+
+CREATE TABLE IF NOT EXISTS states (
+  state BYTEA PRIMARY KEY,
+  deviceid VARCHAR(200) REFERENCES devices(id) ON DELETE CASCADE
+);
+
+CREATE INDEX ON states(deviceid);
```
kitcambridge: We'll want to give this index a name and follow up with #446, to avoid creating new indexes each time we restart the server.

server/src/errors.js
```diff
@@ -0,0 +1,32 @@
+const boom = require('boom');
+
+exports.create = function create(statusCode, errno, message, data) {
```
ianb: What's the advantage of boom here?

server/src/server.js
```diff
@@ -186,6 +196,132 @@ app.get("/:id/:domain", function (req, res) {
   });
 });
 
+const oAuthBaseURI = 'http://127.0.0.1:9010/v1',
+  contentBaseURI = 'http://127.0.0.1:3030',
+  profileBaseURI = 'http://127.0.0.1:1111/v1',
+  oAuthClientId = 'ac7ee3c317531aab',
+  oAuthClientSecret = 'eb7c92bd4616a7c1f86595492f360f55cc453974b062e71456d2d4b104f307f8';
```
ianb: I guess these might cause problems being in the source?

Generally I think we should just read them from process.env.  We can make sure they are set properly on the live server.  I'm not sure what to do in development?

kitcambridge:
I think this is finally ready!

server/src/server.js
```diff
@@ -189,6 +215,76 @@ app.get("/:id/:domain", function (req, res) {
...
+      return registerAccount(req.userId, accountId, accessToken);
+    }).then(() => {
+      res.send({
+        access_token: accessToken
+      });
+    });
+  }).catch(next);
+});
+
+app.use(function (err, req, res, next) {
```
kitcambridge: I moved the error handler last so that `.catch(next)` can catch and report errors in promise chains.

addon/lib/main.js
kitcambridge: Maybe the FxA additions belong in a separate `profile` module? Also, since the OAuth code lives elsewhere now, I should revert the changes to this file...

addon/lib/helperworker.js
```diff
@@ -31,6 +50,64 @@ function resetPageMod(backend) {
...
+
+        let handler = new user.OAuthHandler(backend);
+        handler.getOAuthParams().then(defaults => {
+          worker.port.emit("account", id, toSuccessResponse());
+
+          let params = Object.assign({ action }, defaults);
+          return handler.logInWithParams(params).then(response => {
+            return handler.getProfileInfo();
+          }).then(profile => {
+            // TODO: Don't clobber profile info if the user has already set a
```
kitcambridge: Maybe a separate `setDefaultProfileInfo` call that only updates your nickname and avatar if you haven't set one?

addon/data/viewerworker.js
```diff
@@ -27,6 +27,51 @@ document.addEventListener("error", function (event) {
...
+  }));
+  document.dispatchEvent(event);
+});
+
+self.port.on("profileRefresh", function () {
+  var event = document.createEvent("CustomEvent");
+  event.initCustomEvent("profile-refresh", true, true, null);
+  document.dispatchEvent(event);
+});
+
```
ianb: Maybe all this messaging could be simpler.  We basically want to keep data in sync between the page and the addon.  I think we can consider the addon the canonical source?  So we have messages from the page asking to do things: to sign in, to create an account, to set the display name, and to set the avatar.  From the addon we have: set the state.  That includes the login status, profile information, display name, etc.  The addon may have to wait to set the state if the page isn't ready to accept the message yet, though I think the way the helper initializes, all the Javascript on the page should have executed (including event listeners) before the worker starts up.

So it seems to me we should have some events from the page: request-signin, request-create, set-profile-info

Then from the addon to the page: set-profile-info.

I've used a pattern of names-with-dashes in event names, and camelCase in message names.  This is a silly.  For the most part viewerworker just pipes stuff through, we should probably keep the exact same names, and potentially do some of these in a loop.

kitcambridge: Agreed on simplifying.

* We can ax `got-profile` and `got-profile-update`, since they're taken care of by `profile-refresh` to set the state. If you're already signed in, but the page didn't pick up on that, we can just emit `profile-refresh`.
* Also, `got-account` just tells the page, "Hey, I'm starting the OAuth flow." I thought this might surface errors in the OAuth flow, too, but it doesn't, and they're not user-actionable. So we can ax that.

addon/lib/user.js
```diff
@@ -1,7 +1,11 @@
+const { Cc, Ci, Cu } = require('chrome');
```
ianb: I think jshint gives unused import warnings for some of these.  Not sure how I feel – this is a common import, but I also like to keep things jshint clean even when it makes me do things that feel nitpicky.

addon/lib/user.js
```diff
@@ -64,3 +68,129 @@ function makeUuid() {
 exports.getUserInfo = function () {
   return ss.storage.userInfo;
 };
+
+exports.updateProfileInfo = function (profile) {
```
ianb: I guess `ss.storage.profileInfo` is info taken from your FxA, but `ss.storage.userInfo` is actually device info.  Maybe the latter should be renamed to clarify?  As it is even I get confused about which is which.

addon/lib/user.js
```diff
@@ -64,3 +68,129 @@ function makeUuid() {
...
+  if (! profile) {
```
ianb: Why would you get an empty profile?  I might be include to throw an error here.

addon/lib/user.js
```diff
@@ -64,3 +68,129 @@ function makeUuid() {
...
+    return;
+  }
+  let currentInfo = ss.storage.profileInfo;
```
ianb: since we have to set ss.storage.profileInfo anyway, this could be `let currentInfo = ss.storage.profileInfo || {}` and forget the next if statement

addon/lib/user.js
```diff
@@ -64,3 +68,129 @@ function makeUuid() {
...
+  if (! currentInfo) {
+    currentInfo = ss.storage.profileInfo = {};
+  }
+  let attrs = Object.keys(profile);
+  for (let attr of attrs) {
```
ianb: This logic is complicated...?  Could it just be:

```javascript
for (let attr in profile) {
  currentInfo[attr] = profile[attr];
}
```

?  Or are you worried about null values that don't mean erase, or...?

addon/lib/user.js
```diff
@@ -64,3 +68,129 @@ function makeUuid() {
...
+    }
+  }
+  ss.storage.profileInfo = currentInfo;
+};
+
+exports.getProfileInfo = function () {
+  return ss.storage.profileInfo;
+};
+
+exports.updateLogin = function (backend, info) {
```
ianb: I think this more means `updateProfile` or `updateAccount`?  This is basically updating the `accounts` table with profile info for the user, right?  I suppose it is also assigning the reference between the device and the account...?

addon/lib/user.js
```diff
@@ -64,3 +68,129 @@ function makeUuid() {
...
+    }).then(response => {
+      this.profileDeferred.resolve(new FxAccountsProfileClient({
+        serverURL: parameters.profile_uri,
+        token: response.access_token
+      }));
+      return response;
+    });
+  }
+
+  logIn() {
```
ianb: Is this method used anywhere?

server/src/clientglue.js
```diff
@@ -21,9 +22,7 @@ exports.setModel = function (data) {
 
 function render() {
   setGitRevision(model.gitRevision);
-  let attrs = {
-    staticLink: staticLink
-  };
+  let attrs = { staticLink, ready, signUp, signIn, getProfile, updateProfile };
```
ianb: I don't think these need to be properties?  staticLink is a special case because it is used on the server and is request-specific.  But the rest are only used on the client and can be called globally.  So frame.js can just include these functions and call them.

server/src/views/profile.js
```diff
@@ -0,0 +1,249 @@
...
+function delay(duration) {
+  return new Promise(resolve => setTimeout(resolve, duration));
+}
+
+exports.ProfileButton = class ProfileButton extends React.Component {
+  constructor(props) {
+    super(props);
+    this.onDocumentClick = this.dismiss.bind(this);
+    this.onRefreshProfileBound = this.onRefreshProfile.bind(this);
+    this.state = {
```
ianb: avatar, nickname, and email can probably all be props, while isExpanded and isFetching are probably right to have as state.  I think `getInitialState` is usually how that state is set

fzzzy: getInitialState was the way to do it with old style classes, but with es6 classes the constructor is the recommended place for it

server/src/views/profile.js
```diff
@@ -0,0 +1,249 @@
...
+    this.props.updateProfile(profile).then(() => {
+      this.fetchProfile();
+    }).catch(err => {
+      console.error("Error updating profile", err);
+    });
+  }
+
+  componentDidMount() {
+    window.addEventListener("click", this.onDocumentClick);
+    document.addEventListener("profile-refresh", this.onRefreshProfileBound);
```
ianb: If that data is props instead of state, I expect the handler will be global and will trigger a refresh

ianb:
That's it from me!  We'll have @brampitoyo look at the visual design more closely after merge.

addon/lib/user.js
kitcambridge: > Or are you worried about null values that don't mean erase...

Good catch! I overloaded `updateProfileInfo` to mean, "update this profile record with the user's changes," and "fill in default values from FxA if the user hasn't provided them yet," because I didn't know how I wanted them to fit together. This definitely calls for separate methods now, though.

server/src/views/profile.js
kitcambridge: I could make them props, and have `Frame` pass a function that updates the profile. In that case, though, wouldn't `avatar`, `nickname`, and `email` still need to be state (just on `Frame`, not `ProfileButton`)?

ianb: In clientglue.js I think it should keep the profile information in the model that it uses for render().  Then clientglue is what catches the profile-update event from viewerworker, and then rerenders the entire page.

ianb: clientglue is kind of the controller for the entire shot page

PR #465, opened 2015-06-05
r?ianbicking Fixes #450 Move publish logic into bin/publish.sh

### LINKED ISSUE #450: Create bin/publish script
Right now the publish logic is still in `addon/run`, but we should extract it to its own script in `bin/` – once done `addon/run` can be removed.

fzzzy:


bin/publish.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
```
ianb: There should be a `set -e` at the top of the shell file (causes all errors to be fatal)

ianb:
set -e is my only concern; merge after?

bin/publish.sh
fzzzy: Done.

PR #479, opened 2015-06-09
r? ianbicking Fixes #470 Fully qualify the urls to our own content in the framed content

### LINKED ISSUE #470: iframe on shot page isn't being resized
I'm not seeing any errors in the console

fzzzy:


server/src/static/js/content-helper.js
```diff
@@ -1,11 +1,15 @@
 /*jslint browser: true */
```
ianb: Are all the changes in this file superfluous?

fzzzy: Whoops, yes

PR #482, opened 2015-06-09
r? ianbicking White text clip background

fzzzy:


ianb:
I tried shooting this clip using this branch and everything went weird: http://books.aidanf.net/learn-swift

For instance, I made a text clip of http://books.aidanf.net/learn-swift and got this:

screen shot 2015-06-09 at 1 53 46 pm

(Note I've selected the text – otherwise the text is white on a white background)

Specific to that same page, I errors in the shot page when selecting the text from the clip: `Uncaught error: Error: Invariant Violation: ReactMount: Two valid but unequal nodes with the same `data-reactid`: .1p7tmak1ybk.1.0.6:$u5imc6rqfwy5.0 at http://localhost:10080/static/js/server-bundle.js?rev=a38a371e79242d6d20dc3c90e6fdcdad072e1cb3@22471:14`

In the iframe on that page it also picked up the shot page background color, not sure why, though maybe unrelated.

Probably an existing bug, but clicking through to the text clip location on the shot page doesn't work, on that page and other pages I try.

The drop-shadow on text clips doesn't match regular clips.  I guess I thought regular clips had a shadow, but I see they don't.  The two should look the same.

fzzzy:
Clicking through to the text clip is an existing bug on master -- I opened a new bug #500

fzzzy:
The InvariantViolation only happens on this specific page, which is very very strange... This is going to take more investigation. I opened #501

ianb:
We might also have to explicitly set the color of links in text clips?

PR #502, opened 2015-06-10
r? ianbicking Fixes #500 Implement jumping to a text clip.

### LINKED ISSUE #500: Jumping to text clips doesn't work

fzzzy:


server/src/static/js/content-helper.js
```diff
@@ -33,24 +33,40 @@ window.addEventListener(
...
-    right: loc.right
+    top: 0,
+    bottom: 0,
+    left: 0,
+    right: 0
   };
-  let topLeft = findElement(loc.topLeftElement);
+
+  if (clip.text) {
+    topLeft = findElement("#" + clip.text.location.selectionStart);
```
ianb: Did I use inconsistent selectors for text and images?  Yuck.  I guess maybe for now we should open a ticket to straighten that out.

PR #505, opened 2015-06-10
r?ianbicking Add Convict

kitcambridge:
Convict is a little config utility that can load configs from JSON files, arguments, and environment variables. I think this would make it easier to pass in values like the OAuth client ID and secret in #461.


server/src/config.js
```diff
@@ -0,0 +1,31 @@
+require("envc")();
```
ianb: What does envc do?  A comment might be helpful, as the construct is pretty mysterious looking ;)

server/src/db.js
```diff
@@ -1,11 +1,8 @@
-let user = process.env.DB_USER || process.env.USER;
-let pass = process.env.DB_PASS;
-let host = process.env.DB_HOST || "localhost:5432";
-
-pass = pass ? ":" + pass : "";
+const config = require("./config").root();
 
 let pg = require("pg");
-let constr = `postgres://${user}${pass}@${host}/${user}`;
+let credentials = config.db.password ? `${config.db.user}:${config.db.password}` : config.db.user;
```
ianb: Is URL encoding required in any of this?  (Not that it was there before...)

ianb: Maybe what we should really do is have the user pass in a connection string, and forget about the separate settings.

server/src/db.js
```diff
@@ -1,11 +1,8 @@
-let user = process.env.DB_USER || process.env.USER;
-let pass = process.env.DB_PASS;
-let host = process.env.DB_HOST || "localhost:5432";
-
-pass = pass ? ":" + pass : "";
+const config = require("./config").root();
```
ianb: How does this work?  Why doesn't config.js export anything?  What is this .root() thing?

server/src/db.js
fzzzy: I swear we made that change a long time ago? I remember thinking maybe it got lost in a botched merge?

ianb: Many refactorings later...

server/src/db.js
kitcambridge: `config.js` does export the Convict schema, but it's easy to overlook (`let conf = module.exports = convict({`). I'll move the `module.exports` assignment to the bottom, so that it's easier to follow. `.root()` just converts the exported schema into a POJO.

server/src/db.js
kitcambridge: I think it's convenient to have separate params—certainly for the username and password—that I can override with separate environment variables, instead of building the full URL myself. But I can change it to the URL if you'd like.

ianb: Ah, I didn't think about overriding them with environmental variables.  I suppose that's reasonable.

PR #512, opened 2015-06-12
r? ianbicking Fix microformat shiv

### LINKED ISSUE #473: Errors in a shot
```
console.log: pageshot: Internal error http://toddwschneider.com/posts/mortgages-are-about-math-open-source-loan-level-analysis-of-fannie-and-freddie/
console.log: pageshot: Error from worker:  : {"name":"ReferenceError","message":"ReferenceError: assignment to undeclared variable href"}
console.log: pageshot: Promise rejected with error: {"name":"ReferenceError","message":"ReferenceError: assignment to undeclared variable href"}
console.log: pageshot: Failed assertion: Bad attr to new Shot(): Array ["name","message"]
```

fzzzy:
I'm not sure we actually want to make changes to this library since it is not our code, but if we're cool with that, this pull request fixes #473.

### DIFF
--- a/addon/data/error-utils.js
+++ b/addon/data/error-utils.js
@@ -21,7 +21,8 @@ function getFilename() {
 function makeError(exc, info) {
   var result = {
     name: exc.name || "ERROR",
-    message: exc+""
+    message: exc+"",
+    stack: exc.stack
   };
   if (info) {
     for (var attr in info) {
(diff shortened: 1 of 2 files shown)


fzzzy:
This is the interesting commit here:

https://github.com/fzzzy/pageshot/commit/4851bd0182b96c3eeb5f0ca49a6fb11778e7b77c

Atom automatically fixed all the whitespace, so I did that in a different commit.

ianb:
Yeah, I made a change like this previously (around another undeclared variable I think).  I submitted it upstream but never got a response.  Eh.  Moving to a library based on http://schema.org/ sometime would be good, but for now this is fine.

PR #513, opened 2015-06-12
r?fzzzy Fix #510, set default background color

### LINKED ISSUE #510: PageShot web shows default background if site has no background-color declaration in its <body>
Original:
http://metservice.com/towns-cities/auckland/auckland-central

PageShot:
http://pageshot.dev.mozaws.net/h7lmd592/metservice.com

I’d like it to give the full-site snapshot a background value of white. Maybe it’s tricky to do?

ianb:


fzzzy:
I swear this was set before, but I guess not.

PR #514, opened 2015-06-12
r?fzzzy Fix #416, asciify all domains used for the purpose of IDs

### LINKED ISSUE #416: Unicode domain causes error
[Example page](http://thiébaud.fr/robots.txt.html)

ianb:


addon/lib/shooter.js
```diff
@@ -549,6 +554,14 @@ function urlDomain(urlString) {
       domain = "unknown";
     }
   }
+  if (domain.search(/^[a-z0-9.\-]+$/i) == -1) {
+    // Probably a unicode domain; we could use punycode but it wouldn't decode
+    // well in the URL anyway.  Instead we'll punt.
+    domain = domain.replace(/[^a-z0-9.\-]/ig, "");
```
fzzzy: Won't this cause collisions then? Is that ok?

PR #519, opened 2015-06-16
r? ianbicking Fixes #478 Resize the panel when redisplaying it, depending on whether we are editing the current shot or not

### LINKED ISSUE #478: When you redisplay a page that has an active shot, and you have edited it, it is not resized
It starts over with the small size:

screen shot 2015-06-09 at 10 50 41 am

The large size should be sticky.  Maybe a property on the ShotContext.

fzzzy:


addon/lib/main.js
```diff
@@ -17,6 +17,8 @@ const { ToggleButton } = require('sdk/ui/button/toggle');
 const panels = require("sdk/panel");
 const { watchFunction, watchWorker } = require("./errors");
 
+let isEditing = new Map();
```
ianb: Why can't this just be a property on the ShotContext?

addon/lib/main.js
```diff
@@ -102,13 +104,18 @@ const PanelContext = {
     shotContext.isShowing();
     shootButton.checked = true;
     this.updateShot(this._activeContext, this._activeContext.shot.asJson());
+    if (isEditing.get(this._activeContext.shot.id)) {
```
ianb: I.e., here it would be `if (this._activeContext.isEditing)`

addon/lib/main.js
```diff
@@ -150,19 +157,19 @@ const PanelContext = {
         backend: shotContext.shot.backend,
         id: shotContext.shot.id,
         shot: shotContext.shot.asJson(),
-        activeClipName: shotContext.activeClipName
+        activeClipName: shotContext.activeClipName,
+        isEditing: isEditing.get(shotContext.shot.id)
       }
```
ianb: And then right here, do a shootPanel.resize() based on shotContext.isEditing

addon/lib/shooter.js
```diff
@@ -302,8 +302,8 @@ const ShotContext = Class({
     openLink: function (link) {
       tabs.open(link);
     },
-    setSize: function(size) {
-      this.panelContext.setSize(size);
+    setEditing: function(editing) {
```
ianb: And this is `this.isEditing = true` then `this.updateShot()`

addon/lib/main.js
fzzzy: It can, I just didn't think of it

ianb:
I think the logic will be simpler moving stuff into ShotContext

fzzzy:
PR updated to store isEditing on the shot context.

ianb:
Merging, but something to test:

1. Start PS on tab A
2. Start PS on tab B
3. Switch to tab A and hit "editing"
4. Switch back and forth between two tabs

PR #521, opened 2015-06-16
r? ianbicking No console log

fzzzy:
Replace all use of console.log with warn, info, or error.


ianb:
d034fe5 looks good, but the other pull request needs resolving before merging this

PR #549, opened 2015-06-25
r?fzzzy Add Index page that lists all the shots.

ianb:
Factor the domain-only URL display into a getter

### DIFF
--- a/server/src/servershot.js
+++ b/server/src/servershot.js
@@ -82,3 +82,44 @@ Shot.getRawValue = function (id) {
     };
   });
 };
+
+Shot.getShotsForDevice = function (backend, deviceId) {
+  if (! deviceId) {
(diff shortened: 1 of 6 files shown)


server/src/servershot.js
```diff
@@ -82,3 +82,44 @@ Shot.getRawValue = function (id) {
...
+    return db.select(
+      `SELECT data.id, data.value, data.deviceid
+       FROM data
+       WHERE data.deviceid IN (${idNums.join(", ")})
+       ORDER BY data.created
+      `,
+      ids
+    );
+  }).then((rows) => {
+    console.log("done", Date.now() - start);
```
fzzzy: Should either change this to console.info or remove the debug print.

ianb: Fixed in ed086af0080931a7af128bc92e54857e868aac13

PR #560, opened 2015-07-07
r? ianbicking Put in buttons to copy the screenshot and rich html in addition to the copy link button.

fzzzy:


addon/data/shoot-panel.js
```diff
@@ -89,6 +89,12 @@ class ShareButtons extends React.Component {
       <a href="#" onClick={ this.props.onCopyClick }>
         <img src={ `icons/link-${size}.png` } />
       </a>
+      <a href="#" onClick={ this.props.onCopyImageClick }>
```
ianb: This will probably update location.hash to be `#`, which probably isn't meaningful in this case, but generally I like to avoid that.  Fixing it either requires a change to `onCopyClick` or turning this into something other than an anchor (plus CSS)

addon/data/shoot-panel.js
```diff
@@ -135,6 +141,17 @@ class ShootPanel extends React.Component {
     self.port.emit("copyLink", this.props.shot.viewUrl);
```
ianb: I think the extra argument here is the bug, not below – below there is specific language about what is being copied to the clipboard, so if you passed a different link here that language would be incorrect.

addon/lib/shooter.js
```diff
@@ -296,14 +296,33 @@ const ShotContext = Class({
...
+        title: "Image Copied",
+        text: "Your shot has been copied to the clipboard.",
+        iconURL: self.data.url("../data/copy.png")
+      });
+    },
+    copyRich: function (url, img, title) {
+      // Use "text" instead of "html" so that pasting into a text area or text editor
+      // pastes the html instead of the plain text stripped out of the html.
+      clipboard.set(
+        `<a href="${url}"><img src="${img}" /><div>${title}</div></a>`, "text");
```
ianb: HTML escaping!  Clear XSS vulnerability here.

fzzzy: What do I use to properly escape?

ianb: There's no good function.  You have to do `text.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/"/g, "&quot;")` (or better put it in a function).

Also do we really want "text"?  We need to test a few scenarios:

- Pasting into Gmail (maybe also Yahoo mail or something)
- Pasting into some WYSIWYG, e.g., on WordPress (FCKEdit I think?)
- Pasting into Twitter and Facebook

fzzzy: I tried "html" as the clipboard type, but it didn't work at all. Only the visible text from the html was copied. I'll test it again a bit more with gmail, twitter, facebook, etc.

addon/data/shoot-panel.js
fzzzy: Ok, I added e.preventDefault() to all the anchor onClick event handlers.

addon/data/shoot-panel.js
fzzzy: Ok, I moved the property accesses out of shoot-panel and into shooter. Now the messages will always be consistent with what is copied.

addon/lib/shooter.js
fzzzy: I escaped the title -- let me know if you think we need to escape the viewUrl or the image url.

fzzzy: Ok, I added code that sets the clipboard html type, too. It doesn't seem to have any effect.

I can't seem to get facebook, twitter, or gmail to do anything other than insert the raw html into the textbox. If I set the clipboard html type but not the text type, then the only thing that is pasted is whatever visible text there would have been in that html -- which is not the right thing to paste at all.

PR #564, opened 2015-07-08
r? ianbicking Full page button

fzzzy:


server/src/clientglue.js
```diff
@@ -26,6 +26,32 @@ exports.setModel = function (data) {
...
+    let frameElement = document.getElementById("frame");
+    window.onscroll = function (e) {
+      if (e.pageY > 0) {
+        toolbar.style.visibility = "visible";
+        navigateToolbar.style.visibility = "visible";
+        let frameOffset = frameElement.getBoundingClientRect().top + window.scrollY;
+        let toolbarHeight = toolbar.clientHeight;
+        let visibleHeight = window.innerHeight - toolbarHeight;
+        let frameTop = frameOffset - toolbarHeight;
+        if (e.pageY >= frameTop - visibleHeight - 16) {
```
ianb: What does this 16 here mean?  Is it the height of one of the elements?  If so we could calculate it.

fzzzy: I guess it's 2/3rds of the height of the "Full Page" icon... The value was calculated using trial and error.

fzzzy: It's actually 1/2 the height of the element. I changed the code to calculate it.

PR #581, opened 2015-07-16
r? ianbicking Fixes #554 Transform data urls into http urls on the server and store images in new images table

### LINKED ISSUE #554: Move clip images into separate objects/tables
We want to move the data: urls out of the shot JSON, replacing them with other URLs.

This needs to:

- Put clip images into a different table
- If a clip is deleted, the clip image row should be deleted
- The client needs to know the new (non-data) URL of the clip

I'm thinking we could give a response to the PUT that perhaps is the entire updated shot, or maybe just changes.  Changes would be processed like:

```javascript
let updates = JSON.parse(putRequest.responseText);
let shot = <the original shot>;
function updateShot(shot, updates) {
  for (let key in updates) {
    if (updates[key] === null) {
      delete shot[key];
    } else {
      if (typeof updates[key] == "object") {
        if (key in shot) {
         updateShot(shot[key], updates[key]);
        } else {
          shot[key] = updates[key];
        }
      } else {
        shot[key] = updates[key];
      }
    }
  }
}
```

We could use this update process for the HTML keys, though we'd have to create either/or keys (probably like `bodyHtml` and `bodyHtmlUrl`).  Now that I write it down I regret the update I propose, because for instance with our strict testing in `AbstractShot` we probably require that both these can't be set simultaneously, and so doing property-by-property updates is inappropriate as there is a temporarily invalid state.  We could put this method in AbstractShot and tweak it with Shot-specific stuff.  Also `shot.bodyHtmlUrl = xxx` could automatically wipe `shot.bodyHtml` and vice versa.

fzzzy:


addon/lib/shooter.js
```diff
@@ -536,8 +536,12 @@ class Shot extends AbstractShot {
       url: url,
       content: body,
       contentType: "application/json",
-      onComplete: watchFunction(function (response) {
+      onComplete: watchFunction((response) => {
         if (response.status >= 200 && response.status < 300) {
+          for (let clipId in response.json.clips) {
+            let clip = this.getClip(clipId);
+            clip.image.url = response.json.clips[clipId].image.url;
```
ianb: Though we aren't creating them, we should still guard this with `if (clip.image)`

ianb: We also have to figure out what this means with respect to whether the clip has changed.  I think this isn't really an edit, but the shot will probably(?) report itself as being dirty as a result of this transformation.

I think the shot object itself should take care of this update.  E.g., `shot.updateClipUrls(response.json)`

server/src/dbschema.js
```diff
@@ -2,6 +2,12 @@ const db = require("./db");
 const Keygrip = require('keygrip');
 
 const createSQL = `
+
+CREATE TABLE IF NOT EXISTS images (
+  id VARCHAR(200) PRIMARY KEY,
+  image BYTEA
```
ianb: Could be NOT NULL I think?

Also, seems like it should have a reference to its shot, and ON DELETE CASCADE

server/src/linker.js
```diff
@@ -32,3 +32,13 @@ exports.staticLink = function (resource) {
 exports.staticLinkWithHost = function (req, resource) {
   return req.protocol + "://" + req.headers.host + exports.staticLink(resource);
 };
+
+exports.imageLink = function (urlBase, resource) {
+  if (resource.charAt(0) != "/") {
```
ianb: Why not startsWith?

server/src/server.js
```diff
@@ -161,8 +165,8 @@ app.put("/data/:id/:domain", function (req, res) {
       return shot.update();
     }
     return null;
-  }).then(() => {
-    simpleResponse(res, "Saved", 200);
+  }).then((ok) => {
+    simpleResponse(res, JSON.stringify(shot.asJson()), 200);
```
ianb: This feels a little like a lie: we're sending back the entire shot as JSON, implying this is the new representation, but the client is only looking at the clip URLs.

ianb: That is, I think I'd rather it just be a response like `{updateClipUrls: {clipId: {url: newUrl}}}` – since that's how it is being treated.

server/src/server.js
```diff
@@ -205,6 +209,20 @@ app.get("/content/:id/:domain", function (req, res) {
   });
 });
 
+app.get("/images/:id/:domain/:clipId", function (req, res) {
```
ianb: I am unsure whether we should use the shot id here or not.  There's a danger that you might want to share the image but keep the shot private (since it can easily have unintended extra information).  If you share the image by URL you've implicitly (but not obviously) exposed the shot it came from.

server/src/server.js
```diff
@@ -205,6 +209,20 @@ app.get("/content/:id/:domain", function (req, res) {
...
+  Shot.prototype.Clip.getRawBytesForClip(
```
ianb: Why is this attached to the prototype?  Especially since it doesn't seem to be a method.  I think it would be better just to add to Shot directly (i.e., `Shot.getRawBytesForClip(...)`)

server/src/server.js
```diff
@@ -205,6 +209,20 @@ app.get("/content/:id/:domain", function (req, res) {
...
+    req.params.id, req.params.domain, req.params.clipId
+  ).then((bytes) => {
+    if (bytes === null) {
+      simpleResponse(res, "Not Found", 404);
+    } else {
+      res.header("Content-Type", "image/png");
```
ianb: We need to store this content type, and not hard code it.  It's in the data URL.  In the future we might want to use jpeg for some clips.

server/src/servershot.js
```diff
@@ -8,24 +8,78 @@ class Shot extends AbstractShot {
...
-       VALUES ($1, $2, $3)`,
-      [this.id, this.ownerId, value]
+  convertAnyDataUrls(client, json, possibleClipsToInsert) {
+    return Promise.all(
+      possibleClipsToInsert.map((clipId) => {
+        let clip = this.getClip(clipId);
+        let imageId = `${this.id}/${clipId}`;
+        let data;
+        try {
+          data = clip.imageBinary();
```
ianb: I believe data include the content type.

server/src/servershot.js
```diff
@@ -8,24 +8,78 @@ class Shot extends AbstractShot {
...
+        } catch (e) {
+          if (e.message !== "Bad clip URL") {
+            throw e;
+          }
+          // It's already in the db, and the clip has an http url
+          return true;
```
ianb: This is way over-catching.  We should have a method like `clip.isHttpUrl()` or else test it directly here.  Also we need to account for text clips.

server/src/servershot.js
```diff
@@ -8,24 +8,78 @@ class Shot extends AbstractShot {
...
+        }
+        return db.upsertWithClient(
+          client,
+          "INSERT INTO images (image, id) SELECT $1, $2",
```
ianb: This doesn't seem to handle deletes

ianb:
That's the end of my comments.  A number of things to fix before merging.

server/src/server.js
fzzzy: It's attached to Clip, not Shot.prototype.

addon/lib/shooter.js
fzzzy: Right now the shot doesn't report itself as dirty because we are mutating the clip which does not have accessors, but I'll move the logic into a method on the shot.

server/src/dbschema.js
ianb:
```
CREATE TABLE IF NOT EXISTS images (
  id VARCHAR(200) PRIMARY KEY,
  shotid VARCHAR(200) NOT NULL REFERENCES shot (id) ON DELETE CASCADE,
  image BYTA NOT NULL,
  content_type TEXT NOT NULL
)
```

server/src/linker.js
fzzzy: Because I copied and pasted from code you wrote, otherwise I would have written !== ;-)

server/src/dbschema.js
fzzzy: We should only allow the user that created the shot to update a clip.

server/src/server.js
fzzzy: Sounds good. We can add updateBodyUrl once that is implemented

server/src/dbschema.js
ianb: find list of clipIds that haven't been updated, then run:

```
`DELETE FROM images
 WHERE shotid = $1 AND clipid NOT IN (${" ".join(clipCounts)})
`, [shotId].concat(clipIds)
```

fzzzy: Let's add a new random id, imageId, so that the clip and shot ids don't leak to users other than the creator.

server/src/server.js
```diff
@@ -49,6 +49,10 @@ app.use(function (req, res, next) {
 app.use(function (req, res, next) {
   req.staticLink = linker.staticLink;
   req.staticLinkWithHost = linker.staticLinkWithHost.bind(null, req);
+  if (Shot.prototype.imageLink === undefined) {
```
ianb: Not sure why this is going on the Shot prototype?  Also this implicitly means that the first request permanently defines the host for the life of the process.

server/src/server.js
```diff
@@ -205,6 +210,20 @@ app.get("/content/:id/:domain", function (req, res) {
   });
 });
 
+app.get("/images/:uid", function (req, res) {
```
ianb: What's uid?  Do you mean uuid?

shared/shot.js
```diff
@@ -888,6 +894,12 @@ class _Clip {
     this._image = image;
   }
 
+  isHttpUrl() {
```
ianb: This won't work for https.  Maybe ! clip.isDataUrl() is better

server/src/server.js
fzzzy: Yes, it's not supposed go on the prototype, whoops. (How is this working? I guess because the url just starts with / and not http?)

Yes, it does mean the first request defines the host for the life of the process. Do you have a better idea? The only other idea I had was to put it in a configuration file, which isn't the greatest.

server/src/server.js
fzzzy: No, I just use uid for an instance of a uuid

shared/shot.js
fzzzy: Yeah, let's make it ! clip.isDataUrl

shared/shot.js
```diff
@@ -554,6 +554,12 @@ ${options.addBody || ""}
     }
     return biggest;
   }
+  updateClipUrl(clipId, clipUrl) {
+    let clip = this.getClip(clipId);
+    if (clip && clip.image) {
```
ianb: Should probably be a warning if this isn't true, as this shouldn't happen

server/src/servershot.js
```diff
@@ -8,29 +9,95 @@ class Shot extends AbstractShot {
...
+        let clip = this.getClip(clipId);
+        if (clip.isHttpUrl()) {
+          // It's already in the db, and the clip has an http url
+          return false;
+        }
+        let uid = uuid.v4();
+        let data = clip.imageBinary();
+        return db.queryWithClient(
+          client,
+          "DELETE FROM images WHERE shotid = $1 AND clipid = $2",
```
ianb: I don't think this eliminates any fully deleted clips.

I'd do it like:

```javascript
let clips = shot.getClipNames().map((name) => shot.getClip(name));
let unedited = [];
let toInsert = [];
for (let clip of clips) {
  if (clip.image) {
    if (clip.image.url.startsWith("data:")) {
      toInsert.push(clip);
    } else {
      unedited.push(clip.id);
    }
  }
}
// Should go in db.js or something:
function markersForArgs(starting, numberOfArgs) {
  let result = [];
  for (var i=starting; i<starting+numberOfArgs; i++) {
    result.push("$" + i);
  }
  return result.join(", ");
}
return db.queryWithClient(
  client,
  `DELETE FROM images WHERE shotid = $1 AND clipid NOT IN (${markersForArgs(2, unedited.length)})`,
  [shot.id].concat(unedited)
).then(() => {
  return Promise.all(...)
});
```

fzzzy:
Thanks for all the great comments. I think I addressed them all again.

PR #586, opened 2015-07-17
r? ianbicking Move the shots index to the /shots url instead of /

fzzzy:


server/src/servershot.js
```diff
@@ -119,8 +119,12 @@ Shot.getShotsForDevice = function (backend, deviceId) {
     for (let i=0; i<rows.length; i++) {
       let row = rows[i];
       let json = JSON.parse(row.value);
-      let shot = new Shot(row.deviceid, backend, row.id, json);
-      result.push(shot);
+      if (json === null) {
```
ianb: When would this happen?  Also most problems with the JSON will be an exception, not a null return value.  I suppose the `"null"` literal might have gotten into the database in places?

PR #590, opened 2015-07-20
r? ianbicking Fixes #566 Implement the full page button in css instead of with an image.

### LINKED ISSUE #566: "Full Page" asset
I tried to make the "Full Page" icon from the mockups, but I was having trouble making the background around the edges transparent. Can you create this asset for me.

fzzzy:


server/src/views/frame.js
```diff
@@ -241,8 +242,11 @@ class Frame extends React.Component {
           { nextClipNode }
         </div>
         <div id="full-page-button">
-          <img src={ this.props.staticLink("img/full-page-button.png") }
-            onClick={ this.clickFullPageButton.bind(this) } />
+          <a href="#" className="full-page-button-styles" onClick={ this.clickFullPageButton.bind(this) }>
```
ianb: It occurs to me this could actually be `<a href="#frame">` or something similar to that, with no Javascript.

ianb: Though it also occurs that exactly `#frame` wouldn't be right, since we want a buffer.

fzzzy: Yeah, good point

fzzzy: Is that really what we want though? We do some math to get it just how we want it

ianb: Yes, that is what occurred to me after the first comment ;)

PR #615, opened 2015-07-28
r? ianbicking Move head and body out of the json

fzzzy:


server/src/linker.js
```diff
@@ -42,3 +42,13 @@ exports.imageLink = function (urlBase, resource) {
...
+
+exports.contentLink = function (urlBase, resource) {
+  if (! resource.startsWith("/")) {
+    resource = "/" + resource;
+  }
+  if (resource.startsWith("/content")) {
+    throw new Error("imageLink URL should not start with /content: " + resource);
+  }
+  return urlBase + "/content" + resource;
+};
```
ianb: Couldn't this be a method on Shot?

fzzzy: Why would we want it in a different place than all the other link functions?

server/src/servershot.js
```diff
@@ -100,12 +100,23 @@ class Shot extends AbstractShot {
         // If the key doesn't already exist, go through the clips being inserted and
         // check to see if we need to store any data: url encoded images
         return this.convertAnyDataUrls(client, json, possibleClipsToInsert).then((oks) => {
+          let contentId = uuid.v4();
+          let head = json.head;
+          let body = json.body;
+          json.head = "";
```
ianb: Why set them to empty string, instead of null or deleting the attributes?

server/src/servershot.js
```diff
@@ -100,12 +100,23 @@ class Shot extends AbstractShot {
...
+          json.body = "";
           return db.queryWithClient(
             client,
             `INSERT INTO data (id, deviceid, value)
              VALUES ($1, $2, $3)`,
             [this.id, this.ownerId, JSON.stringify(json)]
           ).then((rows) => {
```
ianb: I guess there shouldn't be `rows` parameters here or below, since it is an insert

server/src/servershot.js
```diff
@@ -126,6 +141,15 @@ class Shot extends AbstractShot {
...
             throw new Error("No row updated");
           }
+          return db.queryWithClient(
+            client,
+            "UPDATE content SET head = $1, body = $2 WHERE shotid = $3",
+            [head, body, this.id]
+          );
+        }).then((rowCount) => {
+          if (! rowCount) {
+            throw new Error("No row updated");
```
ianb: Some failure could cause this (e.g., something that causes the first insert to work without the second insert work).  It would be better to warn and insert.  But it also wouldn't really be a problem if it was one table.

ianb:
I think it would be better if the content was in the `data` table since if we put that change off until later we'd need to do yet another migration.

fzzzy:
Are we just going to drop the existing table then?

ianb:
We can manually fix it up, or try to setup migration (which probably isn't that hard to do).

fzzzy:
Ok, I think I addressed all the comments. Please look over this pr again.

server/src/linker.js
ianb: Things like viewUrl are already on the shot, and the shot already knows both the base and its name.

fzzzy: But the content base will be a different url, so should we pass the content base to the Shot too?

ianb: Maybe?  With this function I'd worry about passing in the wrong urlBase in some place, while the Shot if instantiated correctly should be okay.  Also I suppose the Shot could take the content base URL directly from configuration?  Since it's not dynamically determined from the request at all.

fzzzy: Ok, I'll try having the Shot take it from the config. That seems cleaner.

server/src/servershot.js
```diff
@@ -100,12 +100,16 @@ class Shot extends AbstractShot {
...
           return db.queryWithClient(
             client,
-            `INSERT INTO data (id, deviceid, value)
-             VALUES ($1, $2, $3)`,
-            [this.id, this.ownerId, JSON.stringify(json)]
-          ).then((rows) => {
+            `INSERT INTO data (id, deviceid, value, head, body)
+             VALUES ($1, $2, $3, $4, $5)`,
+            [this.id, this.ownerId, JSON.stringify(json), head, body]
+          ).then((rowCount) => {
```
ianb: Should we be updating the client and telling it not to keep sending the head and body (similar to what we do with clips)?  And then we'd only overwrite head and body if the json head/body weren't null?

fzzzy: Yes.

PR #628, opened 2015-07-29
r? ianbicking Fixes #556 Move /content/ to a different origin, use CSP

### LINKED ISSUE #556: Move /content/ to a different origin, use CSP
Given a separate origin (ideally domain, maybe TLD) for the content frame, and a Content Security Policy restricting the JS locations, we could avoid a number of security problems.

fzzzy:


server/src/clientglue.js
```diff
@@ -1,4 +1,5 @@
-/* jshint browser:true */
+/* jslint browser:true */
+/* global POST_MESSAGE_ORIGIN */
```
ianb: This might be clearer as `CONTENT_HOSTING_ORIGIN` (I was going to say `CONTENT_ORIGIN` but that feels a bit like the source URL origin)

server/src/config.js
```diff
@@ -13,10 +13,17 @@ let conf = convict({
...
     arg: "port"
   },
+  host: {
+    doc: "The server host",
+    format: String,
+    default: "localhost",
+    env: "HOST",
+    arg: "host"
+  },
   contentPort: {
```
ianb: Are we expecting to run two ports in production?  We could also do Host-header matching in that case.  Mostly because it makes the nginx configuration a bit easier, which isn't actually that challenging (except with ports it means nginx has to have the same domain mappings that the configuration will have).

server/src/server.js
```diff
@@ -371,6 +352,41 @@ app.use(function (err, req, res, next) {
   errorResponse(res, "General error:", err);
 });
 
+const contentApp = express();
+
+contentApp.use("/static", express.static(path.join(__dirname, "static"), {
+  index: false
+}));
+
+contentApp.use(function (req, res, next) {
```
ianb: I think this is copied from elsewhere, right?  Could we make it into a named function to be reused?

server/src/static/js/content-helper.js
```diff
@@ -1,19 +1,20 @@
 /*jslint browser: true */
+/* global POST_MESSAGE_ORIGIN */
```
ianb: Here `POST_MESSAGE_ORIGIN` is actually something like `SITE_ORIGIN`, right?  That is, it should be `https://pageshot.dev.mozaws.net`, where above it would be `https://pageshotcontent.dev.mozaws.net`

server/src/server.js
fzzzy: It is copied, but it's not identical, so I didn't make a function.

server/src/static/js/parent-helper.js
```diff
@@ -2,17 +2,29 @@
...
 }
 
 window.onmessage = function(m) {
-  if (m.origin !== location.origin) {
-    console.warn("Parent iframe received message from unexpected origin:", m.origin, "instead of", location.origin);
+  if (m.origin !== POST_MESSAGE_ORIGIN) {
+    console.warn("Parent iframe received message from unexpected origin:", m.origin, "instead of", POST_MESSAGE_ORIGIN);
     return;
   }
+  if (resolveChildReference) {
```
ianb: Hmm... the way I've done this is with a queue/buffer, which I think is a bit simpler.  So it's something like:

```javascript
function sendToChild(message) {
  if (! sendToChild.childReference) {
    sendToChild.queue.push(message);
    return;
  }
  ... normal postMessage...
}
sendToChild.queue = [];
```

and then we just clear the queue when we get the childReference.  Then code that wants to send a message doesn't have to have any special logic.  (We'll be sending more messages soon, if we show all the clips, show more content inline, allow editing of comments inline, etc)

server/src/server.js
ianb: OK

server/src/config.js
fzzzy: What does host header matching look like in express? I think the nginx configuration is going to be pretty much identical, and the port seems like clearer separation to me, but I don't care too much.

server/src/static/js/content-helper.js
fzzzy: Yes.

server/src/static/js/parent-helper.js
fzzzy: If there is going to be more than one message, sure, but shouldn't we do that work when there is more than one message to send? Seems premature to do it now.

ianb:
Those are all my comments; really just clarifying the naming of POST_MESSAGE_ORIGIN, and I'm a little unsure about the post() function and the promise, seems like in the future it might be too easy to mis-call it.  (The bugs that will cause are intermittent since it's based on the timing of the loads, so I want to avoid any misuse if at all possible)

server/src/static/js/parent-helper.js
ianb: I don't think it's any more code, just a clearer function that encapsulates a bit more of the complexity of postMessage

server/src/config.js
ianb: Yeah, on reflection I think it's fine, the extra configuration isn't a big deal.

PR #641, opened 2015-07-31
r? ianbicking Fixes #446 Implement schema migrations with pg-patcher

### LINKED ISSUE #446: Schema migrations

fzzzy:


server/src/dbschema.js
```diff
@@ -94,6 +105,17 @@ exports.createTables = function () {
         if (count != 1) {
           throw new Error("Should have deleted one row");
         }
+        return db.getConnection().then(([conn, done]) => {
```
ianb: We'd avoid this problem if we just ran pgpatcher first, wouldn't we?

server/src/dbschema.js
```diff
@@ -94,6 +105,17 @@ exports.createTables = function () {
...
+          pgpatcher(conn, 2, function(err) {
```
ianb: The 2 here, and the 2 on line 15 could be a single variable

server/src/dbschema.js
fzzzy: What problem?

ianb: The problem where we aren't sure if we are inserting into a table that has a head and body

server/src/dbschema.js
```diff
@@ -1,8 +1,19 @@
 const db = require("./db");
 const Keygrip = require('keygrip');
+const pgpatcher = require("pg-patcher");
+
+// MAKE SURE TO SET THE PATCH LEVEL TO THE HIGHEST LEVEL
+// IN THE STATEMENT INSERT INTO property(key, value) VALUES('patch', 2);
 
 const createSQL = `
```
ianb: I'm pretty sure we can't always unconditionally try to run createSQL - the challenges of creating indexes, for instance.  Also won't we unconditionally run `INSERT INTO property(key, value) VALUES('patch', 2);` ?

server/src/dbschema.js
fzzzy: Well inserting a row without a head and a body also works because head and body allow nulls at the moment, and we do need the tables to be created first before running the patcher. It does seem to make sense to run the sanity test to see if the tables were created properly before running the patcher.

server/src/dbschema.js
fzzzy: Hmm, yes, I added the INSERT INTO at the end without thinking about it too much...

fzzzy:
Ian has better ideas about how to structure this.

PR #642, opened 2015-07-31
r?fzzzy Ga homepage

ianb:
Not 100% finished, but wanted to get it committed before doing db stuff

Also some code duplication between the homepage and shell.js that I don't know how to avoid, might have to just suck it up.


server/src/views/shell.js
```diff
@@ -13,26 +13,36 @@ export class Shell extends React.Component {
...
+         ga("create", "${this.props.gaId}", gaOptions);
          ga("send", "pageview");
        })();
       `;
       gaScript = <script src="//www.google-analytics.com/analytics.js" async="1"></script>;
-      gaCode = <script dangerouslySetInnerHTML={{__html: gaJs}}></script>;
+    } else {
+      gaJs = `
+      window.ga = function () {
+        console.info.apply(console, ["stubbed ga("].concat(arguments).concat([")"]));
```
fzzzy: You could use console.info(...["stubbed ga("].concat(arguments).concat([")"])) here but it's not important.

ianb: True; can't use that in homepage.html though (no es6 translation over there)

PR #665, opened 2015-08-06
r?fzzzy Addon signed

ianb:


addon/data/shoot-panel.js
```diff
@@ -346,7 +346,7 @@ class ShootPanel extends React.Component {
 
   addClip() {
     isAdding[this.props.shot.id] = this.props.shot.clipNames().length;
-    setTimeout(renderData);
+    setTimeout(() => renderData);
```
fzzzy: Isn't this going to return the renderData function, doing nothing? I think you need to call renderData()

gulpfile.js
```diff
@@ -100,6 +100,11 @@ gulp.task("static-addon", function () {
   return gulp.src(["addon/**/*.{html,css,png,svg}", "addon/run", "addon/package.json"]).pipe(gulp.dest("dist/addon"));
 });
 
+/** Remove references to Function.prototype, which triggers addon warnings */
+function removeFunctionPrototype(content) {
```
fzzzy: This is commented out below, is it needed?

ianb:
Yoinks, the setTimeout changes would have led to some very annoying bugs.  Fixed, and some other build tweaks.

PR #672, opened 2015-08-06
r? ianbicking Shot expiry

fzzzy:


server/db-patches/patch-2-3.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE data ADD url TEXT;
```
ianb: Maybe this should be NOT NULL?  Of course the UPDATE below has to run first, but it could then have `ALTER TABLE data ALTER COLUMN url SET NOT NULL` (I think)

ianb:
Just that constraint issue; fix and push?

server/db-patches/patch-2-3.sql
fzzzy: How could the update run before the table has a url column?

fzzzy: I think it would work if we did DEFAULT "", is that ok?

fzzzy: Ok, I understand. Create the column first, do the update, and then alter the column to set not null. Cool.

fzzzy: ALTER TABLE data ALTER COLUMN url SET NOT NULL;

PR #682, opened 2015-08-11
r? ianbicking Fix data url bug

fzzzy:


bin/run-addon
```diff
@@ -115,7 +115,7 @@ if [[ -z "$no_auto" ]] ; then
 fi
 
 cd build/addon
-jpm run --binary $binary --profile ../../Profile --no-copy \
+jpm run --binary $binary --profile ../../dist/Profile --no-copy \
```
ianb: Since we're putting the build files in `build/` now (which arguably is an unnecessary rename), I was presuming that we'd move profiles around to eliminate the dist directory.

server/src/servershot.js
```diff
@@ -18,8 +18,8 @@ class Shot extends AbstractShot {
     let toInsert = [];
     for (let clip of clips) {
       if (clip.image) {
-        if (! clip.isDataUrl()) {
-          toInsert.push(clip);
+        if (clip.isDataUrl()) {
```
ianb: doh :-/

bin/run-addon
fzzzy: Is the dist directory in my local checkout vestigial then? Where is the correct location of the new profile?

ianb: It is vestigal, yes.  I moved it to ./Profile in the root of the project (no particular reason for that location, but I didn't want it in build/ because then make clean would wipe it)

fzzzy: Well there's a bug then, because no profile gets created at ./Profile...

ianb: I probably never tried it from scratch, and maybe jpm doesn't work well if the profile doesn't already exist (I now vaguely recall that `--profile directory` doesn't work with jpm unless `directory` exists already).

server/src/servershot.js
```diff
@@ -18,8 +18,8 @@ class Shot extends AbstractShot {
...
+          toInsert.push(clip.id);
```
fzzzy: Plus this made for double doh (made figuring out what the problem was extremely difficult)

PR #688, opened 2015-08-13
r? ianbicking Delete account

fzzzy:


server/src/servershot.js
```diff
@@ -311,3 +311,29 @@ Shot.getShotsForDevice = function (backend, deviceId) {
...
+      ids.push(rows[i].id);
+    }
+    let deleteSql = `DELETE FROM devices WHERE
+     id IN (${db.markersForArgs(1, ids.length)})`;
+    return db.update(
+      deleteSql,
+      ids
+    );
+
+  }).catch((e) => {
```
ianb: The catch handler here should be in server.js

server/src/views/profile.js
```diff
@@ -109,12 +110,23 @@ exports.Profile = class Profile extends React.Component {
     this.setState({ nickname: e.target.value });
   }
 
+  onClickDeleteEverything(e) {
+    e.preventDefault();
+    let ok = confirm("Are you sure you want to permanently delete everything you have created using PageShot?");
+    if (ok) {
+      events.deleteEverything();
+      helpers.request("POST", "/delete");
+      window.location = "/";
```
ianb: We should probably redirect to something like `/?accountDeleted` and then do a check for that and pop up something on the homepage (which we can do separately from this commit)

ianb:
Just those two small comments

PR #692, opened 2015-08-19
r? ianbicking First run intro

fzzzy:


addon/lib/main.js
```diff
@@ -231,11 +248,42 @@ exports.getBackend = function () {
...
+  }
+}
+
 // For reasons see https://developer.mozilla.org/en-US/Add-ons/SDK/Tutorials/Listening_for_load_and_unload
 exports.main = function (options) {
+  loadReason = options.loadReason;
+  if (options.loadReason === "install") {
+    let helpurl = exports.getBackend() + "/homepage/help.html";
+    let win = winutil.getMostRecentBrowserWindow();
+    win.loadURI(helpurl);
```
ianb: is there a reason to use this instead of `require("sdk/tabs").open(helpurl)` ?

addon/lib/recall.js
```diff
@@ -6,12 +6,16 @@ const ss = require("sdk/simple-storage");
 const clipboard = require("sdk/clipboard");
 const { AbstractShot } = require("./shared/shot");
 
+let showInfoPanel = null;
```
ianb: showInfoPanel appears to be unused?

static/css/introjs.scss
```diff
@@ -0,0 +1,347 @@
+.introjs-overlay {
```
ianb: I think it would be better if these were in `static/vendor/introjs/...`

Also, we should copy this over as CSS and not do SCSS on it (assuming it is actually just plain CSS)

ianb:
That's all my comments

addon/lib/main.js
fzzzy: Yes. Opening a new tab caused issues. I can't remember the exact details at the moment. I should have taken notes.

addon/lib/recall.js
fzzzy: Yes, I thought I would keep it there just for completeness because I injected hideInfoPanel too, but I'll take it out.

static/css/introjs.scss
fzzzy: The static file story is getting messy. I changed the filetype to scss because plain css files are not copied over in the static dir. Is it ok if we open a follow-up bug to clean up the static directories? The one at server/src/static should be merged with the top level static dir, I think.

ianb: Sure.  There is a special rule for addon/data/vendor that just copies everything with no transformation, which we could use for static/vendor too.

fzzzy: Filed #697 to follow up

PR #699, opened 2015-08-20
r?fzzzy Fix 677

ianb:


fzzzy:
It looks good, but it needs to be merged with master.

PR #739, opened 2015-09-15
r?fzzzy Export

ianb:


server/src/reactutils.js
```diff
@@ -1,12 +1,26 @@
...
+  </script>
+  ${footer}`
+    );
+  } else {
+    return `<!DOCTYPE html>\n${body}`;
+  }
+};
+
+exports.extend = function (...objs) {
+  let obj = {};
```
fzzzy: es6 has Object.assign, which you could use like this:

return Object.assign({}, ...objs);

server/src/server.js
```diff
@@ -255,9 +263,14 @@ app.get("/shots", function (req, res) {
   if (! req.deviceId) {
     return simpleResponse(res, "You must have the addon installed to see your shot index", 403);
   }
+  let simple = shouldRenderSimple(req);
```
fzzzy: Why bother calling this here? I think you should do it inside the if on line 269

server/src/server.js
```diff
@@ -275,15 +288,23 @@ app.post("/delete", function (req, res) {
   });
 });
 
+// FIXME: this can't the right way to do this...
+require("./exporter").setup(app);
+
 app.get("/:id/:domain", function (req, res) {
   let shotId = req.params.id + "/" + req.params.domain;
+  let simple = shouldRenderSimple(req);
```
fzzzy: Same here, why not just put it on line 303

server/src/exporter.js
```diff
@@ -0,0 +1,255 @@
...
+      });
+    }
+  });
+};
+
+exports.setup = function (app) {
+
+  app.get("/export", function (req, res) {
+    if (! req.deviceId) {
+      return require("./server").simpleResponse(res, "You must have the addon installed to export your shots", 403);
```
fzzzy: Should move simpleResponse and errorResponse into another module that both exporter.js and server.js depend on, to avoid circular imports.

server/src/exporter.js
```diff
@@ -0,0 +1,255 @@
...
+        }
+      }
+    );
+    subprocess.on("close", function (code) {
+      console.info("Exported to", dir, "in", (Date.now() - start) + "ms", "exit code:", code);
+    });
+    subprocess.on("error", function (e) {
+      console.warn("Error launching wget:", e);
+    });
+    resolve();
```
fzzzy: Yes, this is synchronous, don't you want the promise to resolve inside subprocess.on("close", ...) instead? Should also reject inside on("error", ...)

server/src/exporter.js
```diff
@@ -0,0 +1,255 @@
...
+    // Also maybe this is all synchronous
+  });
+}
+
+function getWgetStatus(deviceId) {
+  return new Promise((resolve, reject) => {
+    let dir = exportPath(deviceId);
+    let result = {errors: []};
+    let steps = 3;
+    function done() {
```
fzzzy: done seems like a bad name, maybe resolveIfDone?

server/src/exporter.js
ianb: I wonder if Express has a pattern that would obsolete those two functions?

server/src/exporter.js
```diff
@@ -0,0 +1,255 @@
...
+      return require("./server").simpleResponse(res, "You must have the addon installed to export your shots", 403);
+    }
+    require("./views/export").render(req, res);
+  });
+
+  app.post("/export", function (req, res) {
+    if (! req.deviceId) {
+      return require("./server").simpleResponse(res, "You must have the addon installed to export your shots", 403);
+    }
+    launchWget(req.deviceId).then(() => {
```
fzzzy: On second thought, since launchWget is async and nothing cares about it's completion, I think you can just make launchWget not return a promise at all but just fire off the async subprocess. The code that calls getWgetStatus will then take care of monitoring for the subprocess' completion.

server/src/exporter.js
```diff
@@ -0,0 +1,255 @@
...
+          keepTime: keepTime
+        };
+      }
+      steps--;
+      done();
+    });
+  });
+}
+
+exports.cleanExports = function () {
```
ianb: It occurred to me this function might be simpler if it called `getWgetStatus()`

Also maybe it should be a script and not written in Node, making it wildly simpler.

fzzzy: Yeah, this code might be a lot nicer if it wasn't in node.

server/src/reactutils.js
ianb: Can we actually use this?  I tried:

```
$ echo 'Object.assign({}, a, b, c)' | ./node_modules/.bin/babel 
"use strict";

Object.assign({}, a, b, c);
```

fzzzy: It is implemented in a polyfill in babel, core.js

    https://github.com/zloirock/core-js#ecmascript-6-object

server/src/exporter.js
ianb: I don't think I'm going to fix it for now.  I tested this code and got it working, in the future when it's not working we should strongly consider rewriting it in Python or something.

fzzzy: sounds good.

ianb:
I think this pretty much covers it

PR #749, opened 2015-09-17
r? ianbicking Auto detect warning

fzzzy:


addon/data/shooter-interactive-worker.js
```diff
@@ -863,7 +872,8 @@ function autoSelect(ids) {
   }
   if (pos.top === null && pos.bottom === null &&
       pos.left === null && pos.right === null) {
-    console.info("No autoSelect elements found, doing no selection");
+    console.warn("Auto-detect selection failed: No suitable element found");
+    showHelpMessage("Auto-detect selection failed. Selected visible area.");
```
ianb: This message feels very internal, not based on user expectation.  Maybe for now open a ticket to follow up with Bram?

fzzzy: #750

PR #753, opened 2015-09-23
r? ianbicking Universal analytics

fzzzy:
Currently tracks on the server side:

login
unload
clip views, logged in and anonymous
raw clip image views, logged in and anonymous

The client still sends analytics for the shot view, shot list, and
homepage.

### DIFF
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -160,11 +160,12 @@ app.get("/clip/:id/:domain/:clipId", function (req, res) {
       return;
     }
     let image = clip.imageBinary();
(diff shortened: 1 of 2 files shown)


server/src/server.js
```diff
@@ -151,6 +160,12 @@ app.get("/clip/:id/:domain/:clipId", function (req, res) {
       return;
     }
     let image = clip.imageBinary();
+    if (req.userAnalytics) {
+      req.userAnalytics.pageview(`/clip/${req.params.id}/${req.params.domain}/${req.params.clipId}`).send();
```
ianb: While I don't think there's a real security issue here, on general principle these should be URL-quoted (i.e., `${encodeURIComponent(req.params.id)}`)

fzzzy:
Added encodeURIComponent calls.

PR #755, opened 2015-09-24
r? ianbicking Auto select help message

fzzzy:
Also, wait half a second before notifying the user, to prevent the help
message from showing up in the screenshot.

### DIFF
--- a/addon/data/shooter-interactive-worker.js
+++ b/addon/data/shooter-interactive-worker.js
@@ -142,6 +142,7 @@ function reportSelection(captureType) {
 
 function showHelpMessage(message) {
   helpMessage = document.createElement("div");
+  helpMessage.className = "pageshot-hide-selection pageshot-help-message";
   helpMessage.id = "help-overlay";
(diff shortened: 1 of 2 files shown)


addon/data/shooter-interactive-worker.js
```diff
@@ -873,7 +873,9 @@ function autoSelect(ids) {
   if (pos.top === null && pos.bottom === null &&
       pos.left === null && pos.right === null) {
     console.warn("Auto-detect selection failed: No suitable element found");
-    showHelpMessage("Auto-detect selection failed. Selected visible area.");
+    setTimeout(function() {
+      showHelpMessage("Cannot detect selection automatically.");
+    }, 500);
```
ianb: In inline-selection.scss, there's a set of rules for `.pageshot-hide-selection`, a class that is added just before screenshotting, and causes any interface elements to be hidden.  So if the help message gets a class and some rules, it won't need this delay.

fzzzy:
Switched to using pageshot-hide-selection instead of setTimeout.

PR #756, opened 2015-09-24
r? ianbicking track exceptions

fzzzy:
ga limits the description to 150 characters.

### DIFF
--- a/addon/lib/errors.js
+++ b/addon/lib/errors.js
@@ -45,7 +45,8 @@ exports.unhandled = function (error) {
   req.request(`${main.getBackend()}/error`, {
     method: "POST",
     content: JSON.stringify(errorObj),
-    contentType: "application/json"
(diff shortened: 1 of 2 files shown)


addon/lib/errors.js
```diff
@@ -37,6 +39,14 @@ exports.unhandled = function (error) {
...
   }
   panel.port.emit("showError", errorObj);
+  if (typeof errorObj === "string") {
+    errorObj = {name: errorObj};
+  }
+  req.request(`${main.getBackend()}/error`, {
+    method: "POST",
+    content: JSON.stringify(errorObj),
+    contentType: "application/json"
+  });
```
ianb: A login shouldn't be required here, so `ignoreLogin: true` should be on here

ianb:
Looks good with that one note

PR #759, opened 2015-09-25
r? ianbicking Track clip changes

fzzzy:


fzzzy:
Check the dashboard to see if what this sends is useful or not. Let's also track clicking on the login button and on the full page button, and make sure we can tell the difference.

fzzzy:
We might want to hash the random part of the url so that we can track accesses to urls without knowing exactly what url it was for privacy reasons.

ianb:
We're going to start again from the top

PR #760, opened 2015-09-25
r? ianbicking Hide referer

fzzzy:


ianb:
This is a good first step, but we'll have to handle all the links in the content too.  We should set the `<meta name="referrer">` tag on the main and content iframe.  For now I think we have to leave off doing redirects inside the content iframe.

PR #761, opened 2015-09-29
r? ianbicking Render react head and body, not document

fzzzy:


server/src/views/frame.js
```diff
@@ -430,16 +482,21 @@ exports.render = function (req, res) {
   };
   let body = React.renderToString(frame);
   let json = JSON.stringify(clientPayload);
-  body = addReactScripts(body, `
+  let result = addReactScripts(
+    `<html>
+  ` + headString + `
```
ianb: Shouldn't this just use `${headString}` ?

server/src/views/frame.js
```diff
@@ -454,8 +511,13 @@ exports.renderSimple = function (req, res) {
...
     simple: true,
     shotDomain: req.url // FIXME: should be a property of the shot
-  });
+  };
+  let headString = React.renderToStaticMarkup(HeadFactory(serverPayload));
+  let frame = FrameFactory(serverPayload);
   let body = React.renderToStaticMarkup(frame);
-  body = '<!DOCTYPE HTML>\n' + body;
+  body = `<!DOCTYPE HTML>
+  ` + headString + `
```
ianb: here too

ianb:
Minor nit on the two string interpolations, otherwise good to go

server/src/views/frame.js
fzzzy: Hah. Of course. Doh.

PR #764, opened 2015-10-01
r? ianbicking Unify copy buttons

fzzzy:


addon/lib/shooter.js
```diff
@@ -142,11 +143,59 @@ const ShotContext = Class({
...
+    // pastes the html instead of the plain text stripped out of the html.
+    let clip = this.shot.getClip(activeClipName);
+    if (clip === undefined) {
+      let names = this.shot.clipNames();
+      if (names.length === 0) {
+        // Copy just the raw url in case for some reason no clips ever become available
+        // But don't send a notification yet, otherwise the user will get two notifications
+        // in a row if the rich shot data is successfully copied to the clipboard later.
+        clipboard.set(this.shot.viewUrl, "text");
+        if (numberOfTries === undefined) {
```
ianb: Is this an undeclared global?  It could just go on `this`

addon/lib/shooter.js
```diff
@@ -142,11 +143,59 @@ const ShotContext = Class({
...
+    }
+    let url = this.shot.viewUrl;
+    let img = clip.image.url;
+    let title = this.shot.title;
+    let origin = new URL(this.shot.url).hostname;
+    let html = (
+`<div>
+  <a href="${escapeForHTML(url)}">
+    <div>${escapeForHTML(title)}</div>
+    <img src="${img}" />
```
ianb: This and ${origin} should also be escaped.  Though probably okay, me not doing this caused an XSS bug elsewhere.

addon/lib/shooter.js
fzzzy: No, it is a parameter to the copyRichDataToClipboard function.

PR #769, opened 2015-10-05
r? ianbicking Fix signing errors

fzzzy:


Makefile
```diff
@@ -132,6 +132,10 @@ build/addon/data/panel-bundle.js: $(shoot_panel_dependencies)
 	@mkdir -p $(@D)
 	# Save the bundle dependencies:
 	browserify --list -e ./build/addon/data/shoot-panel.js | sed "s!$(shell pwd)/!!g" > build/shoot-panel-dependencies.txt
+	cat ./build/addon/lib/shooter.js | ./bin/_fixup_panel_js > ./build/addon/lib/shooter-new.js
+	mv ./build/addon/lib/shooter-new.js ./build/addon/lib/shooter.js
```
ianb: This feels fragile to me.  I think something like:

```
build/addon/lib/shooter.js: addon/lib/shooter.js
    @mkdir -p $(@D)
    $(BABEL) < $< | ./bin/_fixup_panel_js > $@
```

I think that will handle it well.  *Maybe* that should be a general rule for things in `build/addon/data/%.js`

ianb:
That's my only comment

Makefile
fzzzy: Another thing to think about related to this is that I don't think shooter.js and shoot-panel.js need to be in the xpi at all -- they are included in the output of browserify.

ianb: addon/lib/shooter.js isn't, it's an addon module.

You are correct about shoot-panel.js – another application of -prune should fix that one.

fzzzy: Ok, right.

PR #799, opened 2015-10-14
r? ianbicking Better tutorial

fzzzy:


addon/lib/helperworker.js
```diff
@@ -15,9 +15,12 @@ var notifications = require("sdk/notifications");
 var { XMLHttpRequest } = require("sdk/net/xhr");
 const { watchFunction, watchWorker } = require("./errors");
 const user = require("./user");
+const {Cu} = require("chrome");
 
 var existing;
 
+Cu.import("resource:///modules/UITour.jsm");
```
ianb: There's also the two-argument form of Cu, I think like `let UITour = {}; Cu.import("...", UITour); {UITour} = UITour;` – a little awkward, but explicit.

fzzzy: Yes, that sounds like a better thing to do.

addon/lib/helperworker.js
```diff
@@ -81,6 +84,20 @@ function resetPageMod(backend) {
         user.deleteEverything();
       }));
 
+      worker.port.on("contentTourComplete", watchFunction(function () {
+        const winutil = require("sdk/window/utils");
+        let win = winutil.getMostRecentBrowserWindow();
+        let magicCookie = "toggle-button--jid1-neeaf3sahdkhpajetpack-pageshot-recall";
+        let target = {
+          node: win.document.getElementById(magicCookie),
```
ianb: I don't think this is e10s compatible?  Maybe it works via the cross-process object wrapper, but we should avoid writing code that uses CPOWs.  (This might be signalling a warning?)

ianb: Nevermind, the node is a XUL node

PR #810, opened 2015-10-20
r? ianbicking Attempt to fix intro.js in production by not performing react rendering while intro.js is running.

fzzzy:


ianb:
Downside I realize looking at the code: if IntroJS fails in any way, React will never update until the user reloads the page.  Bit of a bummer.

PR #833, opened 2015-10-23
r? ianbicking Hide full page thumbnail

fzzzy:


ianb:
There's probably a followup bug to reenable the thumbnail, that should contain a reference to commit 68d0261

PR #853, opened 2015-11-04
r? ianbicking Change event categories

fzzzy:


ianb:
These aren't using labels, just event action but I'm inclined to just see how these look in GA and go from there.  I'm not sure if there's any reason to group all "clicks" together (for instance), so flattening seems appropriate.

ianb:
It seems to have a conflict: merge after resolving?

PR #854, opened 2015-11-09
r? ianbicking Fix two full page buttons

fzzzy:


server/src/clientglue.js
```diff
@@ -22,30 +22,24 @@ exports.setModel = function (data) {
...
-    let frameElement = document.getElementById("frame");
-    let offset = (fullPageButtonScrollable.clientHeight / 2);
+    let offset = document.getElementById("full-page-button-scrollable").clientHeight / 2;
+    document.body.className = "window-scrolled-up";
     window.onscroll = function (e) {
       if (e.pageY > 40) {
-        toolbar.style.visibility = "visible";
+        let toolbar = document.getElementById("toolbar");
+        let frameElement = document.getElementById("frame");
+        document.body.className = "window-scrolled-down";
```
ianb: If we do any other classes on the body this will wipe them out.  I think [classList](http://caniuse.com/#feat=classlist) should be okay for us to use – I'd be surprised if we work on IE9 right now, though maybe we do.

ianb:
Just the one comment, then merge

PR #885, opened 2015-12-01
r? ianbicking Containerize pageshot with docker and elastic beanstalk

fzzzy:


server/src/server.js
```diff
@@ -40,6 +40,10 @@ app.use("/static", express.static(path.join(__dirname, "static"), {
   index: false
 }));
 
+let xpidir = path.join(__dirname, "..", "xpi");
+console.log("xpidir", xpidir);
```
ianb: Oops, console.log

server/src/dbschema.js
```diff
@@ -8,6 +8,7 @@ const MAX_DB_LEVEL = 6;
 /** Create all the tables */
 exports.createTables = function () {
   console.info("Setting up tables on", db.constr);
+  console.info("Environment: ", process.env);
```
ianb: Is there a reason for the whole environment to be in here?  Could potentially lead to leaking private keys into the logs, for instance.

ianb:
Just wondering about those two log messages, otherwise r+/commitable

fzzzy:
Removed debug logging.

PR #886, opened 2015-12-01
r? ianbicking Fix show tour

fzzzy:


addon/data/shoot-panel.js
```diff
@@ -173,7 +173,7 @@ class ShootPanel extends React.Component {
   }
 
   onLinkClick(e) {
-    self.port.emit("openLink", this.props.shot.viewUrl, "normal");
+    self.port.emit("openLink", this.props.shot.viewUrl, this.props.loadReason, "normal");
```
ianb: `"normal"` is the eventSource parameter, isn't it?  Maybe it should be "addon"?

ianb: Actually I'm confused about the parameters to openLink, and the code doesn't make it clear to me either...

fzzzy: I think the problem is that there are two openLink functions with two different signatures, one in the panel and one in the recall panel.

PR #887, opened 2015-12-01
r? ianbicking Disable export

fzzzy:


server/src/server.js
```diff
@@ -73,6 +73,11 @@ app.use(function (req, res, next) {
   next();
 });
 
+app.use(function (req, res, next) {
+  req.allowExport = config.allowExport;
```
ianb: It should be possible to use `req.config.allowExport`, and not copy the variable over

ianb:
Get rid of `req.allowExport` and just use `req.config` and then r+/commit

fzzzy:
Switched to req.config.allowExport.

PR #912, opened 2015-12-18
Add headless HTTP server that listens for requests and processes pages without X

ben-en:
This pull request adds httpd as a jsm, an autoShot function to the bottom of shooter.js, and a headless module which handles initializing the httpd module as well as defining the listener.

From the documentation:
```
/* Headless server, for use with Pageshot:
 *  https://github.com/mozilla-services/pageshot
 * Relies on the server module from mozilla testing:
 *  http://mxr.mozilla.org/mozilla-central/source/netwerk/test/httpserver/
 *
 * Listening IP address may be assigned to httpd, by default it is localhost. 
 * The default is used in this implementation.
 *    For more info:
 *    http://mxr.mozilla.org/mozilla-central/source/netwerk/test/httpserver/nsIHttpServer.idl#282
 *
 * TODO:
 *    Documentation
 *    Check URL before opening
 *    Fix autoShot - currently it leaves some errors about attrs
 *    Add preference for whether or not headless should run
 * 
 * General usage: 
 * When you make an HTTP request to this server on the appropriate port it will
 * look for a query string and attempt to open the string as a URL, return a
 * URL of where to find it when processing is done, and process the tab with
 * an automatic shot function.
 *
 * Generally CSS parsing is enabled on this branch, which is EXPERIMENTAL.
 *
 * Syntax, assuming default IP, PORT, and target URL of asdf.com:
 *    curl 'http://localhost:10082/?http://www.asdf.com' 
 *
 * Usage without X:
 * To use this without X, start the server as normal. Then, in the bin/
 * directory, run `sh ephemeral-x.sh ./run-addon`.
 */
```

Any feedback regarding use cases for pageshot or implementation details is welcome. For now i believe i will be running this off of the outernet-project repository and pulling occasionally from upstream as it accomplishes what i need it to, more or less. 

I also put together a function for (ab)using firefox's readable mode. See https://github.com/Outernet-Project/pageshot/compare/headless...Outernet-Project:readable-headless for details


ianb:
Looks pretty good!

@fzzzy is going to look closer at this, but first thing is it should be turned off by default with a pref.

To make a new pref look at `addon/package.json` and add another boolean pref, same as inlineCss (say startHttpServer).  You might want another pref for the port to use.  Then use `require("sdk/simple-prefs").prefs.startHttpServer` to read the pref.  That will only work on startup (and require a restart if you change the pref), but that would be enough for merging.

ben-en:
@fzzzy this takes care of adding the prefs as well as importing simple-prefs as advised.

addon/lib/headless.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Usage without X:
+ * To use this without X, start the server as normal. Then, in the bin/
+ * directory, run `sh ephemeral-x.sh ./run-addon`.
+ */
+
+const { Cc, Ci, Cu } = require("chrome");
+const tabs = require("sdk/tabs");
+
+const HTTPD_MOD_URL = module.uri.replace("/headless.js", "/httpd.jsm");
```
fzzzy: Good workaround for getting the url of the module. Out of curiosity, what is the full url that ends up getting used?

ben-en: It's something like `resource://jid1-neeaf3sahdkhpa-at-jetpack/lib/httpd.jsm`. Not sure if the first part changes depending on whether it's in production mode or not.

fzzzy: Ok, that is good to know. I think that string is stable for all versions of pageshot, we have it hardcoded other places in the code. That was the namespace I was talking about needing to be defined to be able to refer to the module; I'm really glad I now know that jetpack extensions have a stable resource: url that can be used.

ben-en: The user `rpl` on IRC was the one who provided that, as well as the babel patch. I completely forgot to mention that anywhere in my code, however, so i'll give him some credit here. :D

PR #913, opened 2015-12-18
r? ianbicking Store in s3 and support virtual hosting

fzzzy:


addon/package.json
```diff
@@ -16,7 +16,7 @@
       "title": "backend URL",
       "description": "URL of the backend service to store pages",
       "type": "string",
-      "value": "https://pageshot.dev.mozaws.net"
+      "value": "https://pageshotfzzzy.dev.mozaws.net/"
```
ianb: I notice there's a trailing `/` here – that could lead to double slashes.  Really addon/lib/main.js:getBackend() should normalize this (maybe doing `.replace(/\/+$/, "")`, but it doesn't right now.  Of course anyone can enter anything in to the pref, so relying on the pref not having a trailing slash isn't really right either.

I'm guessing we do normalize in some other places, but maybe not consistently.

bin/build-docker-image
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+if [ $# -eq 0 ] ; then
+  echo "Usage: $0 dockerhubusername"
```
ianb: This should say something about what this script does.  E.g., does it build the entire docker image and upload it to docker hub under `dockerhubusername/pageshot` ?

addon/package.json
fzzzy: Also related to this, Benson did a successful deploy of this branch yesterday on his infrastructure, and noticed that this value needs to come from the environment, not be hard coded.

bin/run-docker
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
ianb: Is this actually a build file for use *inside* docker?  If so maybe it should be named differently, like with a leading underscore (what I've used for some scripts only intended to be run by make) or just put elsewhere (I'm not particularly happy with the leading underscore approach)

server/db-patches/patch-6-7.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE images DROP COLUMN image;
+ALTER TABLE images ADD COLUMN url TEXT;
```
ianb: Doesn't need to happen in this patch, but I realized it would be nice to store the image sizes (bytes) in the database.

bin/build-docker-image
fzzzy: Yes, it does now (it used to only make the image locally). I'll add a more informative string here and detail it in the docs.

bin/run-docker
fzzzy: Yes, it is intended to run inside the container. I'll add a leading underscore for now.

server/db-patches/patch-6-7.sql
fzzzy: Good idea.

server/src/server.js
```diff
@@ -675,10 +696,23 @@ exports.simpleResponse = simpleResponse;
...
+    mainapp.get("/", function (req, res) {
+      res.send("ok");
+    });
+    console.info(`virtual host server listening on http://localhost:${config.port}`);
+  } else {
+    app.listen(config.port);
+    console.info(`server listening on http://localhost:${config.port}/`);
+    contentApp.listen(config.contentPort);
+    console.info(`content server listening on http://localhost:${config.contentPort}/`);
+  }
```
ianb: I'm betting we could do `mainapp.use(vhost("localhost:10080"))` and `mainapp.use(vhost("localhost:10081"))` and it would work, and then we wouldn't need two sets of logic for two kinds of dispatching.  Then I also assume we can call `mainapp.listen()` twice with the two ports.

ianb: (The docs on vhost aren't explicit about port matching, so I'm not actually sure if that works)

server/src/servershot.js
```diff
@@ -4,8 +4,88 @@ const db = require("./db");
...
+          resolve();
+        }
+      });
+    });
+  };
+} else {
+  const AWS = require('aws-sdk');
+
+  // Test a PUT to s3 because configuring this requires using the aws web interface
+  // If the permissions are not set up correctly, then we want to know that asap
```
ianb: Isn't the test in server.js now?

ianb:
That's all my comments

server/src/server.js
fzzzy: That would be awesome if it worked, but I just tried it and it doesn't listen on the second port

server/src/servershot.js
fzzzy: Oops. yes

fzzzy:
Wow. I completely forgot to implement delete.

fzzzy:
I better implement delete and have you review it again.

fzzzy:
Ok, I implemented delete, verified that it works, and also took care of your other comments. I think it's ready to go, but you should take another quick look.

If you want to do sanity checking on the backend url, you should open another bug for that.

PR #921, opened 2015-12-30
r?fzzzy Remote resource

ianb:
This implements rewriting resources in the document (replacing each resource with a uuid), collecting the data about those resources, rewriting those uuids when serving through the web frontend, and implementing a very simple proxy to serve up those resources.


addon/data/framescripts/make-static-html.js
```diff
@@ -281,6 +366,7 @@ function createStyle(doc) {
     skipRule: function (rule) {
       this.rulesOmitted++;
       this.charsOmitted += rule.cssText.length;
+      this.rules.push(`/* skipped: ${rule.cssText} */`);
```
ianb: Removing this in #925

PR #942, opened 2016-01-13
r? ianbicking Support sentry

fzzzy:


ianb:
There are no files changed...?

server/src/server.js
```diff
@@ -102,6 +110,10 @@ app.use(function (req, res, next) {
   next();
 });
 
+app.get("/trigger-error", function (req, res) {
+  throw new Error("test error");
+});
+
```
ianb: We shouldn't leave this in indefinitely, though I can see the use in being able to test that Sentry is working.  Maybe make a followup bug to remove this?

PR #950, opened 2016-01-21
r? ianbicking Sentry on client

fzzzy:


addon/data/error-panel.js
```diff
@@ -23,6 +23,9 @@ self.port.on("showError", function (error) {
   if (error.stack) {
     console.error(error.stack);
   }
+  unsafeWindow.error = cloneInto(error, unsafeWindow);
+  unsafeWindow.captureException(unsafeWindow.error);
+
```
ianb: Did you not go with the event approach?  I've found this approach fragile in weird ways (e.g., strings go across the boundary fine, but arrays act weird).  Though I'm unclear what cloneInto is?

addon/lib/errors.js
```diff
@@ -9,6 +9,7 @@ const panels = require("sdk/panel");
 const self = require("sdk/self");
 const req = require("./req");
 const main = require("./main");
+const user = require("./user");
```
ianb: What's this import for?

ianb: Oh nevermind, I see.

addon/data/error-panel.js
fzzzy: I decided not to, because cloneInto fixes the weird problems. It does a structured clone into the new window, so there are no proxies or weird wrappers.

PR #951, opened 2016-01-22
r? ianbicking Full page hash

fzzzy:


server/src/views/frame.js
```diff
@@ -222,6 +222,7 @@ class Frame extends React.Component {
 
   clickFullPageButton(e) {
     e.preventDefault();
+    history.pushState({state: "frame"}, "", "#frame");
```
ianb: Why can't this just be `location.hash = "#frame"` ?

fzzzy: it can

ianb: Oh, maybe because it's not a place...

BUT, I believe we have an hashchange handler in clientglue?  And we could move all this code over there.

fzzzy: If we use location.hash, won't the browser scroll right to the element? We want to scroll to the element minus a bit. I think this implementation is better.

ianb: Yeah, that's what I realized.  But if we don't have any click handler, let the links point the user to `#frame` as they will by default, then put in a special case for `frame` in clientglue:refreshHash, I think that'd work well.

fzzzy: but that will still cause the browser to jump to the frame, and then our code will scroll it back a bit. too janky.

ianb: We don't need to have any element with `id="frame"`, just like the clip ids don't point to an element

ianb: Note that it the hash isn't useful unless when I open the URL and it includes `#frame` then I also navigate correctly to the full page view.

fzzzy: but then links that contain #frame won't jump to the frame without more work.

ianb: refreshHash is called on pageload when necessary

fzzzy: but all that is way more code than this implementation that already works. why?

ianb: It moves code around, but I don't think it's more code.  And it means no matter how you get the hash in your URL, you'll get consistent behavior.

fzzzy: there are other parts of code that rely on the frame having an id. also, i think the inconsistent experience is better -- when the user clicks on the full page button, we want to show it to them, because they just clicked it. but, when the user visits the page with the hash, they should just see the full page

ianb: The use case I had in mind was specifically sharing the URL with the hash, just like I'd share the URL with the clip ID to link to a clip.  I don't think there's much purpose to changing the hash except to share it.  (I guess it lets you use back/forward, but that's only a minor and questionable feature)

ianb: Note we could use `#fullpage` instead of `#frame` for the URL hash

fzzzy: What do you mean? That's how it works.

ianb: Won't the toolbar overlap the top of the page?

fzzzy: You're right. The toolbar overlaps part of the page.

ianb: I did forget about leaving room for the button when you click it; it does make sense for it not to display for a page viewer.  I don't have a very good idea for that.  Except for some hacky flag or something.

fzzzy: I guess it won't be a problem if the user that comes to the page from the url with the hash sees the full page button. Maybe it will be a hint that there is something above there.

PR #952, opened 2016-01-22
r? ianbicking Full page hash

fzzzy:


server/src/clientglue.js
```diff
@@ -137,6 +137,12 @@ function refreshProfile(e) {
 }
 
 function refreshHash() {
+  if (location.hash === "#fullpage") {
+    let frameOffset = document.getElementById("frame").getBoundingClientRect().top + window.scrollY;
+    let toolbarHeight = document.getElementById("toolbar").clientHeight;
+    let frameTop = frameOffset - (toolbarHeight * 2);
+    window.scroll(0, frameTop);
```
ianb: Shouldn't this return?

ianb:
I think a return makes it clearer, but besides that r+

PR #953, opened 2016-01-26
r? ianbicking Fixes #936 Save shot timing info in ga.

### LINKED ISSUE #936: Save timing information in GA
We can use [User Timings](https://developers.google.com/analytics/devguides/collection/analyticsjs/user-timings?hl=en) to save time information to GA.  It would be nice to measure a few things using this, like shot uploads.  This might go through `/event` or might have its own server endpoint?  Same pattern though.

fzzzy:


addon/lib/shooter.js
```diff
@@ -104,6 +105,8 @@ function extractWorker(tab, timeLimit) {
     });
     watchWorker(worker);
     worker.port.on("data", function (data) {
+      let timeFinished = new Date();
+      sendTiming("addon", "make-shot", timeFinished - timeStarted);
```
ianb: This isn't a good name – the shot is really done by make-static-html, this is extracting metadata.  So `"extract-data"` would be a better description.

server/src/server.js
```diff
@@ -170,6 +170,27 @@ app.post("/event", function (req, res) {
   });
 });
 
+app.post("/timing", function (req, res) {
+  let bodyObj = req.body;
+  if (typeof bodyObj !== "object") {
+    throw new Error("Got unexpected req.body type: " + typeof bodyObj);
+  }
+  let userKey = dbschema.getTextKeys()[0] + req.deviceId;
```
ianb: We should make this into a function, like `hashUserId(req.deviceId).then(...)`

addon/lib/shooter.js
fzzzy: Is this the right thing to measure, though?

ianb: It's the pokiest stage.  The `collectInformation` method in shooter.js is a good place to catch this info.  I don't know how GA does visualizing different stages, but the add-ids stage (that should generally be very fast, and blocks the rest), then extractWorker (which is slow), the make-static-html framescript call (which is usually pretty fast, but is worth watching), there's the full screen screenshot (which I think is fast but might be worth capturing in case it sometimes isn't).

Also in _activateWorker there's the handler for the "select" message, and that does the clip images.  Those might also be worth measuring...?

fzzzy:
Ok, renamed to extract-data, and pulled out a hashUserId function.

ianb:
We should make a followup bug for adding some more timing tests in the add-on

fzzzy:
opened #963

PR #978, opened 2016-02-24
Work in progress, for code review -- new pageshot with topbar instead of panel

fzzzy:


server/src/clientglue.js
```diff
@@ -23,6 +23,24 @@ exports.setModel = function (data) {
...
+          paddingTop = 16;
+        }
+        container.style.paddingTop = paddingTop + "px";
+      }
+      window.addEventListener("resize", onResize, true);
+
+      onResize();
+
+    }
+
```
ianb: BTW, I have a (rough) lightbox component here: https://github.com/ianb/hashroom/blob/2dbd467cf4d17b371b19f2b85e2ddd87ff1fc7bf/html/js/script.js#L168

It also doesn't deal with resizing, but if I did I'd probably put it in the component onComponentDidMount and hook into window.resize there.  This might require putting the window size in state.

server/src/views/frame.js
```diff
@@ -7,7 +7,70 @@ const { ProfileButton } = require("./profile");
...
+  }
+
+  onClickCopyButton(e) {
+    let target = e.target;
+    target.previousSibling.select();
+    document.execCommand("copy");
+    target.textContent = "Copied";
+    setTimeout(function() {
+      target.textContent = "Copy";
+    }, 1000);
```
ianb: This could use `this.setState({copyText: "Copied"})` and a timeout to undo that state, and then just render the button using `<button>{this.state.copyText}</button>`

server/src/views/frame.js
```diff
@@ -7,7 +7,70 @@ const { ProfileButton } = require("./profile");
...
+        className="copy-shot-link-button"
+        onClick={ this.onClickCopyButton.bind(this) }>
+        Copy
+      </button>
+    </div>;
+  }
+}
+
 class Clip extends React.Component {
+  onClick(e) {
```
ianb: Dead code?

server/src/views/frame.js
```diff
@@ -7,7 +7,70 @@ const { ProfileButton } = require("./profile");
...
+    let fullPageCheckbox = document.getElementById("full-page-checkbox");
```
ianb: This should use setState instead of a hidden checkbox

server/src/views/frame.js
```diff
@@ -225,6 +275,37 @@ class Frame extends React.Component {
     ga("send", "event", "website", "click-full-page-button", {page: location.toString()});
   }
 
+  onChange(e) {
+    let evt = new Event("hashchange");
+    window.dispatchEvent(evt);
+  }
```
ianb: This might be dead code, or worth killing

server/src/views/frame.js
```diff
@@ -490,7 +584,7 @@ class ExpireWidget extends React.Component {
       <span>
         keep for <select ref="expireTime">
           <option value="cancel">Select time:</option>
-          <option value="0">Never</option>
+          <option value="0">Forever</option>
```
ianb: This should be "Indefinite" (I believe there's a ticket for this too)

server/src/views/shell.js
```diff
@@ -17,7 +17,9 @@ export class Shell extends React.Component {
       <script src={ this.props.staticLink("vendor/introjs/intro.js") } key="introjs-js" />,
       <script src={ this.props.staticLink("js/server-bundle.js") } key="server-bundle-js" />,
     ];
-    js = js.concat(gaActivation(this.props.gaId, this.props.deviceId, false));
+    if (this.props.gaId) {
```
ianb: I had intended (but probably forgot) to make gaActivation put in a stub `ga()` function when gaId wasn't provided.  Then this could be unconditionally included.

addon/lib/main.js
```diff
@@ -90,7 +159,8 @@ PanelContext = {
       throw new Error("Hiding wrong shotContext");
     }
     this._activeContext = null;
-    shootPanel.hide();
+    hideNotificationBar();
+    //shootPanel.hide();
```
ianb: can delete

addon/lib/main.js
```diff
@@ -124,7 +194,8 @@ PanelContext = {
   /** Show a ShotContext, hiding any other if necessary */
   show: function (shotContext) {
     if (this._activeContext === shotContext) {
-      shootPanel.show();
+      //shootPanel.show();
```
ianb: can delete commented code

addon/lib/shooter.js
```diff
@@ -150,10 +150,11 @@ const ShotContext = Class({
     this.panelContext = panelContext;
     this._workerActive = false;
     this.watchTab("deactivate", function () {
-      panelContext.hide(this);
+      //panelContext.hide(this);
```
ianb: can remove (and remove the entire deactive handler, since it no longer calls anything)

addon/lib/shooter.js
```diff
@@ -150,10 +150,11 @@ const ShotContext = Class({
...
     });
     this.watchTab("activate", function () {
-      panelContext.show(this);
+      //console.log("WATCHTAB ACTIVATE");
+      //panelContext.show(this);
```
ianb: ditto here

addon/lib/shooter.js
```diff
@@ -522,7 +537,7 @@ const ShotContext = Class({
         // FIXME: maybe pop up an explanation here?
         return;
       }
-      this.copyRichDataToClipboard();
+      //this.copyRichDataToClipboard();
```
ianb: can delete

addon/lib/shooter.js
```diff
@@ -551,16 +566,16 @@ const ShotContext = Class({
           this.shot.update(attrs);
         }, this)));
       watchPromise(allPromisesComplete(promises).then((function () {
-        return this.shot.save();
+        //return this.shot.save();
```
ianb: I guess this whole handler is now unnecessary?  Except you shouldn't be able to save until it completes (which this code currently doesn't protect against)

ianb: Do:

```javascript
this._everythingSaved = watchPromise(allPromisesComplete(promises))
// later:
this._everythingSaved.then(() => {
  this.uploadData();
});
```

Then actual uploading will always block on everything being completed.  Note that allPromisesComplete resolves even when there is a failed promise.

addon/lib/shooter.js
```diff
@@ -551,16 +566,16 @@ const ShotContext = Class({
...
     }).bind(this)));
   },
 
   /** Sets attributes on the shot, saves it, and updates the panel */
   updateShot: function () {
-    watchPromise(this.shot.save());
+    //watchPromise(this.shot.save());
     this.panelContext.updateShot(this);
-    require("./recall").addRecall(this.shot);
+    //require("./recall").addRecall(this.shot);
```
ianb: We'll remove recall in a later commit, so leave this in

addon/lib/shooter.js
```diff
@@ -551,16 +566,16 @@ const ShotContext = Class({
...
-        return this.shot.save();
+        //return this.shot.save();
       }).bind(this)));
     }).bind(this)));
   },
 
   /** Sets attributes on the shot, saves it, and updates the panel */
   updateShot: function () {
-    watchPromise(this.shot.save());
+    //watchPromise(this.shot.save());
```
ianb: Can be removed

server/src/clientglue.js
```diff
@@ -137,6 +155,18 @@ function refreshProfile(e) {
 }
 
 function refreshHash() {
+  let fullPageCheckbox = document.getElementById("full-page-checkbox");
```
ianb: Should remove the stuff related to this checkbox

server/src/views/frame.js
```diff
@@ -225,6 +275,37 @@ class Frame extends React.Component {
...
   }
 
+  onChange(e) {
+    let evt = new Event("hashchange");
+    window.dispatchEvent(evt);
+  }
+
+  onClickUploadFullPage(e) {
+    let uploadFullPageButton = document.getElementById("upload-full-page");
+    uploadFullPageButton.parentNode.removeChild(uploadFullPageButton);
```
ianb: This should change the model and trigger a re-render

server/src/views/frame.js
```diff
@@ -225,6 +275,37 @@ class Frame extends React.Component {
...
+    let node = document.getElementById("full-page-checkbox");
+    node.checked = true;
```
ianb: another checkbox item to remove

server/src/views/frame.js
```diff
@@ -225,6 +275,37 @@ class Frame extends React.Component {
...
+  }
+
+  onClickShareButton(e) {
+    let sharePanel = document.getElementById("share-buttons-panel");
+    console.log("SHARE STYLE", sharePanel.style.display);
+    if (sharePanel.style.display === "block") {
+      sharePanel.style.display = "none";
+    } else {
+      sharePanel.style.display = "block";
+    }
```
ianb: This should be in state, I think?

Also clicking off the share panel should close the panel.  Probably the panel should be a subcomponent, and have its own hidden/showing state, with onComponentDidMount adding a handler to catch document click and hide the panel.

server/src/views/frame.js
```diff
@@ -378,55 +460,61 @@ class Frame extends React.Component {
     </div>
     */
 
+    /*
```
ianb: We can leave this in and try to figure out if it should go back somewhere later

fzzzy:
Ok, I think I addressed all these review comments.

PR #1005, opened 2016-03-03
r?fzzzy Fix 920

ianb:
I'm not seeing the topbar now, but I don't know why that would be related to this patch...?


fzzzy:
I was seeing some weird bugs this morning that sound similar to what you mentioned, but I deleted my build directory and built everything from scratch and things seem to be ok.

PR #1007, opened 2016-03-03
r? ianbicking More cleanup

fzzzy:


server/src/views/frame.js
```diff
@@ -394,14 +373,35 @@ class Frame extends React.Component {
                 { ...this.props } />
     }
 
+    let trashOrFlagButton = null;
+    if (this.props.isOwner) {
+      trashOrFlagButton = <img src={ this.props.staticLink("img/garbage-bin.png") } />;
+    } else {
+      trashOrFlagButton = <img src={ this.props.staticLink("img/flag.png") } />;
+    }
```
ianb: Let's put the onClick handler here, and have onClickDelete and onClickFlag as separate functions

PR #1046, opened 2016-03-11
r? ianbicking Fix no clip error

fzzzy:


server/src/views/frame.js
```diff
@@ -501,10 +501,10 @@ class Frame extends React.Component {
...
     let zoomButton = null;
-    if (this.props.shot.showPage) {
+    if (clipNames.length && this.props.shot.showPage) {
       zoomButton = <img
         style={{
-          position: "absolute",
+          position: "fixed",
           top: "81px",
           right: "15px",
           height: "32px",
```
ianb: Github is hatin' on `'` in string literals in JSX, we should probably escape those.

fzzzy: I noticed that, can we use &apos; instead?

PR #1061, opened 2016-03-15
r? ianbicking Fix reminder message

fzzzy:


ianb:
There's a line between the message box and the arrow:

image

Is it possible to move the arrow down a pixel and use z-index so it goes over the border?  Or else just drop the border.

fzzzy:
Hmm, not on my screen. Must be because of retina. I'll move it down a bit more

ianb:
Cool, that fixes it

PR #1065, opened 2016-03-16
r? ianbicking Add hotkey

fzzzy:
Currently, using Command-option-control-c. Suggestions for another
hotkey? (How do I know if there is a conflict with an existing key?)


ianb:
I have no idea how to decide what hotkeys are appropriate, except I often see controversy over specific choices :-|

PR #1123, opened 2016-04-20
r?fzzzy Fix selection

ianb:
Major refactoring and performance improvement to the selection


ianb:
(I deployed this branch, because I wanted to get the homepage and timeout bug fixes out, but the branch still isn't actually merged)

addon/data/extractor-worker.js
```diff
@@ -15,6 +15,35 @@ var FILENAME = "extractor-worker.js";
     */
 function extractData() {
   let start = Date.now();
+  // FIXME: need to include found images too
+  let startImage = Date.now();
```
fzzzy: This isn't needed any more, just use start instead

static/homepage/index.html
```diff
@@ -133,9 +119,13 @@
             <h1 class="hero-header-text large">Share anything on the web.</h1>
             <h4 class="hero-tagline">Save, highlight, send
             </h4>
-            <p>
+            <p class="if-firefox">
```
fzzzy: I don't see how the globals in the js get defined to css classes

fzzzy:
Ok, looks good with those nits

static/homepage/index.html
ianb: The JS adds a class to body, and the CSS uses selectors to hide or show these classes based on the body class.  I put a comment into the page JS to explain.

PR #1136, opened 2016-04-25
r?fzzzy More misc fixes

ianb:


addon/data/shooter-interactive-worker.js
```diff
@@ -477,10 +463,26 @@ function addHandlers() {
...
+    self.port.emit("deactivate");
+  }
+  if (event.key === "Enter") {
+    self.port.emit("take-shot");
+  }
 }
 
 function removeHandlers() {
-  for (let name of registeredDocumentHandlers) {
+  for (let name in registeredDocumentHandlers) {
```
fzzzy: Why this change? I think "of" is correct. With "in", name will be 0, 1, 2, etc

fzzzy: Never mind, i see that it is not an array

ianb: Of course at one time it *was* an array, hence the bug...

PR #1181, opened 2016-05-26
r?fzzzy Misc fixes

ianb:
Nothing too complicated, I think


ianb:
Note, commits are unrelated, easiest to review per-commit

PR #1205, opened 2016-07-05
r? ianbicking Document the procedure for publishing on Elastic Beanstalk.

fzzzy:


ianb:
You don't actually mention where you configure EB to use RDS.  Is it when you create the Beanstalk app?  Or do you set `RDS_*` variables?

fzzzy:
I did, I said "When you create the Elastic Beanstalk instance...". There is no need to set any RDS_* variables, beanstalk does it automatically.

ianb:
Yeah, just thinking about how it would seem if I wasn't clear about what Elastic Beanstalk did where or when.  But sure.

PR #1209, opened 2016-07-05
r? ianbicking Try really hard to always use the system font, no matter what system we are on.

fzzzy:


ianb:
We really need some common file to include across the project, so we can use `font-family: $system-font`

Issue #1212, opened 2016-07-05
A few sass-lint warnings


pdehaan:
### Steps to reproduce:
1. Create a file named ./.sass-lint.yml in the root directory.
2. Add the following contents and save:
   ```yml
files:
      include: ./static/css/*.scss

options:
      merge-default-rules: true

rules:
      force-attribute-nesting: 0
      force-element-nesting: 0
      force-pseudo-nesting: 0
      hex-length: 0
      hex-notation: 0
      leading-zero: 0
      no-color-keywords: 0
      no-color-literals: 0
      no-css-comments: 0
      no-ids: 0
      no-qualifying-elements: 0
      property-sort-order: 0
      quotes: 0
      single-line-per-selector: 0
      space-after-comma: 0
      url-quotes: 0
      zero-unit: 0
```
3. Install **sass-lint** as a dev dependency: `$ npm i sass-lint -D`
4. Add an npm package.json script for `"sass": "sass-lint -v -q",`
5. Run `$ npm run sass`

### Actual results:
```
$ npm run sass

...
✖ 131 problems (0 errors, 131 warnings)
```

Most seem to be formatting or mixed tabs-spaces, but a few could be actual errors (like some duplicated rules in a selector.
It should be pretty easy to add/remove other rules to the .sass-lint.yml file if we want to silence some rules for now until we can get down to a more reasonable number. Also pretty easy for me to jump in there and fix some formatting and the easy warnings to get the number down pretty quickly, then re-audit.

**Comment by pdehaan** [comment]
### Current output

```

> pageshot@0.0.1 sass /Users/pdehaan/dev/github/mozilla-services/pageshot
> sass-lint -v -q


./static/css/inline-selection.scss
   53:11  warning  A value of `none` is not allowed. `0` must be used                                         border-zero
   56:24  warning  Class '.pageshot-topLeft' should be written in lowercase with hyphens                      class-name-format
   62:1   warning  Space expected between blocks                                                              empty-line-between-blocks
   71:1   warning  Space expected between blocks                                                              empty-line-between-blocks
   72:24  warning  Class '.pageshot-topRight' should be written in lowercase with hyphens                     class-name-format
   78:1   warning  Space expected between blocks                                                              empty-line-between-blocks
   86:1   warning  Space expected between blocks                                                              empty-line-between-blocks
   94:1   warning  Space expected between blocks                                                              empty-line-between-blocks
[...]

PR #1213, opened 2016-07-06
Add sass-lint workflow

pdehaan:
Ref: #1212 

Turns out you can mostly clean your Sass by converting your SCSS to SCSS using Sass' `sass-convert` tool:

```sh
$ sass-convert static/css -R -T scss -F scss -i
```

This leaves us with 24 warnings which will consider a bit more thought (or relaxing of sass-lint rules):

```sh
$ npm run sass

> pageshot@0.0.1 sass /Users/pdehaan/dev/github/mozilla-services/pageshot
> sass-lint -v -q


./static/css/inline-selection.scss
   63:24  warning  Class '.pageshot-topLeft' should be written in lowercase with hyphens                      class-name-format
   81:24  warning  Class '.pageshot-topRight' should be written in lowercase with hyphens                     class-name-format
  107:24  warning  Class '.pageshot-bottomLeft' should be written in lowercase with hyphens                   class-name-format
  124:24  warning  Class '.pageshot-bottomRight' should be written in lowercase with hyphens                  class-name-format
  150:2   warning  Class '.pageshot-topLeft' should be written in lowercase with hyphens                      class-name-format
  165:2   warning  Class '.pageshot-topRight' should be written in lowercase with hyphens                     class-name-format
  186:2   warning  Class '.pageshot-bottomLeft' should be written in lowercase with hyphens                   class-name-format
  201:2   warning  Class '.pageshot-bottomRight' should be written in lowercase with hyphens                  class-name-format
  311:11  warning  @extend must be used with a %placeholder                                                   placeholder-in-extend
  344:3   warning  Vendor prefixes should not be used                                                         no-vendor-prefixes
  347:49  warning  Class '.pageshot-panel-arrowUp' should be written in lowercase with hyphens                class-name-format
  373:11  warning  @extend must be used with a %placeholder                                                   placeholder-in-extend
  475:5   warning  Vendor prefixes should not be used                                                         no-vendor-prefixes
  476:5   warning  Vendor prefixes should not be used                                                         no-vendor-prefixes
  593:3   warning  Rule `.pageshot-reset input, textarea, select` should be merged with the rule on line 584  no-mergeable-selectors
  596:5   warning  Property `*font-size` appears to be spelled incorrectly                                    no-misspelled-properties

./static/css/pageshot-notification-bar.scss
   1:78  warning  Protocols and domains in URLs are disallowed                       no-url-protocols
   4:2   warning  Vendor prefixes should not be used                                 no-vendor-prefixes
   6:5   warning  Vendor prefixes should not be used                                 no-vendor-prefixes
  74:52  warning  Class '.messageImage' should be written in lowercase with hyphens  class-name-format

./static/css/panel.scss
  252:10  warning  No empty blocks allowed  no-empty-rulesets

./static/css/styles.scss
  224:20  warning  No empty blocks allowed  no-empty-rulesets
  316:12  warning  No empty blocks allowed  no-empty-rulesets
  521:10  warning  No empty blocks allowed  no-empty-rulesets

✖ 24 problems (0 errors, 24 warnings)
```

### DIFF
--- a/static/css/styles.scss
+++ b/static/css/styles.scss
@@ -615,5 +615,5 @@ a.subheading-link {
 
 .intro-js-highlight-overlay {
   background: rgba(#000, 0);
-  border: 2px solid rgba(0, 137, 219, 1);
+  border: 2px solid rgba(#0089db, 1);
 }
--- a/static/css/styles.scss
+++ b/static/css/styles.scss
@@ -2,8 +2,7 @@ $primary-color: #fcfcfc;
 $toolbar-text-color: #464239;
 $text-color: #f2f2f2;
 $background-color: #282d33;
-$shadow-color: rgba(0, 0, 0, 0.5);
-
+$shadow-color: rgba(#000, 0.5);
 
 html {
   height: 100%;
@@ -22,17 +21,17 @@ body {
 
 #container {
   position: absolute;
-  left: 0px;
-  right: 0px;
-  top: 0px;
-  padding-top: 0px;
-  padding-left: 0px;
-  padding-right: 0px;
+  left: 0;
+  right: 0;
+  top: 0;
+  padding-top: 0;
+  padding-left: 0;
+  padding-right: 0;
   text-align: center;
 }
 
 #frame {
-  border: none;
+  border: 0;
   margin-top: 65px;
   margin-bottom: 4em;
   box-shadow: 0 -5px 5px $shadow-color;
@@ -40,9 +39,9 @@ body {
 
 #toolbar {
   position: fixed;
-  top: 0px;
-  left: 0px;
-  right: 0px;
+  top: 0;
+  left: 0;
+  right: 0;
   height: 35px;
   padding: 15px;
   padding-left: 150px;
@@ -54,7 +53,7 @@ body {
   color: $toolbar-text-color;
   text-align: left;
   background-color: $primary-color;
-  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.5);
+  box-shadow: 0 1px 2px rgba(#000, 0.5);
 }
 
 #private-notice {
@@ -67,7 +66,7 @@ body {
   padding: 7px;
   padding-bottom: 1px;
   background-color: #ebd55a;
-  color: black;
+  color: #000;
   cursor: pointer;
   z-index: 2;
 }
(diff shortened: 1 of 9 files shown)


static/css/styles.scss
```diff
@@ -599,19 +601,19 @@ a.subheading-link {
...
 
   &:hover {
     text-decoration: underline;
   }
 }
 
 .intro-js-highlight-overlay {
-  background: rgba(0,0,0,0);
+  background: rgba(#000, 0);
   border: 2px solid rgba(0, 137, 219, 1);
```
pdehaan: We could also probably convert the `rgba(0, 137, 210, 1)` to the Sass shorthand of `rgba(#0089db, 1)` (or even just `#0089db` for that matter).

fzzzy: I'd say just use `#0089db`

pdehaan: Done. Converted all `rgba(r,g,b,a)` to the cleaner Sass `rgba(#rgb, a)` syntax.

fzzzy:
What's the rationale behind switching all the 0px to plain 0? Is that sass specific? I don't think that will actually do the correct thing in css, and I strongly prefer 0px.

pdehaan:
> Is that sass specific?

Nope. CSSLint has rules for it too https://github.com/CSSLint/csslint/wiki/Disallow-units-for-zero-values

I think the rationale is that 0px is the same as 0pt is the same as 0em is the same as 0rem is the same as 0{anything}...
Basically when your value is 0, units aren't relevant at all.

Or if you _strongly_ prefer the units on the end, we could always just ignore that rule.

fzzzy:
Maybe this has changed, but in the past I have had problems where 0 does not work but 0px does.

pdehaan:
I'm fine with whatever you want. It turns out we can tweak the rule to say that units are always required: https://github.com/sasstools/sass-lint/blob/develop/docs/rules/zero-unit.md

I'll happily set `include: true` and add all the units back. Whatever gets me +1'ed and merged (and more importantly the code consistently consistent).

fzzzy:
Ok, so the CSS spec does indeed say the units are optional when the value is 0. I wouldn't be surprised if I had problems with 0 vs 0px more than 10 years ago, which is when I was writing the most css. I'm fine with leaving it off.

(It sure stuck in my head, such that I always always add the units when I am writing css, so it must have been particularly difficult to debug. :)

PR #1218, opened 2016-07-06
Fix add-on, Node.js, and GitHub typos

### LINKED ISSUE #1217: Add-on vs Addon
Noticed some minor inconsistencies on https://pageshot.dev.mozaws.net/ where we say **Addon** or in some cases hyphenated **add-on**.

We should probably stick with one style, and probably use the hyphenated version to be in sync w/ [Test Pilot](https://testpilot.firefox.com/) and https://addons.mozilla.org/firefox/

pdehaan:
Fixes #1217

### DIFF
--- a/static/homepage/index.html
+++ b/static/homepage/index.html
@@ -97,7 +97,7 @@
         <div class="navbar-collapse collapse">
           <ul class="nav navbar-nav navbar-right main-nav">
             <li><a href="./" class="scrollnav active" >Overview</a></li>
-            <li><a href="https://github.com/mozilla-services/pageshot" target="_blank">Github</a></li>
+            <li><a href="https://github.com/mozilla-services/pageshot" target="_blank">GitHub</a></li>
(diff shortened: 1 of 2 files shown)


static/homepage/index.html
```diff
@@ -120,7 +120,7 @@ <h4 class="hero-tagline">Save, highlight, send
               <a class="btn btn-primary btn-lg main-get-started installer" href="https://chrome.google.com/webstore/detail/pageshot/flhjebpmjblnihhioankmibfdkdiaahi">Install</a>
             </p>
             <p class="if-not-firefox-or-chrome">
-              <a class="btn btn-primary btn-lg" href="https://www.mozilla.org/en-US/firefox/new/">To use PageShot: Install Firefox!</a><br>
```
pdehaan: Removed the `/en-US/` here since mozilla.org will redirect for us, and give the user their preferred/best locale (whenever possible).

PR #1250, opened 2016-07-14
r? ianbicking Fixes #1191 Implement the new overlay menu flow.

### LINKED ISSUE #1191: Replace top panel with a focused overlay menu
My initial design:

edit initial alternate 1x

Helen Huang is going to help us refine Pageshot’s visual and UI design, and I think that this issue is a perfect one to start on.

fzzzy:


addon/data/selector-ui.js
```diff
@@ -159,6 +159,72 @@ const ui = (function () { // eslint-disable-line no-unused-vars
...
+        this.cancelButton.style.left = "0";
+        this.cancelButton.onclick = callbacks.onCancel;
+        this.myShotsButton = document.createElement("button");
+        this.myShotsButton.className = "pageshot-select-mode-button";
+        this.myShotsButton.textContent = "Open My Shots";
+        this.myShotsButton.style.right = "0";
+        this.myShotsButton.style.backgroundColor = "#248aeb";
+        this.myShotsButton.style.color = "white";
+        this.myShotsButton.onclick = callbacks.onOpenMyShots;
+
```
ianb: We gotta setup some helpers for this stuff, but that can happen in a follow up.

fzzzy: I was thinking we should be able to use react in here somehow?

addon/data/selector-ui.js
```diff
@@ -159,6 +159,72 @@ const ui = (function () { // eslint-disable-line no-unused-vars
     boxBottomEl: null
   };
 
+  /** Displays the box which allows the user to choose between selecting a region,
+      archiving the page, or going to my shots.
+  */
+  exports.SelectMode = {
+    display: function (callbacks) {
+      if (document.querySelector(".pageshot-preview-overlay") !== null) {
+        return;
```
ianb: What does this check accomplish?

addon/data/selector-ui.js
```diff
@@ -159,6 +159,72 @@ const ui = (function () { // eslint-disable-line no-unused-vars
...
+      }
+      let oldNode = document.getElementById("pageshot-select-mode-background");
```
ianb: And what's this one for?  Shouldn't select mode handle that element?

addon/data/selector-ui.js
fzzzy: It prevents opening the mode selector again when already making a selection with the resize handles. It would be better to do it at a higher level and not even allow the creation of the shot context, but I tried a few times and couldn't figure out how to correctly do it that way.

addon/data/selector-ui.js
```diff
@@ -159,6 +159,72 @@ const ui = (function () { // eslint-disable-line no-unused-vars
...
+      if (oldNode !== null) {
+        oldNode.parentNode.removeChild(oldNode);
+      }
+
+      if (this.selectModeBackground === undefined) {
+        this.selectModeBackground = document.createElement("div");
+        this.selectModeBackground.id = "pageshot-select-mode-background";
+        this.selectModeBackground.className = "pageshot-select-mode-background";
+        this.selectMode = document.createElement("div");
+        this.selectMode.className = "pageshot-select-mode";
```
ianb: These are all children of this.selectModeBackground, right?  (If not then .remove() isn't thorough enough.)  Can they all just be local variables instead of instance variables?  Or if they are instance variables, then they should get cleaned up too.

addon/data/selector-ui.js
fzzzy: This one prevents opening the select mode panel when the select mode panel is already open... Again, would be better to prevent starting a new shot flow higher up but this is what I got to work.

addon/data/selector-ui.js
fzzzy: Yes, they are all children. They should probably all be local variables instead.

addon/data/selector-ui.js
ianb: Given the discussion, I think we should remove these and fix the bug (maybe in a follow up).  This code covers up a bug that I expect will show itself in other ways.

fzzzy: Ok. Let's do a follow-up. I tested a bunch and it seems to do the right thing right now... But maybe not.

addon/data/selector-ui.js
ianb: It runs in the context of the page, though somewhat sandboxed – I feel like it's worth it to avoid getting too clever.

addon/data/selector-ui.js
```diff
@@ -159,6 +159,72 @@ const ui = (function () { // eslint-disable-line no-unused-vars
...
...BsNoU50v+xWAyFQgHz8/OYnZ2FJUYdlpeXsbm5GWk4fgefzWZRKpUg+6zt7W0/beTt+/v7qNVqHViRJ6IA4DgOZJ/VrV1ZWUEqleo2R2a5WCyiXq+fOF5fWPl8HplM5oQ4Sgu8c+8u/1q6FbQsCRAsiUJdIVhqRlJBsCQKdYVgqRlJBcGSKNQVgqVmJBUES6JQV3xvSlWbPt1o4tUnG7t1CzZ7AeBiEu6z0/M7caTPAA/XbVSPOm8G9Gqv/AbYEwcuJR3cvmyhdNe46apD811v7PHFBwcHx3y/LSDl9pNBnb06EbOAnZqNQ3d9y3e/7DvMz8uPwOockDJuvS+P/64wdldrAIwJijfjyCaAPHuM5MsZFlW8PJuLuxHXWurd/p1FFo/gJ+9t9s9O1NEIweIN53iuZYGCx/N2nl2WXsXULnzwS5H7bbYC2svFqdmMO/hUO4J4/xRmkX6Nrwnz1hrDMt7QvK3uloPyy50bn5/VWzH3UhB9VI8MAm8u/GbPBt6kb0JjWLxDH0QRfnkyCbsYR/Xjdw4evLFR++v2IIzGC78/WTYMuxhH1re6495H2SG3WPgdqWwYMqOhcGd8GQ5F60NuhPFlKLISZcMAZ0xkpQDSvkqEX8qGAbBSNgwASUgoGwoSQ/5P2VDjBFE2DAOWyEoavvoiFX4pGwbASdkwACQhGclsWGdj8LyEPeog/B4OYNTBOBu2xuAR+lj4oPzywDDOhnwsvMrGsh69bc22TJ6z3Ckw/szIx+V531I97swq9Gr/0Z7d4Xvkv4lRGim9MQlslIFjd0DLwu4vzp5NV7X+IPqW9mLf7BzUVMrBBJt+C7sYR9bre2P4fAB8OXTc+bwp9kYlv6anx1vzh+v34ycu0V7tX6uOu7/p8xaujFuID+DewRgWn3WeYXOGM1nvVl+94H3eTe3Xc95+vL2cjtW4gz+d5gz3XgmWxvkhWARLg4CGlCKLYGkQ0JD63josLS3B65MMjX2PtJR/BdZdfGFVKpVubeSXZZ81Ps5uval4EkgkEkgmk53vDRcWFpBOpz3FUTZySIuLi4jH453PfjkQ27axt7cH9ilwlPnIY+eAcrmcC4ob5TfSUkEVXwKyz/JV0ApJgGBJFOoKwVIzkgqCJVGoKwRLzUgqCJZEoa4QLDUjqSBYEoW68gekmoYeRy70ZgAAAABJRU5ErkJggg==";
```
ianb: You should leave a comment in to indicate where the image source comes from, so it can be rebuilt.

addon/data/selector-ui.js
```diff
@@ -159,6 +159,72 @@ const ui = (function () { // eslint-disable-line no-unused-vars
...
+        this.chooseArchiveMode.appendChild(archiveImg);
+        this.chooseArchiveMode.appendChild(document.createTextNode("Archive Full Page"));
+        this.chooseArchiveMode.onclick = callbacks.onChooseArchiveMode;
```
ianb: For basic hygiene I think we should stick to addEventListener.

addon/data/shooter-interactive-worker.js
```diff
@@ -214,7 +249,7 @@ stateHandlers.crosshairs = {
     if (isChrome) {
       ui.ChromeInterface.showSaveFullPage();
     } else {
-      self.port.emit("showSaveFullPage");
+      //self.port.emit("showSave");
```
ianb: Should put a FIXME on this

addon/lib/main.js
```diff
@@ -63,21 +63,13 @@ function showNotificationBar(shotcontext) {
...
-  });
+  myShots.onclick = watchFunction(exports.openMyShots);
   let messageNode = thebox.ownerDocument.createElement("span");
   messageNode.style.border = "none";
   messageNode.style.marginLeft = "10px";
   messageNode.style.fontWeight = "normal";
-  messageNode.appendChild(thebox.ownerDocument.createTextNode("Select part of the page to save, or save full page without making a selection."));
-  fragment.appendChild(myShots);
+  messageNode.appendChild(thebox.ownerDocument.createTextNode("Select part of the page to save:"));
+  //fragment.appendChild(myShots);
```
ianb: Should have a FIXME about removing My Shots button code

addon/lib/shooter.js
```diff
@@ -377,11 +392,12 @@ const ShotContext = Class({
       } catch (e) {
         // Ignore... it's just a best effort to cancel the state
       }
-      this.interactiveWorker.destroy();
+      setTimeout(() => {
+        this.interactiveWorker.destroy();
+        this.interactiveWorker = null;
+      }, 1000);
```
ianb: What's going on here?  (I think I might have gotten an exception related to this in testing, but I'm not sure)

fzzzy: There was a race between the worker finishing and destroy destroying it... I guess there should be a fixme in here to send a message to the worker, which sends a message back when it is finished, and then destroy gets called after that... But this seemed to be enough for the worker to tear down the selection handles before destroy gets called. (Of course there's still a risk that it will race. But much less.)

static/css/inline-selection.scss
```diff
@@ -253,6 +253,57 @@ $overlay-z-index: 900000;
   z-index: $overlay-z-index;
 }
 
+.pageshot-select-mode-background {
```
ianb: This all needs to use `@extend .pageshot-reset;`, I'm seeing a lot of style conflicts.  E.g.:

image

Or:

image

fzzzy: Is it enough to put the reset on the root element of the tree?

ianb:
There's some nits in there, and removing the workaround for hitting the shot button twice.  It's very easy to hit styling problems, so that should get fixed with the reset (we might find new problems, but not quite so widely).

Do we really want a gigantic topbar?

image

Also I'm getting errors everywhere about this:

```
Object
    - name = Error
    - message = Timeout after 15000ms in pageshot@documentStaticData
    - help = An unexpected error occurred
    - stack = lib/framescripter.js:130
```

I'm not sure what's causing that?  Seems like it might be because things get called in a different order.

static/css/inline-selection.scss
ianb: It might be, I'm not sure – it could be a problem if the reset was on the root, and the site stylesheet has a specific selector that is considered more specific than reset one.  You might also have to do `& * {@extend .pageshot-reset;}` to get the reset to apply to children.  There's some examples already in that file.

fzzzy:
How can I remove the workaround for hitting the shot button twice? I already tried a couple of different approaches and failed...

static/css/inline-selection.scss
fzzzy: Ok. I'll probably just add the reset to every element.

ianb:
I mean just leave the bug in, instead of working around it.

fzzzy:
Hmm. Ok. I'll open a new bug to fix starting a new shot flow when there's already a shot in progress higher up in main.js.

Issue #1233, opened 2016-07-08
Running full `make all` build is slow


pdehaan:
Currently, running `$ make all` seems to take about 7-8 minutes locally and on Travis-CI. Not sure if there is an easy way to optimize that and make it faster, but it's kind of a drag.

```sh
$ time make all
...
make all  374.82s user 35.14s system 99% cpu 6:52.02 total
```

Here's a snippet of some of the output:

```sh
...
cp static/img/profile.png build/static/img/profile.png
cp build/static/img/profile.png build/server/static/img/profile.png
cp static/img/question-mark.svg build/static/img/question-mark.svg
cp build/static/img/question-mark.svg build/server/static/img/question-mark.svg
cp static/img/twitter-16.png build/static/img/twitter-16.png
cp build/static/img/twitter-16.png build/server/static/img/twitter-16.png
cp static/img/twitter-32.png build/static/img/twitter-32.png
cp build/static/img/twitter-32.png build/server/static/img/twitter-32.png
cp static/img/up-arrow.png build/static/img/up-arrow.png
cp build/static/img/up-arrow.png build/server/static/img/up-arrow.png
cp static/img/zoom-in.svg build/static/img/zoom-in.svg
cp build/static/img/zoom-in.svg build/server/static/img/zoom-in.svg
cp static/img/zoom-out.svg build/static/img/zoom-out.svg
cp build/static/img/zoom-out.svg build/server/static/img/zoom-out.svg
cp static/img/zoom.svg build/static/img/zoom.svg
cp build/static/img/zoom.svg build/server/static/img/zoom.svg
babel --retain-lines static/js/content-helper.js | bin/_fixup_panel_js > build/static/js/content-helper.js
cp build/static/js/content-helper.js build/server/static/js/content-helper.js
babel --retain-lines static/js/parent-helper.js | bin/_fixup_panel_js > build/static/js/parent-helper.js
cp build/static/js/parent-helper.js build/server/static/js/parent-helper.js
cp static/vendor/raven.js build/server/static/vendor/raven.js
cp -p server/src/export-shots.sh build/server/export-shots.sh
cp server/db-patches/patch-0-1.sql build/server/db-patches/patch-0-1.sql
cp server/db-patches/patch-1-0.sql build/server/db-patches/patch-1-0.sql
cp server/db-patches/patch-1-2.sql build/server/db-patches/patch-1-2.sql
[...]

PR #1267, opened 2016-07-19
r? ianbicking Use travis firefox

fzzzy:


.travis.yml
```diff
@@ -16,12 +19,14 @@ before_install:
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16 -extension RANDR"
 
 before_script:
-  - npm install -g jpm mozilla-download
+  - npm install -g jpm
   - npm install
+  - export FIREFOX_BIN=`which firefox`
```
pdehaan: Do we use the `FIREFOX_BIN` var anywhere (we export the same value below as `JPM_FIREFOX_BINARY`)?

fzzzy: dunno, it was cargo culted from activity-stream

.travis.yml
```diff
@@ -16,12 +19,14 @@ before_install:
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16 -extension RANDR"
 
 before_script:
-  - npm install -g jpm mozilla-download
+  - npm install -g jpm
```
pdehaan: Not even 💯 sure we need to globally install `jpm` here since we already have it installed as a devDependency, per [./package.json:25](https://github.com/mozilla-services/pageshot/blob/f096400e270901679450bda07f33ae9d51a3dfd9/package.json#L25).

We can probably just reference that one via `./node_modules/.bin/jpm` path, or if we refer to it in a package.json script, it will automatically prepend `./node_modules/bin/` to our `$PATH` env variable for us.

Similarly, we can probably remove the `npm install` below on .travis.yml:23 since the Travis-CI `install` hook should have run before the `before_script` hook. You can see this in action by looking at the latest Travis-CI logs:

```sh
$ npm install        # 212.85s
$ npm install -g jpm #  23.86s
$ npm install        #   7.66s
```

So as you can see, installing `jpm` a second time, adds ~24s to the build, then running `npm install` again adds another ~8s (but doesn't add any additional modules since we just finished doing the `npm install` on line 165 of the Travis-CI log).

Builds seem to take 12-15 minutes to complete though, so +/-30 seconds isn't a huge impact, but still. 🕐

ianb: Sometime we should probably try to reduce the Travis time generally – there must be tricks, I imagine.  As it is either way we'll just be letting it build and bug us if there's a problem, rather than watching it.

pdehaan: I think #1233 may be the cause of some of the slowness locally and on Travis-CI. I'd wager that being more aggressive with globs and copying directories instead of individual files would speed things up, as well as not running `babel` on each individual file. We could maybe start with concatenating a bunch of the client/server files into bundles and then only need to run babel 1-2 times.

ianb: cp definitely isn't causing any problem, it's fast.  Babel is slowish, though browserify is even slower, so I don't know if changing the order there would help – plus then we'd have to specifically opt-out browserify'd files from babel, even though in many contexts we need babel.

Though... we could probably remove babel from the server if we upgrade to Node 6.2.  I believe it has all the features we need.  And Firefox is at pretty much the same place.  So the one place we will keep needing Babel for a while is for the code delivered to the browser, which all goes through browserify.  So maybe I take that back... if we turn off babel generally, then just use it for browserify, we might speed things up really well.

pdehaan: Sure, for `cp`, but I'd still argue that this:

```sh
$ time cp *.png tmp/
# 0.00s user 0.01s system 92% cpu 0.013 total
```

Is going to be marginally faster than copying files one at a time, like this:

```sh
$ time (cp alpha-badge.png tmp2 && cp bgtexture.png tmp2 && cp chrome-store-icon.png tmp2 && cp clipboard-16.png tmp2 && cp clipboard-32.png tmp2 && cp clock.png tmp2 && cp close.png tmp2 && cp comment-bubble-open.png tmp2 && cp comment-bubble.png tmp2 && cp comment.png tmp2 && cp copy-image-16.png tmp2 && cp copy-image-32.png tmp2 && cp down-arrow.png tmp2 && cp email-16.png tmp2 && cp email-32.png tmp2 && cp facebook-16.png tmp2 && cp facebook-32.png tmp2 && cp flag.png tmp2 && cp garbage-bin.png tmp2 && cp help-recall.png tmp2 && cp help-shooter.png tmp2 && cp icon-recall.png tmp2 && cp icon-shoot-active.png tmp2 && cp icon-shoot.png tmp2 && cp link-16.png tmp2 && cp link-32.png tmp2 && cp menu.png tmp2 && cp my-shots.png tmp2 && cp pageshot-icon-128.png tmp2 && cp pageshot-icon-16.png tmp2 && cp pageshot-icon-19.png tmp2 && cp pageshot-icon-256.png tmp2 && cp pageshot-icon-32.png tmp2 && cp pageshot-icon-38.png tmp2 && cp pageshot-icon-48.png tmp2 && cp pageshot-invert-16px.png tmp2 && cp pageshot-invert-32px.png tmp2 && cp pageshot-invert-64px.png tmp2 && cp pageshot-mac-16px.png tmp2 && cp pageshot-mac-32px.png tmp2 && cp pageshot-mac-64px.png tmp2 && cp pageshot-win-16px.png tmp2 && cp pageshot-win-32px.png tmp2 && cp pageshot-win-64px.png tmp2 && cp pinterest-16.png tmp2 && cp pinterest-32.png tmp2 && cp profile-open.png tmp2 && cp profile.png tmp2 && cp twitter-16.png tmp2 && cp twitter-32.png tmp2 && cp up-arrow.png tmp2)
# 0.04s user 0.08s system 98% cpu 0.121 total
```

But yeah, point taken that the real bottleneck is browserify and tackling that would have the best performance boost.

.travis.yml
```diff
@@ -16,15 +24,22 @@ before_install:
...
+  - echo $JPM_FIREFOX_BINARY
   - make all
   - cd ..
-  - mozilla-download --branch mozilla-central --product firefox firefox
-  - export JPM_FIREFOX_BINARY=$TRAVIS_BUILD_DIR/../firefox/firefox
   - cd $TRAVIS_BUILD_DIR/
 
 script:
-  - pwd && ls && cd $TRAVIS_BUILD_DIR/build/addon && jpm test -v
+  - pwd && ls && cd $TRAVIS_BUILD_DIR/build/addon && $TRAVIS_BUILD_DIR/node_modules/.bin/jpm test -v
```
pdehaan: I'd probably start by testing a "script" in package.json that says:

```js
"jpm:test": "jpm test -v"
```

Then in the .travis.yml file, see if you can change this line to:
```sh
pwd && ls && cd $TRAVIS_BUILD_DIR/build/addon && npm run jpm:test
```

.travis.yml
```diff
@@ -16,15 +24,22 @@ before_install:
...
 before_script:
-  - npm install -g jpm mozilla-download
+  - pwd
+  - ls node_modules || true
   - npm install
+  - ls node_modules || true
+  - ls node_modules/.bin || true
+  - ls ~/node_modules || true
+  - npm config ls -l
+  - find / -name node_modules
```
pdehaan: This line seems to be breaking the build.

.travis.yml
```diff
@@ -16,15 +24,22 @@ before_install:
...
+  - export JPM_FIREFOX_BINARY=`which firefox`
+  - firefox -v
+  - echo $JPM_FIREFOX_BINARY
   - make all
   - cd ..
-  - mozilla-download --branch mozilla-central --product firefox firefox
-  - export JPM_FIREFOX_BINARY=$TRAVIS_BUILD_DIR/../firefox/firefox
   - cd $TRAVIS_BUILD_DIR/
```
pdehaan: According to https://docs.travis-ci.com/user/environment-variables#Default-Environment-Variables

> `TRAVIS_BUILD_DIR`: The absolute path to the directory where the repository being built has been copied on the worker.

Since the `TRAVIS_BUILD_DIR` is absolute, you probably don't need to `cd ..` directly before this. Or maybe don't even need this at all if you're not going up into the parent directory (especially since the first command in the `script` section does a `cd $TRAVIS_BUILD_DIR/build/addon` anyways).

PR #1274, opened 2016-07-20
r?fzzzy Db doc fixes

ianb:


server/src/server.js
```diff
@@ -319,6 +319,7 @@ app.post("/api/login", function (req, res) {
       cookies.set("user", vars.deviceId, {signed: true});
       let userAnalytics = ua(config.gaId, req.deviceId, {strictCidFormat: false});
       userAnalytics.pageview("/api/login").send();
+      // FIXME: since the client can't access the Sentry instance, should we be returning this info?
```
fzzzy: This is the public DSN, which the client can access.

ianb:
Oops, I included bad commits, will re-do...

PR #1317, opened 2016-08-09
r? ianbicking remove script-src 'unsafe-inline'

fzzzy:


ianb:
Wow, that is much more complicated than I would have expected.  Oh GA...

PR #1321, opened 2016-08-11
r? ianbicking Fixes #1313 properly tear down the ShotContext when pressing the cancel button in the overlay menu.

### LINKED ISSUE #1313: "Cancel" keeps PageShot from working on page
To reproduce:

* Open a regular page
* Start PageShot
* Hit Cancel
* Start PageShot again on the same page

The console will have the message `Tried to take a shot while a shot was already in progress.`

fzzzy:


ianb:
This won't work for Chrome, but we'll have to do a second set of fixes for Chrome anyway

PR #1332, opened 2016-08-12
r? ianbicking Fixes #1314 Go directly to the select screen, add a my shots button there, and polish the save/cancel buttons

### LINKED ISSUE #1314: Visual polish on the interstitial modal & buttons
Comments from meeting today

- interstitial modal is odd now that there's only one thing to do.
- reducing clicks could be an option here to fix this.
- polish on save/cancel buttons.

fzzzy:


static/css/inline-selection.scss
```diff
@@ -282,12 +286,56 @@ $overlay-z-index: 900000;
...
+  right: 15px;
+  height: 38px;
+  width: 130px;
+  padding-right: 15px;
+  vertical-align: middle;
+  background-image: none;
+  font-weight: normal;
+  border-radius: 4px;
+  border: none;
+  background: #dddddd;
```
ianb: Apparently sass-lint really wants this to be `#ddd`

fzzzy: Oh yeah that has tripped me up a few times...

addon/data/selector-ui.js
```diff
@@ -31,15 +31,31 @@ const ui = (function () { // eslint-disable-line no-unused-vars
...
+    display: function (callbacks) {
       if (! this.overlayEl) {
         this.overlayEl = document.createElement("div");
+        this.overlayEl.className = "pageshot-preview-overlay";
         let instructions = document.createElement("div");
         instructions.className = "pageshot-preview-instructions";
-        instructions.textContent = "Click an element to select, drag a region to select, or press ESC to cancel.";
+        instructions.textContent = "Drag or click on the page to select a region. Press ESC to cancel.";
         this.overlayEl.appendChild(instructions);
+        if (callbacks && callbacks.onOpenMyShots !== undefined) {
```
ianb: Why is there a conditional here?  Is there a case where this is being called without the callbacks?

fzzzy: There used to be, but I got rid of it. I'll get rid of the conditional too.

addon/data/shooter-interactive-worker.js
```diff
@@ -273,6 +281,9 @@ stateHandlers.crosshairs = {
   },
 
   mousedown: function (event) {
+    if (event.target.className === "pageshot-myshots") {
+      return;
+    }
```
ianb: Will this work if you click on an element inside `.pageshot-myshots` (such as the `<span class="pageshot-myshots-text">...</span>`)?  The isHeader function below makes a more thorough check.

fzzzy: I originally added this in the isHeader method, but I couldn't tell if that was supposed to be only for Google Chrome. Is it?

ianb: The functionality was only needed in Chrome because it had the in-content top bar, but the idea is the same.  Maybe move `isHeader` into a standalone function that handles both cases?

fzzzy: Sounds good.

ianb:
r+ with comments addressed

static/css/inline-selection.scss
```diff
@@ -282,12 +286,56 @@ $overlay-z-index: 900000;
...
+  font-size: 12pt;
   text-align: center;
   padding-top: 1.5em;
+  padding-left: 150px;
+  padding-right: 150px;
   color: #fff;
   z-index: 999999999;
 }
+.pageshot-myshots {
+  display: block;
```
ianb: This probably needs to use `@extend .pageshot-reset;`

PR #1367, opened 2016-08-17
r? ianbicking Fixes #1351 Track the back button using the pageshow event (ready only fires for pages not in the cache) and clean up tabsBeingShot

### LINKED ISSUE #1351: Back button leads to exception
Need to work on reproduction, but I think this was a case of navigating, take shot, then go back (to the non-shot page), and I got this exception:

```
console.warn: mozilla-pageshot: Error in bound  : Error: Couldn't find the worker to receive this message. The script may not be initialized yet, or may already have been unloaded.
console.error: mozilla-pageshot: 
JPM [error]   Message: Error: Couldn't find the worker to receive this message. The script may not be initialized yet, or may already have been unloaded.
  Stack:
    send@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/content/worker.js:82:13
...
```

fzzzy:


ianb:
Does `tabs.on("pageshow", (tab) => {...})` not work?

fzzzy:
Nope, tried that first.

PR #1368, opened 2016-08-17
r?fzzzy Delete account

ianb:


server/src/pages/delete-account/server.js
```diff
@@ -0,0 +1,39 @@
+const express = require("express");
+const reactrender = require("../../reactrender");
+const { Shot } = require("../../servershot");
+const { addDeviceActivity } = require("../../users");
+
+let app = express();
+
+exports.app = app;
+
+app.get("/", function (req, res) {
```
fzzzy: I think this was copy pasted from the shot index? I don't think / is needed in delete-account

server/src/pages/shotindex/view.js
```diff
@@ -47,6 +48,9 @@ class Body extends React.Component {
         <div id="shot-index">
           {children}
         </div>
+        <Footer>
+          <a href="/delete-account">Disable account / delete all shots</a>
```
fzzzy: I think Delete account is a better thing to say. Disable sounds like it can be re-enabled.

server/src/pages/delete-account/server.js
fzzzy: Ok, I see now that this / is relative to delete-account, and /delete below is actually /delete-account/delete. I guess only the error message in the html needs to change to "You must have the addon installed to delete your account"

PR #1376, opened 2016-08-18
r? ianbicking Fixes #1167 Go to the my shots page if you try to click the page shot button when looking at about:blank, about:newtab, or activity center

### LINKED ISSUE #1167: PageShot button on about:newtab should open My Shots
Right now Chrome about pages aren't supported at all.  Which is fine, but then you can't get to the My Shots button.

fzzzy:


ianb:
I believe this opens My Shots in a new tab, but in this case it should open it in the current tab.

fzzzy:
Good point.

fzzzy:
Ok, I switched it to open in the same tab.

PR #1402, opened 2016-08-22
Use growl instead of playing a sound to notify about builds.

fzzzy:


bin/run-addon
```diff
@@ -100,13 +100,13 @@ while [[ -n "$1" ]] ; do
...
 if [[ -n "$upload" ]] ; then
-  playsound bin/elevator-ding-reverse.mp3
+  which growlnotify > /dev/null && growlnotify -m "Page Shot Add-on build started." || echo "Page Shot Add-on build started."
   make addon xpi
   echo "Uploading to the browser"
   echo "  (if this fails, install: https://addons.mozilla.org/en-US/firefox/addon/autoinstaller/)"
   # Doesn't return any content, leads to an unnecessary error:
   wget --post-file=build/mozilla-pageshot.xpi http://localhost:8888 || echo "no content"
-  playsound bin/elevator-ding.mp3
+  which growlnotify > /dev/null && growlnotify -m "Page Shot Add-on build finished." || echo "Page Shot Add-on build finished."
```
pdehaan: Do we need to remove `playsound()` above on L60:

```sh
playsound () {
  if [[ -e /usr/bin/afplay ]] ; then
    afplay $1 &
  fi
}
```

bin/run-server
```diff
@@ -13,11 +13,11 @@ fi
 
 run () {
   if [[ -n "$restart" ]] ; then
-    playsound bin/hole-punch-reverse.mp3
+    which growlnotify > /dev/null && growlnotify -m "Page Shot Server build started." || echo "Page Shot Server build started."
   fi
   make server
   if [[ -n "$restart" ]] ; then
-    playsound bin/hole-punch.mp3
+    which growlnotify > /dev/null && growlnotify -m "Page Shot Server build finished." || echo "Page Shot Server build finished."
```
pdehaan: Do we need to remove `playsound()` below on L32-36?

pdehaan:
Possibly also need to remove from [./bin/run-chrome-builder:14-18](https://github.com/mozilla-services/pageshot/blob/bff7f0b6104e0b6a484f9634278112dcb8b95e0f/bin/run-chrome-builder#L14-L18) and [./bin/run-chrome-builder:22](https://github.com/mozilla-services/pageshot/blob/bff7f0b6104e0b6a484f9634278112dcb8b95e0f/bin/run-chrome-builder#L22)?

fzzzy:
Addressed pdehaan's comments.

ianb:
Might be good to put a note in README, since I assume from the code that this is effectively an optional dependency

fzzzy:
I actually did put a note in the readme :)

ianb:
pffbt, I was looking at the comments instead of the changes :P

PR #1409, opened 2016-08-23
r? ianbicking Fixes #1406 Make a quick and dirty hidden install page.

### LINKED ISSUE #1406: Make a hidden page with an install button

fzzzy:


static/homepage/install.html
```diff
@@ -0,0 +1,17 @@
...
+html {
+  font-family: sans-serif;
+}
+    </style>
+  </head>
+  <body>
+    <center>
+      <img src="/static/img/pageshot.svg" height="256" width="256" />
+      <h1>Page Shot</h1>
+      <a href="https://testpilot.firefox.com/files/pageshot/pageshot-0.1.201608231911.xpi">Install</a>
```
pdehaan: Personal curiosity, but would it be possible to host the add-on on the local PageShot server?
That way we'd be able to go to dev/stage/prod servers and get the appropriate latest add-on that matches the server.

fzzzy: it already is, at /xpi/mozilla-pageshot.xpi

ianb:
The domain is whitelisted in Firefox, so having it on testpilot.firefox.com is advantageous

PR #1444, opened 2016-08-31
fix legal pages and delete account page

johngruen:


server/src/pages/delete-account/controller.js
```diff
@@ -1,5 +1,5 @@
 /* globals sendEvent */
-//const page = require("./page").page;
+const page = require("./page").page;
```
johngruen: Not actually sure why i did this, it doesn't look necessary here.

PR #1456, opened 2016-09-02
r?fzzzy Selection fixes

ianb:


fzzzy:
There are a few debug prints that should be removed, but other than that r+

ianb:
Removed and merged

PR #1465, opened 2016-09-06
r? ianbicking Fixes #1256 Check to see if test pilot is installed at page shot install time, and record it in ga.

### LINKED ISSUE #1256: Use `getAddonByID` to determine if Test Pilot is installed. Send a GA event if not.

fzzzy:


ianb:
Thanks for also keeping METRICS.md up to date!

PR #1472, opened 2016-09-07
r? ianbicking Fixes #1323 Add a test.

### LINKED ISSUE #1323: Add a test

fzzzy:


package.json
```diff
@@ -46,10 +39,22 @@
   },
   "devDependencies": {
```
ianb: Do we have anything to make sure all dev dependencies get installed?  Or does that happen automatically with `npm install`?

fzzzy: It happens automatically on npm install. They are only not installed on npm install --production apparently.

pdehaan: The `devDependencies` should be installed by default, except for in the case of:

> With the `--production` flag (or when the `NODE_ENV` environment variable is set to `production`), npm will not install modules listed in `devDependencies`.

&mdash; via https://docs.npmjs.com/cli/install#description

**TL;DR:** nope, should be installed by default unless you're doing `npm install --production` or `NODE_ENV=production npm install`.

PR #1493, opened 2016-09-08
r? ianbicking Fixes #1492 Remove use of dangerouslySetInnerHTML

### LINKED ISSUE #1492: Remove last usages of dangerouslySetInnerHTML
There are two that are dead code, and the Raven activation should be moved to a script url instead of an inline script.

fzzzy:


server/src/views/export.js
```diff
@@ -88,9 +88,7 @@ class Status extends React.Component {
       <div id="container">
         <h1>Wait for export to be created...</h1>
         <p>Please stay on this page (which will keep checking for when your export is ready)</p>
-        <script dangerouslySetInnerHTML={{__html: `
-          setTimeout(function () {location.reload();}, 3000);
-        `}} />
+        <script src={ req.staticLinkWithHost("js/auto-refresh.js") } />
```
pdehaan: ESLint sayeth:

```
/home/travis/build/mozilla-services/pageshot/server/src/views/export.js
  91:23  error  'req' is not defined  no-undef
```

server/src/views/frame.js
```diff
@@ -283,7 +281,7 @@ class Head extends React.Component {
         {oembed}
         {ogTitle}
         {ogImage}
-        <script nonce={ this.props.cspNonce } dangerouslySetInnerHTML={{__html: `var CONTENT_HOSTING_ORIGIN = "${postMessageOrigin}";`}}></script>
+        <script src="/set-content-hosting-origin.js" />
```
ianb: Does this mean the nonce stuff has all been removed?

fzzzy: No, the nonce is still used in a few places where strings are concatenated with scripts in them (in the react bootstrap process). But it is not used anywhere in jsx any more.

PR #1520, opened 2016-09-13
r?fzzzy Fix 1516

ianb:


ianb:
And an add-on but unrelated commit because I was lazy...

PR #1521, opened 2016-09-13
r?fzzzy Fix #1453, add cd0/cd1 dimensions to selection events

### LINKED ISSUE #1453: Pass selection dimension with GA events
This would be an `eventValue` (as noted [here](https://support.google.com/analytics/answer/1033068#Anatomy)).  Since it's just a single integer, maybe we'd want to do x+y as the number (or x*y?)

Or we can use [custom dimensions](https://developers.google.com/analytics/devguides/collection/analyticsjs/custom-dims-mets)

ianb:
Fix how event options (eventValue, cd0/cd1) are sent to server; previously eventValue was being thrown away.

### DIFF
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -290,11 +290,17 @@ app.post("/event", function (req, res) {
   }
   hashUserId(req.deviceId).then((userUuid) => {
     let userAnalytics = ua(config.gaId, userUuid.toString());
-    userAnalytics.event(
-      bodyObj.event,
-      bodyObj.action,
-      bodyObj.label
-    ).send();
+    let params = Object.assign(
+      {},
+      bodyObj.options || {},
+      {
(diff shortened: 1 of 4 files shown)


server/src/server.js
```diff
@@ -290,11 +290,17 @@ app.post("/event", function (req, res) {
...
   hashUserId(req.deviceId).then((userUuid) => {
     let userAnalytics = ua(config.gaId, userUuid.toString());
-    userAnalytics.event(
-      bodyObj.event,
-      bodyObj.action,
-      bodyObj.label
-    ).send();
+    let params = Object.assign(
+      {},
+      bodyObj.options || {},
```
fzzzy: I was curious about this, so I looked it up: Object.assign does not throw an exception for null or undefined source values, so you don't actually need the `|| {}` here.

Issue #333, opened 2015-04-23
Include padding around autoselect


ianb:
The autoselect selection is often too tight around the selection, it should be a little bit bigger, even 2px would give it a little room.

ianb:
I played around with some different padding, but not much really worked.  I'm going to assume that if doing a manual selection, the user will select what they want.  Only in the autoselect cast, the left margin was chronically crowded.  I also added just 1px margin to the top of the selection, though typically the top of letters doesn't align as closely with the edge as the left side of letters do.

PR #1529, opened 2016-09-15
r? ianbicking Fixes #1508 Add a download button between the share and trash buttons.

### LINKED ISSUE #1508: Add a Download link on the frame page between share and delete
I will provide an SVG icon shortly

fzzzy:


static/img/download.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:xlink="http://www.w3.org/1999/xlink"
+     width="16"
+     height="16"
+     viewBox="0 0 16 16">
+  <style>
+    path {
+      fill: #000;
```
johngruen: Can you change this to #333 to match the garbage icon

johngruen: @fzzzy er #4d4d4d;

fzzzy: done

server/src/server.js
```diff
@@ -645,6 +645,14 @@ app.get("/images/:imageid", function (req, res) {
...
       }
       res.header("Content-Type", obj.contentType);
+      if (req.query.filename) {
+        let extension = "." + obj.contentType.split("/").pop();
+        let filename= req.query.filename;
+        if (! filename.endsWith(extension)) {
+          filename = filename + extension;
+        }
+        res.header("Content-Disposition", `attachment; filename="${filename}"`)
+      }
```
ianb: This allows a potential attacker to construct a link that makes someone download the image with whatever filename they wish (though with the necessary extension).  We shouldn't allow that.

I don't think we need any server support for this, the [download](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) attribute is sufficient for what we're doing here.

fzzzy: We're not using an <a> tag.

ianb: Is there any reason not to?

fzzzy: Because the styles apply to buttons. Maybe the styles will apply to an a tag without a problem.

ianb: Example of an attack on this code: `?filename=attack.xpi%3bdummy=` – better to skip the attack location entirely.

fzzzy: Can we sign the filename with a hash that uses a salt that only the server knows? The shot resource knows about the shot creation date, and the image resource knows about the content type (and thus the file extension), but neither knows enough to construct the entire filename by itself.

ianb: This seems like a lot of trouble, and you have to pass around the signed value in the model, and maybe the links expire... can't we just give the content type to the page so it can construct a proper name?  Probably the type should be in the clip object itself, which would be fairly easy to add – right now it's always PNG, though I'd like to allow for JPEG eventually.

fzzzy: You wouldn't have to pass anything around in the model. The server would take the filename and the salt and generate the hash, and compare it with the hash provided by the client.

If the button styles apply to an anchor tag and we switch to using the download attribute, we'd still have to include both the uuid and the filename in the url (because the filename isn't enough to uniquely identify the image, and the image uuid isn't enough information to generate the filename), so the same attack would still work.

ianb: You have to pass the generated URL through the model.

The download attribute works like  `<a href="uuid-link" download="filename">` with `download` as a suggestion, and the browser handles making sure it is cleaned up.  So it's simpler and only requires switching to an `<a>` tag, which is more semantic anyway.

fzzzy: Hmm, ok, good. That works. I thought the download attribute was more like download=true. I should have looked closer.

fzzzy:
Ok, that is significantly nicer than Content-Disposition. Take a look.

PR #1568, opened 2016-09-20
Switch Twitter meta tags from 'property' to 'name'

pdehaan:
Ref: https://github.com/mozilla-services/pageshot/issues/1467#issuecomment-248410458


pdehaan:
Once the new XPI gets deployed to dev, I can re-run the Facebook/Twitter validators and see if that removes any warnings.

PR #1587, opened 2016-09-22
r? ianbicking Fix build-docker-image bustage from the xpi name changing

fzzzy:


ianb:
The XPI name shouldn't have changed?  John had a commit that changed it, but he removed that before it got merged

fzzzy:
Hm, well it seems the xpi is still being built at mozilla-pageshot.xpi, but the problem is that jpm sign is now outputting the xpi at something matching page_shot-*-fx.xpi.

fzzzy:
we are reverting the change that would require this

PR #1707, opened 2016-10-05
r?fzzzy Misc fixes

ianb:


addon/lib/main.js
```diff
@@ -177,9 +177,11 @@ exports.onUnload = function (reason) {
       deviceInfo: JSON.stringify(deviceInfo)
     }
   });
-  // We can't rely on the server removing cookies, because the request may not
-  // complete locally, since the addon is being destroyed
-  removeCookies();
```
fzzzy: doh!

ianb: I actually doubt this was the problem, but it's an attempt to be defensive.  The cookies should be removed on upgrade, then the new version is installed, and login happens and the cookies get re-added.  So this change is a just-in-case

PR #1754, opened 2016-10-17
Use template literals

sadika9:


ianb:
That's some good cleanup, thanks @sadika9 !

PR #1841, opened 2016-10-25
r?fzzzy Refactor frame

### LINKED ISSUE #1713: Change frame.js to use reactruntime
This would mean changing frame.js to use the Head/Body stuff, making a page and moving the resources into that page.  Some items might get moved out of server.js and into a page-specific server.js (though it's unclear what parts); the server would have to be mounted on "/".

ianb:
As a whole this fixes #1713

Tried to split up refactoring steps into multiple commits

page/shot/view.js is still pretty hairy, but I moved a couple things out.

I am pretty sure I got everything from #1837 reapplied.  The Share button wraps again though, but will make a followup ticket.


server/src/caching.js
```diff
@@ -7,6 +7,7 @@ if (! config.setCache) {
 }
 
 exports.setCache = function (res, options) {
+  console.log("setcache?", config.setCache, process.env.SET_CACHE);
```
fzzzy: This is probably useful to leave in, but it should be an info. Or take it out if you don't think we always need to see this.

server/src/pages/shot/controller.js
```diff
@@ -1,51 +1,46 @@
...
 
   if (firstSet) {
     document.addEventListener("helper-ready", function onHelperReady(e) {
       document.removeEventListener("helper-ready", onHelperReady, false);
       let event = document.createEvent("CustomEvent");
       event.initCustomEvent("page-ready", true, true, null);
       document.dispatchEvent(event);
-      requestProfile();
-      requestHasSavedShot(model.shot.id);
+      //requestHasSavedShot(model.shot.id);
```
fzzzy: I think this should just be removed, to clean up the codebase.

server/src/pages/shot/controller.js
```diff
@@ -1,51 +1,46 @@
...
     }, false);
-    document.addEventListener("refresh-profile", refreshProfile, false);
+    /*
```
fzzzy: Ditto.

server/src/pages/shot/controller.js
```diff
@@ -192,6 +165,7 @@ function sendShowElement(clipId) {
   }
 }
 
+/*
```
fzzzy: Remove.

server/src/pages/shot/view.js
```diff
@@ -350,7 +166,7 @@ class Frame extends React.Component {
   }
 
   onClickUploadFullPage(e) {
```
fzzzy: At this point maybe we should remove more code related to old features that are no longer implemented? The product is firming up a lot and it's unlikely that we will implement new features by using old code, but we will implement them with new code instead.

server/src/server.js
```diff
@@ -641,6 +641,7 @@ app.post("/api/delete-shot", function (req, res) {
   });
 });
 
+/*
```
fzzzy: Let's just remove this too...

server/src/server.js
```diff
@@ -822,31 +824,8 @@ app.get("/contribute.json", function (req, res) {
 });
 
 // FIXME: this can't the right way to do this...
```
fzzzy: Great, the non-right way of doing this is no longer the way we do it! Remove the fixme and the next commented line.

ianb:
I addressed the comments except code removal – per my comment above, I'd like to deal with that work separately, as it is a wider scope than the frame.js refactoring.

server/src/pages/shot/controller.js
fzzzy: Ok.

server/src/pages/admin/server.js
```diff
@@ -7,9 +7,10 @@ let app = express();
 exports.app = app;
 
 // FIXME: this should be an FxA id, but due to a bug we're using device IDs
-let authorizedAccountIds = ['anon63daec35-fb29-8d49-ac6f-af7763523c0a'];
+let authorizedAccountIds = ['anon63daec35-fb29-8d49-ac6f-af7763523c0a', 'anon202dc00'];
 
 app.use(function (req, res, next) {
+  req.deviceId = 'anon202dc00'
```
fzzzy: Hmm, I don't think this is right!

PR #1866, opened 2016-10-30
Trim clipFilename when it's longer than 255 bytes. Fixes #1820

### LINKED ISSUE #1820: Generated file name can be too long
The title web page can be long.
Hence, the generated file name when downloading the shot (`download` attribute in the download button) can be too long and be rejected when saving in some file systems.
Also, as the title starts to be long, the shot-info div starts to grow and, at some point, it doesn't fit on the width of the browser and the user has to scroll horizontally to find the Share, Download and Delete buttons.
I'll upload an image to better show what I mean.

I'd be happy to help with this :)

Akronix:


ianb:
Looks good, thanks @Akronix !

PR #1872, opened 2016-10-31
r?fzzzy Metrixy things

ianb:


fzzzy:
The build actually failed because of some css lint things this time, instead of a timeout or something.

PR #1911, opened 2016-11-04
r?fzzzy Fix autoselect

ianb:


ianb:
No changelog change really needed, these are all intra-release regressions I believe

PR #2067, opened 2017-01-03
f? ianbicking This is just a suggestion, but I think it would be better if in the absence of the new environment variable the server would still run the controller tasks

fzzzy:


ianb:
Yeah, I think this is better too – I felt a little funny about how it was setup before when I wrote it, but didn't actually do anything with that feeling ;)

PR #2074, opened 2017-01-05
r? ianbicking Fixes #2055 Add .png extension to the filename of the downloaded file.

### LINKED ISSUE #2055: Release 4.0 Deneb

fzzzy:


ianb:
I thought the extension was automatically added?  At least currently this is used with `<a download={shot.filename}>` where I'm pretty sure the extension will be added.

fzzzy:
I tested on mac and it was not added. Cosmin tested on windows and it is not. So I have no idea why it isn't getting added automatically?

ianb:
Is this only with the download button during selection, or is it also on the download button in the shot page?  I thought we tested on the shot page, and this method will affect both locations.

fzzzy:
Yes, it was for the download button during selection. You're right, the download button on the shot page adds png. It is probably better to put it in the filename getter rather than having multiple places in the code hardcode adding png, in case shots become capable of being saved in other formats at some point.

ianb:
Note you don't need string interpolation once you remove the extension

ianb:
Otherwise r+

PR #2092, opened 2017-01-10
r? ianbicking Fixes #2051 Allow specifying the max db level with a MAX_DB_LEVEL environment variable.

### LINKED ISSUE #2051: Rearchitect the db upgrade so we can roll back properly
This is part 2 of #1977.

fzzzy:


fzzzy:
Hah, we might as well use your version then.

Issue #2082, opened 2017-01-05
Do A/B test on My Shots button


ianb:
In this test we want to present the My Shots button in different ways, and test if people are more likely to click the button because of how it is presented.

First step is to collect a couple different presentation styles for My Shots.

ianb:
I'll start the work by adding a class to the container (the element that contains My Shots) based on the A/B test and implement some placeholder style, then John can go in and do some styling.

ianb:
@johngruen if you check out the ab-myshots-button branch you'll get the test.

You can edit `server/src/ab-tests.js` to change the available options for the test.

To force a test, start the server with something like:

```sh
$ FORCE_AB_TESTS='styleMyShotsButton=bright' ./bin/run-server
```

The name of the test gets added to the iframe document body, like `.styleMyShotsButton-{testValue}` so you can style things like:

```css
.styleMyShotsButton-bright .pageshot-myshots {
  font-size: 200%;
}
```

(I've put that style into the commit as a placeholder)

Issue #2064, opened 2016-12-28
After saving a shot, the Delete, Download and Share buttons are not all in view if the shot title is too long


SoftVision-CiprianMuresan:
[Affected versions]:
- Nightly 53.0a1
- Mozilla Page Shot 4.0.201612191249

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Page Shot add-on was previously installed on a new profile from https://pageshot.dev.mozaws.net/homepage/install-test-local.html.
- Resize the browser window to half size and open the Bookmarks sidebar (Ctrl+B).

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to http://www.imfdb.org/wiki/Django_Unchained.
2. Save a shot of the page.
3. Observe the top right corner of the newly opened page.
 
[Expected result]:
- The Page Shot Delete, Download and Share buttons are in view.

[Actual result]:
- Not all the buttons are in view.

[Notes]:
- If you click the Share button, the dropdown opens under the button that is in view instead of opening under the Share button.
- The issue is also reproducible on the Test Pilot version of the add-on.
- Attached a screen recording of the issue:
![delete download and share buttons are not all in view if the shot title is too long](https://cloud.githubusercontent.com/assets/18304511/21525229/8d87f656-cd24-11e6-9c32-1df21167f1a0.gif)

Issue #1792, opened 2016-10-19
make shot page responsive


johngruen:
30% of shares are viewed on mobile. We should have a mobile responsive view for this sort of page:

https://pageshot.net/ELQi9ixejX33Xc6k/accounts.firefox.com

youwenliang:
I quickly made a mockup for the shared page mobile view: https://mozilla.invisionapp.com/share/3S90SSAH8 

I added an "image focus" mode where all the UI disappear and showing only the image (And users can zoom in & out with finger gestures)
Any thoughts? @johngruen

johngruen:
HEy @youwenliang I think the only thing this is missing is a bit of branding...We want people to know they're looking at a Page Shot image :)

ianb:
Do we need sharing in the mobile view, or can we rely on people using the native sharing techniques?

johngruen:
@ianb @youwenliang i would prefer if we used native sharing UI here as well

youwenliang:
Thanks for the feedback @ianb @johngruen 
I've updated the mockup: https://mozilla.invisionapp.com/share/2W93B20U8
(Add back "made with pageshot" and add native sharing UI for both Android & iOS)

johngruen:
Moving sprints

PR #2119, opened 2017-01-20
add bright button a/b test

johngruen:
This actually fixes all of my elnath issues

### DIFF
--- a/addon/data/selector-ui.js
+++ b/addon/data/selector-ui.js
@@ -205,7 +205,7 @@ const ui = (function () { // eslint-disable-line no-unused-vars
         <div class="pageshot-preview-instructions">
           Drag or click on the page to select a region. Press ESC to cancel.
         </div>
-        <div class="pageshot-myshots">
+        <div class="pageshot-myshots pageshot-myshots-button">
           <div class="pageshot-pre-myshots"></div>
           <div class="pageshot-myshots-text">My Shots</div>
           <div class="pageshot-post-myshots"></div>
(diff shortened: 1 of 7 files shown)


johngruen:
Fixes #2082
Fixes #1792 
Fixes #2064

johngruen:
Also cleans up all the scss-lint warnings except the inline-selection page

ianb:
I took a shot of [this page](https://github.com/mozilla-services/pageshot/issues/2120) and got this:

image

It didn't take the title of the page, and there's a big empty space to the left of the title.

In general I realize there's an elastic space to the left of the title, instead of the right.

Aha, looking more closely for some reason the extracted Open Graph title of the page is `mozilla/page-shot`, though that's not what it should be.  I'll open a separate bug for that (#2121).  Still the whitespace is a problem.

Responsive design mode in Firefox looks okay, in Chrome it doesn't work at all, though I'm going to assume that's a Chrome problem:

image

I assume it was your intention to change the color of all the buttons, but that's not what the test description describes:

image

My only blocking concern is the whitespace issue, though being clearer about the A/B test description would also be good.  The description is in `server/src/ab-tests.js`

johngruen:
@ianb huh, yeah, i can change just the one button

also not sure what's going on with the responsive stuff, will investigate

johngruen:
@ianb force pushed changes to iframe button colors and the broken title layout on the shot page. WRT responsive view might be nice to deploy this to stage to actually debug with mobile devices. I'm guessing (hoping) that there's a chrome bug happening.

johngruen:
one sec...think i figured out the breakage

ianb:
I'm still seeing the whitespace when the window is wide:

image

PR #2131, opened 2017-01-25
r?fzzzy Script release

ianb:


bin/commit-new-version
```diff
@@ -0,0 +1,90 @@
...
+base="$(dirname $bin)"
+
+if [[ -z "$ORIGINAL_SCRIPT" ]] ; then
+  # Because this command uses git, we have to make sure we're only running the
+  # non-version-tracked copy of this script
+  copy_filename="$bin/.$(basename $0)-copy"
+  cp "$0" "$copy_filename"
+  chmod +x "$copy_filename"
+  export ORIGINAL_SCRIPT="$0"
+  exec $copy_filename "$@"
```
fzzzy: Will this result in both the original script and the new script running? Shouldn't there be an exit here?

fzzzy: No, exec replaces the current process.

Issue #2199, opened 2017-02-22
Convert Page Shot to a WebExtension


ckprice:

ckprice:
PR in #2185

ianb:
It is landed!

PR #2224, opened 2017-02-27
follow up for become an WebExtension

yfdyh000:
Follow up #2199.

Remaining:
1. `"main"` in /package.json, need to change to `webextension/background/main.js`? or remove?
2. Not understand the `"build/shot.js",` in webextension/manifest.json, irrelevant entry?

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,3 @@
-addon/data/raven.js
-addon/data/vendor
-addon/lib/notificationbox.js
 static/homepage/js
 static/vendor
 /scratch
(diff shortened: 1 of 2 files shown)


ianb:
Ah, I forgot .eslintrc existed!

1. I didn't notice `"main"` before – I'm guessing it makes no sense for our package (you can't run the add-on through Node anyway), so I'll remove that.
2. build/shot.js is built from shared/shot.js into webextension/build/shot.js (the file is shared between server and add-on, but with different modules)

PR #2225, opened 2017-02-27
r? _6a68 or dcoates Update and freshen up READMEs

ianb:


6a68:
@ianb looks good. I think it'd make sense to just say that it'll ship in FF 54, but your call. Feel free to self-merge (unless that's not typically a thing on pageshot?)

ianb:
@johngruen wanted to make a more proper announcement about 54, even though the intention is leaking out in a bunch of places.  Usually on r+ the reviewer also merges and deletes the branch (but I'll merge this one).

PR #2234, opened 2017-02-28
r?fzzzy Config backend

ianb:
Technically I haven't tested due to S3 probably keeping dockerhub from working, but it seems to get as far as uploading


ianb:
woops, messed something up

Issue #2241, opened 2017-02-28
Audit code for error catching


ianb:
Any promise without .catch() should be wrapped in `catcher.watchPromise()`.  Any function called by external code (e.g., event handler) should be wrapped in `catcher.watchFunction()`.  Some known problems should be handled explicitly with `catcher.unhandled()` (especially auth and request errors).

PR #2242, opened 2017-02-28
r? fzzzy or _6a68 Error catching

ianb:
Gets error catching infrastructure back in place.  Doesn't actually catch all those errors (followup for that in #2241)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,4 @@ static/vendor
 /scratch
 /Profile
 /build
+/webextension/build
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,8 @@ partials_source := $(wildcard static/css/partials/*.scss)
 imgs_source := $(wildcard static/img/*)
 imgs_server_dest := $(imgs_source:%=build/server/%)
 
+raven_source := $(shell node -e 'console.log(require.resolve("raven-js/dist/raven.js"))')
+
 ## General transforms:
 # These cover standard ways of building files given a source
 
@@ -83,7 +85,7 @@ build/%.html: %.html
 	cp $< $@
 
 .PHONY: addon
-addon: npm set_backend webextension/manifest.json webextension/build/shot.js webextension/build/inlineSelectionCss.js
+addon: npm set_backend webextension/manifest.json webextension/build/shot.js webextension/build/inlineSelectionCss.js webextension/build/raven.js
 
 .PHONY: zip
 zip:
@@ -109,6 +111,10 @@ webextension/build/shot.js: shared/shot.js
 webextension/build/inlineSelectionCss.js: build/server/static/css/inline-selection.css
 	@mkdir -p $(@D)
 	./bin/build-scripts/css_to_js inlineSelectionCss $< > $@
+
+webextension/build/raven.js: $(raven_source)
+	@mkdir -p $(@D)
+	cp $< $@
 
 ## Server related rules:
 
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ build/%.html: %.html
 addon: npm set_backend webextension/manifest.json webextension/build/shot.js webextension/build/inlineSelectionCss.js webextension/build/raven.js
 
 .PHONY: zip
-zip:
+zip: addon
 	# FIXME: should remove web-ext-artifacts/*.zip first
 	./node_modules/.bin/web-ext build --source-dir webextension/
 	mv web-ext-artifacts/page_shot*.zip build/pageshot.zip
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,7 @@ all: addon server
 
 .PHONY: clean
 clean:
-	rm -rf build/ webextension/build/
+	rm -rf build/ webextension/build/ webextension/manifest.json
 
 .PHONY: help
 help:
--- a/webextension/background/analytics.js
+++ b/webextension/background/analytics.js
@@ -1,26 +1,33 @@
-/* globals main, auth */
+/* globals main, auth, catcher */
 
 window.analytics = (function () {
   let exports = {};
 
   exports.sendEvent = function (action, label) {
-    let eventCategory = "addon";
-    let url = main.getBackend() + "/event";
-    let req = new XMLHttpRequest();
-    req.open("POST", url);
-    req.setRequestHeader("content-type", "application/json");
-    req.onload = () => {
-      if (req.status >= 300) {
-        console.warn("Event gave non-2xx response:", req.status);
-      }
-    };
-    // FIXME: add cdX and other details from req.js
-    req.send(JSON.stringify({
-      deviceId: auth.getDeviceId(),
-      event: eventCategory,
-      action,
-      label
-    }));
+    return new Promise((resolve, reject) => {
+      let eventCategory = "addon";
+      let url = main.getBackend() + "/event";
+      let req = new XMLHttpRequest();
+      req.open("POST", url);
+      req.setRequestHeader("content-type", "application/json");
+      req.onload = () => {
+        if (req.status >= 300) {
+          let exc = new Error("Bad response from POST /event");
+          exc.status = req.status;
+          exc.statusText = req.statusText;
+          reject(exc);
+        } else {
+          resolve();
+        }
+      };
+      // FIXME: add cdX and other details from req.js
+      req.send(JSON.stringify({
+        deviceId: auth.getDeviceId(),
+        event: eventCategory,
+        action,
+        label
+      }));
+    });
   };
 
   return exports;
--- a/webextension/background/analytics.js
+++ b/webextension/background/analytics.js
@@ -10,7 +10,7 @@ window.analytics = (function () {
       let req = new XMLHttpRequest();
       req.open("POST", url);
       req.setRequestHeader("content-type", "application/json");
-      req.onload = () => {
+      req.onload = catcher.watchFunction(() => {
         if (req.status >= 300) {
           let exc = new Error("Bad response from POST /event");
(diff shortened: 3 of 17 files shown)


webextension/background/senderror.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+    if (! Raven.isSetup()) {
+      Raven.config(e.sentryPublicDSN).install();
+    }
+    delete e.sentryPublicDSN;
+    let exception = new Error(e.message);
+    exception.stack = e.multilineStack || e.stack || undefined;
+    let rest = {};
+    for (let attr in e) {
+      if (["name", "message", "stack", "multilineStack", "popupMessage", "version", "sentryPublicDSN", "help"].indexOf(attr) == -1) {
```
pdehaan: Nit: `===`

Or, you could use [`Array.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes), which returns a Boolean.

Makefile
```diff
@@ -110,6 +112,10 @@ webextension/build/inlineSelectionCss.js: build/server/static/css/inline-selecti
 	@mkdir -p $(@D)
 	./bin/build-scripts/css_to_js inlineSelectionCss $< > $@
 
+webextension/build/raven.js: $(raven_source)
+	@mkdir -p $(@D)
+	cp $< $@
```
pdehaan: It looks like we're getting some ESLint errors:

```
/home/travis/build/mozilla-services/pageshot/webextension/build/raven.js
    13:128  error    'define' is not defined                                              no-undef
    13:140  error    'define' is not defined                                              no-undef
    13:337  warning  'define' is defined but never used                                   no-unused-vars
    13:344  warning  'module' is defined but never used                                   no-unused-vars
    13:351  warning  'exports' is defined but never used                                  no-unused-vars
  1953:2    error    Unnecessary semicolon                                                no-extra-semi
  2172:21   error    'groups' is already defined                                          no-redeclare
  2338:14   warning  'escapeCodeAsRegExpForMatchingInsideHTML' is defined but never used  no-unused-vars
  2395:13   warning  'reference' is assigned a value but never used                       no-unused-vars
  2407:25   error    Expected a conditional expression and instead saw an assignment      no-cond-assign
  2516:13   warning  'source' is defined but never used                                   no-unused-vars
```

Not sure if this path needs to be added to .eslintignore [since it seems to be copied from _/node_modules/_].

webextension/selector/ui.js
```diff
@@ -42,20 +42,13 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
...
-  }
-
   let substitutedCss = inlineSelectionCss.replace(/MOZ_EXTENSION([^\"]+)/g, (match, filename) => {
     return chrome.extension.getURL(filename);
   });
 
   function makeEl(tagName, className) {
     if (! iframe.document) {
-      console.trace();
+      throw new Error("Attempted makeEl before iframe was initialized");
```
6a68: I find this error a little confusing. If `makeEl` sometimes races with iframe initialization, is there a better way to enforce synchronization, so that this couldn't happen? Maybe a Promise that resolves when the iframe's document or window fires a ready event?

.eslintignore
```diff
@@ -3,3 +3,4 @@ static/vendor
 /scratch
 /Profile
 /build
+/webextension/build
```
6a68: might want to `.gitignore` the webextension build dir and manifest, too

webextension/background/analytics.js
```diff
@@ -1,26 +1,33 @@
...
+        if (req.status >= 300) {
+          let exc = new Error("Bad response from POST /event");
+          exc.status = req.status;
+          exc.statusText = req.statusText;
+          reject(exc);
+        } else {
+          resolve();
+        }
+      };
+      // FIXME: add cdX and other details from req.js
```
6a68: Could you just add the other details as part of this PR? Or if not, file a bug and include the bug number in the FIXME comment

webextension/background/auth.js
```diff
@@ -9,7 +9,13 @@ window.auth = (function () {
   let authHeader = null;
   let sentryPublicDSN = null;
 
-  chrome.storage.local.get(["registrationInfo"], (result) => {
+  chrome.storage.local.get(["registrationInfo"], catcher.watchFunction((result) => {
+    if (chrome.runtime.lastError) {
```
6a68: Would you mind changing this to use a Promise instead, maybe wrap `storage.local.get` in a `catcher.watchPromise` call?

According to the [`runtime.lastError` docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/lastError), you don't need to use `lastError` if you use the Promise-based versions of APIs.

It seems like pageshot uses promises pretty extensively, so it'd be nice to make these spots consistent while we're in here.

webextension/background/senderror.js
```diff
@@ -0,0 +1,102 @@
...
+    rest.stack = e.multilineStack || e.stack;
+    Raven.captureException(exception, {
+      logger: 'addon',
+      tags: {version: e.version, category: e.popupMessage},
+      message: exception.message,
+      extra: rest
+    });
+  };
+
+  catcher.registerHandler((errorObj) => {
```
6a68: Where is `errorObj` defined?

webextension/catcher.js
```diff
@@ -45,6 +61,16 @@ window.catcher = (function () {
     });
   };
 
+  exports.registerHandler = function (h) {
+    handler = h;
+    if (queue.length) {
```
6a68: I don't think you need to check the length when doing for...of

webextension/background/deviceInfo.js
```diff
@@ -1,10 +1,13 @@
-/* globals chrome */
+/* globals chrome, catcher */
 
 window.deviceInfo = (function () {
   let manifest = chrome.runtime.getManifest();
 
   let platformInfo = {};
   chrome.runtime.getPlatformInfo(function(info) {
+    if (chrome.runtime.lastError) {
```
ianb: getPlatformInfo doesn't actually produce a promise.  It seems like some more frequently used APIs have had promises added, but many like this one haven't been updated.

webextension/background/loadSelector.js
```diff
@@ -26,6 +26,7 @@ window.loadSelector = (function () {
         }, () => {
           if (chrome.runtime.lastError
               && chrome.runtime.lastError.message != "Script returned non-structured-clonable data") {
+              catcher.unhandled(chrome.runtime.lastError, {script: script});
```
ianb: This particular load pattern was the product of some experimentation.  executeScript doesn't always give an error object that matches chrome.runtime.lastError.

webextension/catcher.js
```diff
@@ -1,22 +1,38 @@
...
-
 window.catcher = (function () {
   let exports = {};
 
-  /** Call with an error object (with .name, .message, .help, etc) */
-  exports.unhandled = function unhandled(error, info) {
-    callBackground("reportError", makeError(error, info));
+  let handler;
+
+  let queue = [];
```
ianb: Ideally nothing earlier in the load process than the registered handler will emit an error – but if it does, we won't get Sentry notification without this queue, and we might miss errors.  The queue ensures that very early errors won't get dropped.

webextension/selector/callBackground.js
```diff
@@ -8,6 +8,9 @@ window.callBackground = function (funcName, ...args) {
       throw new Error(result.name);
     } else {
       console.error("Unexpected background result:", result);
+      let exc = new Error("Bad response from background page");
+      exc.resultType = result.type || "undefined";
```
ianb: It gets serialized as JSON, and the undefined value is dropped from JSON representations.

webextension/selector/ui.js
ianb: It shouldn't happen, and I don't believe there are currently races, but it's an error when it does happen.  I believe we wait for initialization somewhere in uicontrol.

.eslintignore
ianb: I believe those are both in there

webextension/background/senderror.js
ianb: It's a function parameter

webextension/background/auth.js
ianb: It appears chrome.storage.local also does not use promises :(

ianb:
@6a68 : back at ya!

webextension/background/deviceInfo.js
6a68: > getPlatformInfo doesn't actually produce a promise

Are you sure? Looks like it does in release channel FF:

screen shot 2017-03-02 at 7 18 28 pm

At any rate, this change is a nice to have, so it's fine to leave as is.

webextension/background/loadSelector.js
6a68: interesting :+1:

webextension/selector/callBackground.js
6a68: :+1:

webextension/background/auth.js
6a68: Hmm. Maybe this is because you're using `chrome` instead of `browser`?

screen shot 2017-03-02 at 7 31 29 pm

webextension/background/deviceInfo.js
6a68: Aha. I think the issue is actually that you are using `chrome`, not `browser`.

webextension/background/senderror.js
6a68: doh

.eslintignore
6a68: :+1:

PR #2265, opened 2017-03-02
Fix/2262

### LINKED ISSUE #2262: Remove SSL configuration & instructions
It turns out that running using https on localhost is no longer necessary (there used to be a different architecture where it was required, but that was changed).

To fix this we need to remove the setting from `.env.dev`, change any other references to `https://localhost:10080` to `http://localhost:10080` and change the README to remove references to SSL

niharikak101:
fixes #2262

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -5,4 +5,4 @@ LOG_LINT=true
 LOG_LEVEL=debug
 # To test caching locally:
 #SET_CACHE=true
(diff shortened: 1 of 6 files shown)


ianb:
I would leave some comments, but it all looks good so I have no comments.  Thanks!

niharikak101:
@ianb Yay! Thank you!

PR #2273, opened 2017-03-03
r?someone Add lifecycle

ianb:
This is some general hygiene for the content worker and its modules


webextension/background/loadSelector.js
```diff
@@ -2,6 +2,8 @@
 
 window.loadSelector = (function () {
   const scripts = [
+    "background/loadSelector.js",
+    "selector/lifecycle.js",
```
dannycoates: I notice that the order of this list matters. I think there ought to be a comment about that so someone doesn't reorder accidentally.

webextension/selector/lifecycle.js
```diff
@@ -0,0 +1,51 @@
...
+    let watchFunction = catcher.watchFunction;
+    let moduleNames = loadSelector.moduleNames;
+    // Make sure we unload catcher last, and don't call lifecycle.unload():
+    moduleNames = moduleNames.filter((n) => {
+      return n != "catcher" && n != "lifecycle";
+    });
+    moduleNames.reverse();
+    moduleNames.push("catcher");
+    for (let moduleName of moduleNames) {
+      if (moduleName == "catcher" || moduleName == "lifecycle") {
```
dannycoates: These get filtered above, yeah?

webextension/background/loadSelector.js
```diff
@@ -2,6 +2,8 @@
 
 window.loadSelector = (function () {
   const scripts = [
+    "background/loadSelector.js",
```
dannycoates: I thought it odd at first that we'd load loadSelector on the page to get the module names but I can't think of a better way to do it

webextension/selector/lifecycle.js
```diff
@@ -0,0 +1,51 @@
+/* globals loadSelector, catcher */
+
+window.lifecycle = (function () {
+  let exports = {};
+  if (window.lifecycle) {
+    window.lifecycle.unload();
+  }
+  document.dispatchEvent(new Event("pageshot-unload"));
+
+  exports.unload = function () {
```
dannycoates: More a nit than problem... This unload function doesn't follow the pattern of cleaning up itself, it cleans up everything, so it might be better to name it something else. Then we wouldn't have to special case it below.

ianb:
Addressed comments, also removed some defensive coding in the module (using events) that was unnecessary.

webextension/selector/lifecycle.js
```diff
@@ -0,0 +1,40 @@
...
+      }
+    }
+    if (window.lifecycle === exports) {
+      delete window.lifecycle;
+    }
+  };
+
+  return exports;
+})();
+null;
```
pdehaan: What's `null;` here?

PR #2282, opened 2017-03-04
moved clipboard copy to content-script context

dannycoates:
From [mdn](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard):

> you can write to the clipboard like this in all execution contexts except background pages. In Firefox you can't select text or focus an input field in background pages, so you can't write to the clipboard from a background page.

So I moved `clipboard.js` from background and wired up the promise chain to send a response to the content page.


dannycoates:
> This still doesn't work for me

Hmm, it's working for me every time now. I'm using `bin/run-addon` so fresh profile / nightly.

> Also clipboard.js should be removed from manifest.json.template

fixed

webextension/selector/shooter.js
```diff
@@ -78,7 +78,9 @@ window.shooter = (function () { // eslint-disable-line no-unused-vars
       selectedPos,
       shotId: shot.id,
       shot: shot.asJson()
-    }));
+    })).then((url) => {
```
dannycoates: debugging: @ianb if you put a log line here, do you see it hit?

webextension/background/takeshot.js
```diff
@@ -28,18 +28,18 @@ window.takeshot = (function () {
...
+    return catcher.watchPromise(capturePromise.then(() => {
       return uploadShot(shot);
     }).then(() => {
       let id = makeUuid();
-      clipboard.copy(shot.viewUrl);
       chrome.notifications.create(id, {
         type: "basic",
         iconUrl: "img/clipboard-32.png",
         title: "Link Copied",
         message: "The link to your shot has been copied to the clipboard"
```
fzzzy: I feel like this sentence should have a period after it, but it's probably not a big deal.

fzzzy:
I tested this, and the link was properly copied to my clipboard.

webextension/selector/shooter.js
ianb: Yes, it does.

Also, looking more closely at the code I realize there's a problem here – `catcher.watchPromise()` should wrap the entire promise, including the new .then expression.

ianb:
Approved but with pr/2282 branch applied.  On a fresh profile it worked for me.  I'll make a followup about catching the error.

ianb:
As a general note, watchPromise() should wrap the outermost promise anyplace where there isn't an explicit .catch().  Then if something goes wrong it'll be reported.  But you don't use it if the promise is returned, as it's presumably the responsibility of the caller to itself return or watch the promise.

PR #2289, opened 2017-03-06
removed 'copied' notification if the copy failed

### LINKED ISSUE #2284: Catch failing execCommand("copy")
Sometimes (in `webextension/clipboard.js`) the `document.execCommand("copy")` command will fail.  When it does the command returns false.  We should notice that, and change the notification popup accordingly.

I don't know why it sometimes doesn't work, but on one profile of mine it didn't work while on another fresh profile on my computer it worked.  Emitting a Sentry message (`catcher.unhandled`) might also be appropriate.

dannycoates:
fixes #2284 

I moved the notification and tab to be after the copy and report an error if the copy failed.

### DIFF
--- a/webextension/background/main.js
+++ b/webextension/background/main.js
@@ -1,8 +1,9 @@
 /* globals chrome, console, XMLHttpRequest, Image, document, setTimeout, navigator */
-/* globals loadSelector, analytics, communication, catcher */
+/* globals loadSelector, analytics, communication, catcher, makeUuid */
 window.main = (function () {
   let exports = {};
 
+  const pasteSymbol = (window.navigator.platform.match(/Mac/i)) ? "\u2318" : "Ctrl";
   const { sendEvent } = analytics;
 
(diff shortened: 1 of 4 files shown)


webextension/background/main.js
```diff
@@ -62,5 +63,19 @@ window.main = (function () {
     chrome.tabs.create({url: backend + "/shots"});
   });
 
+  communication.register("openShot", ({url, copied}) => {
```
dannycoates: I'm not sure if main is the right place for this handler, but "openMyShots" is here so maybe its a decent spot.

dannycoates: Actually, after seeing the PR for the `/creating` page I should change this to not open the tab.

ianb: I suspect you've found some completely dead code, there was a point when the content worker was uploading the shot and then the background page opened the tab.

ianb: I opened #2297 to do a review of this

PR #2313, opened 2017-03-07
Wrap the webextension in a bootstrap addon

### LINKED ISSUE #2222: Design Embedded WebExtension
For the purposes of Telemetry (#2218) we're going to need to be an [Embedded WebExtension](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Embedded_WebExtensions#Starting_the_WebExtension).  The wrapper will be a bootstrap extension (not SDK).

I'd like to keep developing as a WebExtension, which means my development won't emit Telemetry, which is OK.  Then we'd build as a bootstrap add-on, and do testing and integration stuff with the bootstrap add-on.

This probably informs our migration path (#2169), as we'll have more options with the bootstrap extension.  The WebExtension should always be able to run without the wrapper, as this is how it will run on Chrome (assuming we eventually are usable on Chrome), and because it's easier to develop that way.

ianb:
Fixes #2222 

I don't know if this will be conflicty or not, since it's mostly file moves I think it'll be fine.

After landing I'll look through good first bug tickets that refer to the webextension/ directory to update them.


ianb:
#2314 as a followup on install.rdf FIXMEs

6a68:
@ianb yup, looks like this one could use a rebase, too

README.md
```diff
@@ -24,7 +24,7 @@ Clone the [repository](https://github.com/mozilla-services/pageshot/).
 There are two scripts to run the server locally and develop the add-on:
 
 - `./bin/run-server` will run the server on `http://localhost:10080` and automatically restart if there are changes.
-- `./bin/run-addon` will build a few parts of the addon (into `webextension/build/`) and start Firefox with the add-on installed.  The add-on will be refreshed automatically as you change files.  We recommend you open `about:debugging` to help debug the extension.
+- `./bin/run-addon` will build a few parts of the addon (into `addon/webextension/build/`) and start Firefox with the add-on installed.  The add-on will be refreshed automatically as you change files.  We recommend you open `about:debugging` to help debug the extension.
```
6a68: Unfortunately, [according to MDN](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Embedded_WebExtensions#Debugging) at least, you'll have to use the Browser Toolbox to debug.

ianb:
@6a68: rebased, probably the notable conflicts were in the previous PR, this one was simple enough.

README.md
ianb: about:debugging gives a button that opens a remote debugger, more or less Browser Toolbox I think.  It's what I've been using successfully

addon/install.rdf
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<RDF xmlns="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+     xmlns:em="http://www.mozilla.org/2004/em-rdf#">
+  <Description about="urn:mozilla:install-manifest">
+    <em:id>pageshot@mozilla.org</em:id>
+    <em:name>Page Shot</em:name>
+    <em:targetApplication>
+      <Description>
+        <em:id>{ec8030f7-c20a-464f-9b0e-13a3a9e97384}</em:id> <!--Firefox-->
+        <em:minVersion>51.0a1</em:minVersion>
```
pdehaan: Do we want `minVersion` of 51 here, or 54?

README.md
6a68: The MDN page I linked claims that you can't use about:debugging to debug an embedded webextension. maybe MDN is wrong, though :-)

ianb: I improved the docs here – `run-addon` still just runs webextension, but `run-addon --bootstrap` runs with the whole bootstrap wrapper.  You are right about about:debugging for that latter case.

addon/install.rdf
ianb: I haven't tested on 51, but that's when Embedded Extension support landed.  I have tested 53 and it works, we have a fallback for the one 54-only API that we use.  OTOH, it's probably safer for pre-54 people to use the WebExtension and not the bootstrap extension.

Note ESR is Firefox 52, so there's some value is supporting 52.

PR #2330, opened 2017-03-08
Clean up shot.js

### LINKED ISSUE #2156: Review shot.js
Probably there's lots of extra stuff.  For instance text clips.

Right now clips are plural on a shot, which is something we don't support in any way in the interface.  I'm not sure if we should fix that.

ianb:
Fixes #2156
Remove leftover references to body/head DOM attributes


6a68:
@fzzzy any chance you could review this one? I'm still a bit hesitant about reviewing wide-ranging changes to the page shot code

fzzzy:
@6a68 yep, sure

server/src/server.js
```diff
@@ -767,9 +726,9 @@ app.post("/api/set-title/:id/:domain", function (req, res) {
     simpleResponse(res, "Not logged in", 401);
     return;
   }
-  Shot.get(req.backend, shotId).then((shot) => {
-    if (shot.deviceId !== req.deviceId) {
-      simpleResponse(res, "Not the owner", 403);
+  Shot.get(req.backend, shotId, req.deviceId).then((shot) => {
```
fzzzy: So since req.deviceId is getting passed in to get now, if the user doesn't own the shot, they just will get a 404 instead of 403 Not the owner, right?

fzzzy:
Nice cleanup!

PR #2343, opened 2017-03-10
fix: Redirects new tab to my shots

### LINKED ISSUE #2238: Reimplement about:newtab -> My Shots
When you hit the Page Shot button on a new tab page you should be redirected to My Shots.

There might have to be changes to the add-on permission to run on about pages (if so, those changes go in `addon/webextension/manifest.json.template`).

The location of My Shots is in `addon/webextension/background/main.js` in the `main.getBackend()` function, `+ '/shots'`, and it should change the URL of the current (about:blank) tab.

niharikak101:
fixes #2238

### DIFF
--- a/addon/webextension/background/main.js
+++ b/addon/webextension/background/main.js
@@ -30,12 +30,12 @@ window.main = (function () {
 
   chrome.browserAction.onClicked.addListener(function(tab) {
     if(tab.url.match(/about:(newtab|blank)/i)) {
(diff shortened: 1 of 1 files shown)


ianb:
I added an event in here (which I wouldn't have expected you to know about) – when I made that change I also noticed the whitespace was a little funny.  I fixed it too, but something to keep an eye on.

Issue #2333, opened 2017-03-08
Support `extensions.%NAME%.userdisabled`


wresuolc:
%NAME% is simply whatever we decide to name this thing. Right now it's "pageshot".

#2332 is our high-level disable flag.  No one should be messing with that flag (random power-user aside) -- it's entirely there to enable us to roll this out slowly with Shield.

*This* issue is about implementing the code that allows the user to disable Page Shot.  Some details:

- I'm suggesting using the flag `extensions.pageshot.userdisabled` but pick a different one if you like
- The code should look for this flag pretty early on startup (could be right after the code in #2332)
- This code actually needs some UI in about:preferences.  @johngruen where should this go and what should it look like?

6a68:
I'm not sure about preferences UI for embedded webextensions. This also depends on whether we want users to see page shot in about:addons or not (I don't recall what the decision was there). 

@andymckay, if we want to allow users to disable page shot (embedded webextension system addon), should we implement that UI using the [webextension approach](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Implement_a_settings_page) or the bootstrapped addon approach?

andymckay:
Let's try to avoid writing more XUL and focus on the webExtension approach please.

Note that chrome_style for the settings page is landing in 55, sadly a release too late.

https://bugzilla.mozilla.org/show_bug.cgi?id=1275287

6a68:
@andymckay sounds good to me! t

ianb:
If we want to disable the WebExtension entirely when the pref is off (by not calling `.startup()`), then I don't think we can use the WebExtension approach, can we?  Also since the add-on won't appear in about:addons, there's no place for the preferences button.

6a68:
TL;DR: yes, you can shutdown and restart embedded webextensions, but I 

It turns out that embedded webextensions have an undocumented shutdown method. Here's how you shut down the webextension without shutting down the outer extension (we could do this in a pref listener):

```js
const { EmbeddedExtensionManager } = Cu.import("resource://gre/modules/LegacyExtensionsUtils.jsm");
const webext = EmbeddedExtensionManager.getEmbeddedExtensionFor({id: 'pageshot@mozilla.org'});
webext.shutdown();
```

[...]

Issue #2332, opened 2017-03-08
Support `extensions.%NAME%.disabled`


wresuolc:
%NAME% is simply whatever we decide to name this thing.  Right now it's "pageshot".

Over in #2304 I talk about reporting the status of `extensions.pageshot.disabled`.  This issue is to actually add and implement that preference.  This preference should be relatively straight-forward.  Check out my sweet pseudocode:

```
   # Somewhere in an init function:
   if extensions.pageshot.disabled == 1:
     return;
```

If `extensions.pageshot.disabled` is missing, we should enable as usual.  The point of this pref is to bail as early as possible if it's not enabled so we don't use up any resources loading libraries and what not.

6a68:
We have to manually start the webextension from inside the bootstrap.js startup function, so, if the pref is set, we will not start the webextension. I think that should do it.

ianb:
I'm going to name this extensions.pageshot.systemDisabled to make it clearer how it is different than #2333

Though the prefs will be controlled differently, they will have exactly the same effect.  But some people will probably go into about:config and turn them both off just to be Extra Careful.

wresuolc:
Sounds good.  For the record, once we're done rolling this out (a couple weeks?) we can just remove this pref and the code for it altogether.

6a68:
Grabbing a few of these prefs / initialization / disabling bugs

PR #2350, opened 2017-03-10
Add missing bootstrap.js lifecycle functions

### LINKED ISSUE #2340: add other lifecycle methods to bootstrap.js
I can't find a good reference for this claim, but I believe that we need to have all four lifecycle functions in bootstrap.js with the `data, reason` function signature:

```js
// somewhere in bootstrap.js
function shutdown(data, reason) {}
function install(data, reason) {}
function uninstall(data, reason) {}
```

6a68:
Fixes #2340.

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -7,8 +7,8 @@ function startup(data, reason) {
   });
 }
 
-function install() {
-}
+function shutdown(data, reason) {}
+function install(data, reason) {}
+function uninstall(data, reason) {}
 
-function handleMessage() {
-}
+function handleMessage() {}


addon/bootstrap.js
```diff
@@ -7,8 +7,8 @@ function startup(data, reason) {
   });
 }
 
-function install() {
-}
+function shutdown(data, reason) {}
```
pdehaan: Q: Do we need to remove the listener that we added in `startup()` above during one of these lifecycle functions?

6a68: Good question. That's not covered in the [MDN docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Embedded_WebExtensions#Starting_the_WebExtension) for embedded webextensions, and to do so, we'd have to close over the `browser` object. I'm not sure if that's a good idea or not.

There are some bugs filed (#2332, #2333) around handling prefs to disable/enable the webextension, so maybe we can figure out the runtime.onMessage listener question there?

Issue #2335, opened 2017-03-09
"Page Shot went haywire" error is displayed multiple times if the Page Shot button is clicked on "about:*" pages (On Mac OS)


SoftVision-CosminMuntean:
[Notes]:
- **On Windows the errors are not displayed as system notification, they only appear in the browser console.**
- **On Windows if the Page Shot button is clicked multiple times, the browser becomes unresponsive.** I have noticed that the browser remains this way until all the errors are loaded in browser console. Here is a screen recording with the issue: https://goo.gl/xITz4M.

[Affected versions]:
- Nightly 55.0a1 (Build ID: 20170308030207)	
- Page Shot 5.2.14901

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" is set to "false" and latest Page Shot (5.2.14901) custom build is installed on a new clean profile.

[Steps to reproduce]:
1. Open Firefox with the profile from prerequisites.
2. Go to any Firefox "about:*" page (eg: "about:home" page).
3. Click the Page Shot button from the browser toolbar.
4. Observe the browser behavior.

[Expected result]:
- A single "Page Shot went haywire" error is displayed.

[Actual result]:
- 15 "Page Shot went haywire" errors are displayed.

[Additional Notes:]
- This issue depends on the decision that will be reached in issue #2236.
- Here is a screen recording with the issue: https://goo.gl/YFNy7a

dannycoates:
fixed in #2368

PR #2351, opened 2017-03-10
Chrome to browser

ianb:


addon/webextension/background/auth.js
```diff
@@ -10,30 +10,18 @@ window.auth = (function () {
...
     if (result.abTests) {
       abTests = result.abTests;
     }
     if (result.registrationInfo) {
       registrationInfo = result.registrationInfo;
     } else {
       registrationInfo = generateRegistrationInfo();
-      chrome.storage.local.set({
+      console.info("Generating new device authentication ID", registrationInfo);
+      return browser.storage.local.set({
```
6a68: Did you mean to return a Promise here?

6a68:
looking now

addon/webextension/background/loadSelector.js
```diff
@@ -25,19 +25,19 @@ window.loadSelector = (function () {
...
           file: script
         }, () => {
-          if (chrome.runtime.lastError
-              && chrome.runtime.lastError.message != "Script returned non-structured-clonable data") {
-              catcher.unhandled(chrome.runtime.lastError, {script: script});
-              console.error("Error loading script", script, ":", chrome.runtime.lastError);
+          if (browser.runtime.lastError
+              && browser.runtime.lastError.message != "Script returned non-structured-clonable data") {
+              catcher.unhandled(browser.runtime.lastError, {script: script});
+              console.error("Error loading script", script, ":", browser.runtime.lastError);
```
6a68: Is there some reason you're sticking with the callback error check here?

Seems like you could do 

```js
return browser.tabs.executeScript({ file: script }).catch(err => {
  if (err.message != "Script returned non-structured-clonable data") {
    catcher.unhandled(err, {script: script});
    console.error(`Error loading script: ${err}`);
  }
});
```

The code comments say that order matters, otherwise I'd suggest using a `Promise.all` here, which would nicely auto-fail if any of the scripts doesn't load.

6a68: Hmm, actually it looks like [Promise.all should preserve array order](http://stackoverflow.com/questions/28066429/promise-all-order-of-resolved-values), so that may be a cleaner approach here.

addon/webextension/background/main.js
```diff
@@ -25,30 +25,28 @@ window.main = (function () {
...
 
-  chrome.runtime.onInstalled.addListener(function () {
-  });
-
-  chrome.browserAction.onClicked.addListener(function(tab) {
+  browser.browserAction.onClicked.addListener(function(tab) {
     if(tab.url.match(/about:(newtab|blank)/i)) {
       sendEvent("goto-myshots", "about-newtab");
-      chrome.tabs.update({url: backend + "/shots"});
+      browser.tabs.update({url: backend + "/shots"});
```
6a68: `browser.tabs.update` returns a promise, please wrap in a `watchPromise` call

addon/webextension/background/main.js
```diff
@@ -65,13 +63,13 @@ window.main = (function () {
   });
 
   communication.register("openMyShots", () => {
-    chrome.tabs.create({url: backend + "/shots"});
+    browser.tabs.create({url: backend + "/shots"});
```
6a68: `browser.tabs.update` returns a promise, please wrap in a `watchPromise` call

addon/webextension/background/main.js
```diff
@@ -65,13 +63,13 @@ window.main = (function () {
...
   });
 
   communication.register("openShot", ({url, copied}) => {
     if (copied) {
       const id = makeUuid();
-      chrome.notifications.create(id, {
+      browser.notifications.create(id, {
```
6a68: This also returns a promise, please wrap in a `watchPromise` call.

The [MDN Page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/notifications/create) for `browser.notifications.create` contains a rather grim warning: "If you call notifications.create() more than once in rapid succession, Firefox may end up not displaying any notification at all."

I don't know how quickly openShot might return, but would it be a good idea to debounce this function?

addon/webextension/background/senderror.js
```diff
@@ -57,7 +57,7 @@ window.errorpopup = (function () {
         message = error.message;
       }
     }
-    chrome.notifications.create(id, {
+    browser.notifications.create(id, {
```
6a68: As mentioned in the other occurrence of `browser.notifications.create`, repeatedly triggering this function may lead to weird results. This might be the root cause of #2335. I'd say debouncing would be a good idea here.

Also need to wrap this in a `watchPromise`.

addon/webextension/site-helper.js
```diff
@@ -7,7 +7,7 @@ function sendEvent(type, detail) {
 }
 
 document.addEventListener("has-saved-shot", (event) => {
-  chrome.runtime.sendMessage({
+  browser.runtime.sendMessage({
```
6a68: Any reason you skipped the Promise conversion for these `browser.runtime.sendMessage` calls?

addon/webextension/background/auth.js
ianb: Yes, the promise goes back to the parent watchPromise this way.

addon/webextension/background/loadSelector.js
ianb: This particular error shows up in the console, but doesn't show up as a rejected promise.  It's a nuisance error.

AFAIU Promise.all doesn't ensure that the promises complete in order.  That is, the first module should be completely loaded before attempting the load the second module.

addon/webextension/background/senderror.js
ianb: I created #2353 for the limit, otherwise out of scope for this change

ianb: oh, and we shouldn't watch anything in senderror because we'll get loops!

addon/webextension/site-helper.js
ianb: site-helper.js needs a more general refactoring, so I've only done minimal changes here

ianb:
Change requests applied, except as explained

6a68:
LGTM! merging

PR #2344, opened 2017-03-10
Extract strings using WebExtension i18n library

6a68:
Fixes part of #2294. Doesn't yet manage the transformation from a pontoon .properties file to this messages.json format.

I haven't been super consistent about the l10n keys. If anyone has strong opinions here, I am open to suggestions (cc @flodolo).

PR #2358, opened 2017-03-11
2294 pontoon to webextension

### LINKED ISSUE #2294: implement l10n in addon
refs #2205

6a68:
This is built on top of the string extraction commit from #2344, which hasn't landed yet.

Commit message pretty much says it all:

-----

Add pontoon strings & build step to transform into webextension strings

- Add pontoon-to-webext.js script from bwinton/SnoozeTabs repo and
  related npm dependencies (to be removed when/if that script is
  published on npm as a standalone module).

- Add extracted strings to a properties file at the location expected by
  Pontoon.

- Add a pontoon-to-webextension build step to the Makefile

- Also, replace the hand made messages.json with the more nicely-formatted
  script output.

This commit, together with the fix for #2344, closes #2294.


Makefile
```diff
@@ -35,6 +35,9 @@ imgs_server_dest := $(imgs_source:%=build/server/%)
 
 raven_source := $(shell node -e 'console.log(require.resolve("raven-js/dist/raven.js"))')
 
+l10n_source := $(wildcard locales/*/*.properties)
+l10n_dest := $(l10n_source:%/*.properties=addon/webextension/_locales/%/messages.json)
```
6a68: not sure about this, but `make` doesn't seem to complain, so I guess it's alright?

6a68:
@flodolo Hi, here's the other localization step, with the strings translated into Pontoon-ready .properties format. Comments again appreciated. The pontoon-to-webext script from snooze tabs worked perfectly, thanks for the suggestion :+1: (and huge thanks to @lmorchard for writing such a great tool).

Makefile
ianb: This doesn't look right to me, but I'm not sure.  What should the value of $(l10n_source) be?  If it's like `locales/en-US/something.properties locales/es-ES/something.properties` and you want to get `addon/webextensions/_locales/en-US/messages.json addon/webextensions/_locales/en-US/messages.json` then it's right.  This presumes that each directory has only one properties file and that it's name isn't important.

6a68: I was thinking we'd eventually have multiple properties files in this directory, hence the double wildcard. However, now that I think about it, we would only ever want to package up the addon's strings in the addon...so I'll update this to grab `$(wildcard locales/*/webextension.properties)`

Makefile
```diff
@@ -35,6 +35,9 @@ imgs_server_dest := $(imgs_source:%=build/server/%)
 
 raven_source := $(shell node -e 'console.log(require.resolve("raven-js/dist/raven.js"))')
 
+l10n_source := $(wildcard locales/*)
+l10n_dest := $(l10n_source:%/webextension.properties=addon/webextension/_locales/%/messages.json)
```
6a68: @ianb meh, decided to just solve today's problem, and worry about building server strings when we get there

locales/en-US/webextension.properties
```diff
@@ -0,0 +1,27 @@
...
+contextMenuLabel = Create Page Shot
+myShotsLink = My Shots
+screenshotInstructions = Drag or click on the page to select a region. Press ESC to cancel.
+saveScreenshotSelectedArea = Save
+saveScreenshotVisibleArea = Save visible
+saveScreenshotFullPage = Save full page
+cancelScreenshot = Cancel
+downloadScreenshot = Download
+notificationLinkCopiedTitle = Link Copied
+# LOCALIZATION NOTE(notificationLinkCopiedDetails): The string "{meta_key}-V" should be translated
```
flodolo: You can avoid the LOCALIZATION NOTE part, since tools will associate and display the comment to the following string.

Possible comment:
```
# The string "{meta_key}-V" should be translated to the region-specific
# shorthand for the Paste keyboard shortcut. {meta_key} is a placeholder for the
# on Windows systems.
```

6a68: :+1: Thanks for the feedback! I'll update that comment.

6a68:
@ianb Is this good to land?

Makefile
```diff
@@ -103,6 +106,9 @@ signed_xpi: addon
 	./node_modules/.bin/web-ext sign --api-key=${AMO_USER} --api-secret=${AMO_SECRET} --source-dir addon/webextension/
 	mv web-ext-artifacts/*.xpi build/pageshot.xpi
 
+addon_locales: $(l10n_dest)
```
ianb: I'm not confident exactly how this will work.  Generally the Makefile pattern would be to set the requirement and then have other rules for how to generate `addon/webextension/_locales/%/messages.json`, like:

```
addon/webextension/_locales/%/messages.json: locales/%/webextension.properties
  script-that-translates-one-file $< $@
```

Since this appears to generate all the files in one go, it doesn't quite fit that pattern.  I can't decide if $(l10n_source) belongs here, or maybe not, or maybe nothing at all?

Also there should be `.PHONY: addon_locales` above this

Makefile
```diff
@@ -103,6 +106,9 @@ signed_xpi: addon
...
+	./bin/pontoon-to-webext.js --dest addon/webextension/_locales
```
ianb: pontoon-to-webext should go in `bin/build-scripts/` since it isn't called directly, instead it's only called through the Makefile

package.json
```diff
@@ -16,8 +16,10 @@
     "envc": "2.5.0",
     "escape-html": "1.0.3",
     "express": "4.15.2",
+    "habitat": "^3.1.2",
```
ianb: We've generally been using exact versions, then upgrading in batch, so that we know everyone has the exact same version.

addon/webextension/_locales/en_US/messages.json
```diff
@@ -1,33 +1,79 @@
 {
```
ianb: Should this file be in the source at all?  Isn't it generated from the properties file?

bin/pontoon-to-webext.js
```diff
@@ -0,0 +1,165 @@
...
+        const localeList = [];
+
+        // Get rid of the top level, we're only interested with what's inside it
+        dirTree.splice(0,1);
+        dirTree.forEach((localeLocation) => {
+          // Get the locale code from the end of the path. We're expecting the structure of Pontoon's output here
+          const langcode = localeLocation.split(path.sep).slice(-1)[0];
+
+          if (langcode) {
+            localeList.push(langcode);
```
ianb: Why wouldn't this be present?  Seems like it should be a warning.

bin/pontoon-to-webext.js
```diff
@@ -0,0 +1,165 @@
...
+  return new Promise((resolve, reject) => {
+    if (supportedLocales === '*') {
+      FS.listDirectoryTree(path.join(process.cwd(), config.src)).then((dirTree) => {
+        const localeList = [];
+
+        // Get rid of the top level, we're only interested with what's inside it
+        dirTree.splice(0,1);
+        dirTree.forEach((localeLocation) => {
+          // Get the locale code from the end of the path. We're expecting the structure of Pontoon's output here
+          const langcode = localeLocation.split(path.sep).slice(-1)[0];
```
ianb: Isn't `localeLocation.split(path.sep).slice(-1)[0]` equivalent to `localeLocation.split(path.sep)[1]` ?  That is, split with a string argument always splits 0 or 1 times.

(I find that split behavior confusing, so I'm always suspicious when I see it being used.)

bin/pontoon-to-webext.js
```diff
@@ -0,0 +1,165 @@
...
+          if (langcode) {
+            localeList.push(langcode);
+          }
+        });
+        return resolve(localeList);
+      }).catch((e) => {
+        reject(e);
+      });
+    } else {
+      supportedLocales = supportedLocales.split(',').map(item => item.trim());
```
ianb: Probably should be `supportedLocales.split(/,/g)`

bin/pontoon-to-webext.js
```diff
@@ -0,0 +1,165 @@
...
+  for (const entry of entries) {
+    const publicPath = path.join(process.cwd(), config.dest, entry.locale.replace('-', '_'));
+    const localesPath = path.join(publicPath, 'messages.json');
+
+    FS.makeTree(publicPath).then(() => {
+      return FS.write(localesPath, JSON.stringify(entry.content, null, 2));
+    }).then(() => {
+      log(`Done compiling locales at: ${localesPath}`);
+    }).catch((e) => {
+      error(e);
```
ianb: These errors won't lead to a build error, but they should.  It's okay if build scripts are a little finicky as long as they fail hard.

I'd recommend:

```js
function fatal(...args) {
  console.error(...args);
  process.exit(1);
}
```

Also this is all kind of brutal in Node.js compared to Python, but eh.

Makefile
ianb: Upon further reflection, I guess an empty dependencies list is the right thing here, given that we don't have recipes for how to generate those individual files.  Either way I believe the .json files get rebuilt on every make run because we're not telling make enough for it to know when a rebuild is necessary.

6a68: :+1: The rebuild is pretty cheap, for now at least (one locale...)

addon/webextension/_locales/en_US/messages.json
6a68: :+1: removed

bin/pontoon-to-webext.js
6a68: (To be clear, this is upstream code that I'm hoping Les will turn into an npm package. We're using copypasta because it's expedient.)

So, this line is getting the last directory's name off an absolute path; the `slice(-1)[0]` is effectively the same as doing a `pop()` off the end of the array.

```js
let localeLocation = '/Users/jhirsch/codez/github/mozilla-services-pageshot/locales/en-US'
localeLocation.split('/')
// Array [ "", "Users", "jhirsch", "codez", "github", "mozilla-services-pageshot", "locales", "en-US" ]
localeLocation.split('/').slice(-1)
// Array [ "en-US" ]
localeLocation.split('/').slice(-1)[0]
// "en-US"
```

bin/pontoon-to-webext.js
6a68: Hmm. I agree that it's hard to see how `langcode` could fail to exist.

bin/pontoon-to-webext.js
6a68: I think we're good here, since `split` splits globally by default:

```js
'a,b,c'.split(',') // returns [ "a", "b", " c" ]

bin/pontoon-to-webext.js
6a68: :+1: changed (I went ahead and [upstreamed](https://github.com/bwinton/SnoozeTabs/pull/260) that suggestion :-)

package.json
6a68: :+1:

addon/webextension/_locales/en_US/messages.json
6a68: er, removed and gitignore'd

6a68:
@ianb Updated. Mind taking another look?

PR #2359, opened 2017-03-11
Fix #2250, make sendEvent conditional on whether the user has opted-out of Telemetry generally

### LINKED ISSUE #2250: Read Telemetry opt-in/opt-out preference
When someone turns off Telemetry data collection, Page Shot should notice that and stop collecting information itself (i.e., sendEvent should disable).

This will have to happen from the bootstrap addon portion, not the WebExtension.

ianb:


addon/webextension/background/main.js
```diff
@@ -30,10 +30,14 @@ window.main = (function () {
 
   chrome.browserAction.onClicked.addListener(function(tab) {
     if(tab.url.match(/about:(newtab|blank)/i)) {
-      sendEvent("goto-myshots", "about-newtab");
+      catcher.watchPromise(analytics.refreshTelemetryPref().then(() => {
```
6a68: Rather than constantly re-check the prefs, you should just have the bootstrap code check once, at startup, then use a [preferences observer](https://developer.mozilla.org/en-US/Add-ons/Code_snippets/Preferences#check_For_Existence) the rest of the time.

The bootstrap code will have to send over update messages when the pref changes, a bit different than the approach you've taken here, where the webextension initiates the exchange. You can probably default telemetry to disabled, dropping packets if the initial pref message hasn't arrived before the user starts clicking.

docs/METRICS.md
```diff
@@ -2,6 +2,10 @@
 
 A summary of the metrics Page Shot will record, and what we're looking for in those metrics.
 
+### Opt-out
+
+The add-on reads the Telemetry opt-out preference (`toolkit.telemetry.enabled`) each time the user presses the Page Shot button.  If this preference is false, or if there is any issue trying to fetch the preference, then no data is sent to the Page Shot server.
```
6a68: You can update this to say that the add-on tracks the value of the Telemetry preference.

addon/bootstrap.js
```diff
@@ -11,4 +15,9 @@ function shutdown(data, reason) {}
 function install(data, reason) {}
 function uninstall(data, reason) {}
 
-function handleMessage() {}
+function handleMessage(msg, sender, sendReply) {
+  if (msg && msg.funcName === "getTelemetryPref") {
+    let enableTelemetry = prefs.getBoolPref("toolkit.telemetry.enabled");
```
6a68: You need to `getPrefType` before you `getBoolPref`, otherwise `getBoolPref` can throw if it's unset for any reason. See [MDN](https://developer.mozilla.org/en-US/Add-ons/Code_snippets/Preferences#check_For_Existence) for more details

addon/webextension/background/main.js
ianb: I specifically considered a preference observer, and it felt much more complicated to me without much benefit.  The bootstrap code has to initiate the message and it causes persistent code in the addon to watch things (where in this version nothing is observed when the add-on is passive).  This seems reasonably respectful, since it only gets it wrong for a small amount of time if you happen to change the pref in the middle of a Page Shot session, and the impact is very limited.

6a68: Hmm. I thought preference observers were needed to minimize the performance impact of repeatedly checking prefs, but I don't see that in the docs. The only [documented](https://developer.mozilla.org/en-US/Add-ons/Code_snippets/Preferences#Using_preference_observers) reason to use a preference observer is to detect changes in prefs that normally require a restart to be toggled.

Hi @Mossop, could you weigh in on the performance cost associated with Page Shot using `browser.runtime` messaging to check a pref each time the Page Shot toolbar button is pressed? I thought that repeatedly accessing prefs caused a significant performance hit, but that information might have been wrong or out of date, as I don't see it documented anywhere on MDN.

addon/bootstrap.js
Standard8: If you're willing to limit the add-on to 54 and above, then [bug 1338306](https://bugzilla.mozilla.org/show_bug.cgi?id=1338306) just landed, which lets you supply an optional argument which will avoid the throw.

addon/webextension/background/main.js
Mossop: Pref lookups are cheap, it's just a hash table in effect. I don't know the cost of browser.runtime messaging though, @kmaglione would know that better than me.

ianb: Some browser.runtime message is inevitable, since we need to get the pref value to the webextension somehow.

ianb:
Merging as I believe all comments are addressed

PR #2362, opened 2017-03-13
Fixes #2318: Provides export_addon makefile rule to export pageshot into a mozilla-central based repository.

### LINKED ISSUE #2318: Create script to copy Page Shot into Firefox
We should have a script that will import the addon (everything in `addon/` after running `make addon`) into the Firefox tree.

Standard8:
The environment variable EXPORT_MC_LOCATION provides for changing where the add-on is exported to.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 PATH := ./node_modules/.bin:./bin/:$(PATH)
 SHELL := /bin/bash
 BABEL := babel --retain-lines
+RSYNC := rsync --archive
 .DEFAULT_GOAL := help
 # Sets $(PAGESHOT_BACKEND) to http://localhost:10080 only if it isn't set
 PAGESHOT_BACKEND ?= http://localhost:10080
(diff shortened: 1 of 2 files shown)


Standard8:
[Bug 1346825](https://bugzilla.mozilla.org/show_bug.cgi?id=1346825) has the additional patch to apply directly to mozilla-central for getting pageshot as part of the build.

Standard8:
> But generally I think this sort of thing works best as a shell script, not a Makefile target. It's easier to add more steps and tests, and call out to make as necessary to ensure everything is refreshed. E.g., I might add if [[ ~ -d "$EXPORT_MC_LOCATION/browser/" ]] ; then echo "Firefox directory $EXPORT_MC_LOCATION does not look like a checkout" ; exit 1 ; fi

I'll work on that in the next patch - I'll be adding more wrapping around things anyway, so it might be better suited to moving it to a script.

PR #2366, opened 2017-03-13
removed rich copy button from share panel

### LINKED ISSUE #2322: Disable rich copy button
The rich copy button in the share panel (share-panel.js) doesn't work and we should remove it.

dannycoates:
fixes #2322

I wasn't sure if we want to strip out all the rich copy code or just remove the button so I started with the latter.


ianb:
@dannycoates : I think we should strip out all the code, I don't think it will be easy to refresh for a WebExtension so it's unlikely to go back

dannycoates:
> I think we should strip out all the code

updated

PR #2368, opened 2017-03-13
disable the pageshot button on about,data,moz-extension pages

### LINKED ISSUE #2236: Verify add-on behavior for special pages
We should test `about:` pages, `data:` pages, `moz-extension:` pages, and pages that have a `<frameset>`

Probably most of these pages should just emit an error, though `data:` or `file:` pages should be okay.

dannycoates:
It seems like disabling the pageshot button on these pages is better than doing anything else. It needs the `tabs` permission (I don't know if that's ok or not) since `activeTab` doesn't kick in until a user action is performed.

I don't know what to do about framesets yet.

closes #2236

### DIFF
--- a/addon/webextension/background/main.js
+++ b/addon/webextension/background/main.js
@@ -40,8 +40,12 @@ window.main = (function () {
       });
   }
 
+  function opensMyShots(url) {
+    return /^about:(?:newtab|blank)/i.test(url) || /^resource:\/\/activity-streams\//i.test(url);
+  }
+
   browser.browserAction.onClicked.addListener(catcher.watchFunction((tab) => {
-    if (tab.url.match(/about:(newtab|blank)/i)) {
+    if (opensMyShots(tab.url)) {
       catcher.watchPromise(analytics.refreshTelemetryPref().then(() => {
         sendEvent("goto-myshots", "about-newtab");
       }));
@@ -77,6 +81,26 @@ window.main = (function () {
         .then(() => sendEvent("start-shot", "context-menu")));
   }));
 
+  function urlEnabled(url) {
+    if (opensMyShots(url)) {
+      return true;
+    }
+    if (url.startsWith(backend) || /^(?:about|data|moz-extension):/i.test(url)) {
+      return false;
+    }
+    return true;
+  }
+
+  browser.tabs.onUpdated.addListener((id, info, tab) => {
+    if (info.url && tab.selected) {
+      if (urlEnabled(info.url)) {
+        browser.browserAction.enable(tab.id);
+      }
+      else {
+        browser.browserAction.disable(tab.id);
+      }
+    }
+  });
 
   communication.register("sendEvent", (sender, ...args) => {
     catcher.watchPromise(sendEvent(...args));
--- a/addon/webextension/background/main.js
+++ b/addon/webextension/background/main.js
@@ -40,12 +40,12 @@ window.main = (function () {
       });
   }
 
-  function opensMyShots(url) {
+  function shouldOpenMyShots(url) {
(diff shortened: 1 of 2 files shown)


ianb:
In response to the description, not the PR: I don't think we can proactively disable the button for framesets, but that it's acceptable once someone tries to use Page Shot to give them an error.  In that case, the worker would start up, I think notice that the page has no `document.body`, signal an error and tear itself down.

addon/webextension/background/main.js
```diff
@@ -55,6 +55,17 @@ window.main = (function () {
     catcher.watchPromise(loadSelector());
   }));
 
+  browser.tabs.onUpdated.addListener((id, info, tab) => {
+    if (info.url && tab.selected) {
+      if (/^(?:about|data|moz-extension):/.test(info.url)) {
```
ianb: We do have behavior for about:new and about:blank (opening the shots page).  So we need a match to treat those separately.

ianb: Ideally that test would be in some new function here and in the onClick handler.

dannycoates:
I also need to fix switching tabs

ianb:
The bummer is this is where we actually could affect browser performance generally, even if a user doesn't use Page Shot.

I'm open to error popups in all cases to avoid the event listener, though I don't know how passionate @johngruen is about it.

ianb:
You might want to add a fix for https://github.com/mozilla-services/pageshot/issues/2237 to this – the error message should be slightly different, but otherwise it's the same logic.

dannycoates:
Since it was easy to add the tab id I went ahead and finished it this way. We can change to an error prompt if this proves too slow, but I think this is a much better UX. @ianb ready for review again.

addon/webextension/background/main.js
```diff
@@ -40,8 +40,12 @@ window.main = (function () {
       });
   }
 
+  function opensMyShots(url) {
```
ianb: That one "s" between openMyShots and opensMyShots makes quite a bit of difference – easy to misread?  `shouldOpenMyShots` ?

addon/webextension/background/main.js
```diff
@@ -77,6 +81,26 @@ window.main = (function () {
         .then(() => sendEvent("start-shot", "context-menu")));
   }));
 
+  function urlEnabled(url) {
+    if (opensMyShots(url)) {
+      return true;
+    }
+    if (url.startsWith(backend) || /^(?:about|data|moz-extension):/i.test(url)) {
```
ianb: There might be a whitelist of WebExtension schemes, but I don't know what it is exactly

addon/webextension/background/main.js
```diff
@@ -77,6 +81,26 @@ window.main = (function () {
...
+      return false;
+    }
+    return true;
+  }
+
+  browser.tabs.onUpdated.addListener((id, info, tab) => {
```
ianb: catcher.watchFunction

addon/webextension/background/main.js
dannycoates: I looked on dxr, didn't see anything obvious

SoftVision-CosminMuntean:
The "Firefox Screenshots" button is now disabled on the mentioned pages, but the "Create Screenshot" option is still available in context menu.  
@ianb , @dannycoates, the option from context menu should not be disabled too? I'm asking this because I am not sure what is the intended behavior.

dannycoates:
Good question @SoftVision-CosminMuntean. Ideally we should remove or disable the menu item from those pages but I don't think we can do that without more code. We could easily add `documentUrlPatterns` to the menu which would filter out *most* of the urls we want.

PR #2384, opened 2017-03-14
Optional bootstrap

ianb:
and a cleanup commit (not related)

### DIFF
--- a/addon/webextension/background/loadSelector.js
+++ b/addon/webextension/background/loadSelector.js
@@ -1,4 +1,4 @@
-/* globals browser, catcher */
+/* globals browser */
 
 window.loadSelector = (function () {
   // These modules are loaded in order, and some need to be listed before others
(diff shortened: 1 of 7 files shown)


addon/webextension/background/loadSelector.js
```diff
@@ -25,14 +25,12 @@ window.loadSelector = (function () {
...
-          file: script
-        }, () => {
-          if (browser.runtime.lastError
-              && browser.runtime.lastError.message != "Script returned non-structured-clonable data") {
-              catcher.unhandled(browser.runtime.lastError, {script: script});
-              console.error("Error loading script", script, ":", browser.runtime.lastError);
-          }
+        return browser.tabs.executeScript({file: script}).then(() => {
+          return;
+        }, (error) => {
```
pdehaan: should this just use a `.catch()` block instead of an empty `.then()`?

PR #2386, opened 2017-03-14
WIP solution to issue#2315

garbados:
Hello! First PR :D

This solution to [fix bin/run-addon to work on non-osx](https://github.com/mozilla-services/pageshot/issues/2315) uses `uname` to determine the host OS, setting `$open` with the command to open Firefox appropriately. By default, `$open` assumes a Mac host OS which has the `open` command available.

Currently, the script only modifies the default `$open` value on Linux systems, excluding Windows, FreeBSD, etc., because I don't have those systems available for testing.

This solution lacks tests because I couldn't find documentation within the project about testing procedures for scripts. I don't consider it complete it until it has tests, and existing TODOs have been resolved.

Anyway, what do you think?


bin/run-addon
```diff
@@ -12,7 +12,9 @@ firefoxes="
 /Applications/FirefoxNightly.app
 /Applications/FirefoxDeveloperEdition.app
 /Applications/FirefoxAurora.app
+/usr/bin/firefox
 "
+# TODO add more firefoxes locations? use `which` maybe?
```
ianb: Adding `$(which firefox)` to this list should work, but there's a possibility if it's not found that it will cause the script to error, in which case you'd want to use `$(which firefox || true)` (to swallow the error).  But please test first, I'm not sure what errors in scripts in string substitutions do.

bin/run-addon
```diff
@@ -79,6 +81,14 @@ while [[ -n "$1" ]] ; do
   esac
 done
 
+open="open $binary" # assumes Mac OS, availability of 'open' command
```
ianb: This seems fine – defaulting to Linux and looking for `"$(uname)" = "Darwin" ` would also be fine.

bin/run-addon
```diff
@@ -79,6 +81,14 @@ while [[ -n "$1" ]] ; do
...
+unamestr=$(uname)
+if [[ "$unamestr" == "Linux" ]]; then
+  open=$binary
+# elif [[ "$unamestr" == "" ]]; then
+#  # TODO add cases for Windows, FreeBSD, etc.
```
ianb: FreeBSD should be the same as Linux, which is maybe a good reason to test for Mac and default to Linux.  On Windows the development process is probably very broken in ways I haven't even considered!

bin/run-addon
```diff
@@ -79,6 +81,14 @@ while [[ -n "$1" ]] ; do
   esac
 done
 
+open="open $binary" # assumes Mac OS, availability of 'open' command
+unamestr=$(uname)
+if [[ "$unamestr" == "Linux" ]]; then
```
ianb: You can use a single `=` here

bin/run-addon
```diff
@@ -92,15 +102,15 @@ if [[ -n "$setup_profile" ]] ; then
...
     echo "Creating profile in $profile_dir running:"
-    echo " " open "$binary" -n --args -new-instance -no-remote -CreateProfile "PageShotTestProfile $profile_dir"
-    open "$binary" -n --args -new-instance -no-remote -CreateProfile "PageShotTestProfile $profile_dir"
+    echo " " $open -n --args -new-instance -no-remote -CreateProfile "PageShotTestProfile $profile_dir"
+    $open -n --args -new-instance -no-remote -CreateProfile "PageShotTestProfile $profile_dir"
     echo "Opening profile running:"
-    echo " " open "$binary" -n --args -new-instance -no-remote -profile "$profile_dir" -url http://locahost:10080
-    open "$binary" -n --args -new-instance -no-remote -profile "$profile_dir" -url http://locahost:10080
+    echo " " $open -n --args -new-instance -no-remote -profile "$profile_dir" -url http://locahost:10080
+    $open -n --args -new-instance -no-remote -profile "$profile_dir" -url http://locahost:10080
```
ianb: You should update your repository – it looks like you are working from an old revision.  In particular this has the typo of `locahost` instead of `localhost` which was fixed a while ago (and gives this PR a merge conflict)

garbados:
Thanks for the feedback! I made changes responding to the review. Anything else to do on my end?

ianb:
Looks good, thanks!

PR #2388, opened 2017-03-15
Add CONTRIBUTORS.md file

### LINKED ISSUE #2380: Add Contributors.md
Could be two bugs, but we should add all things contrib related!

pdehaan:
I generated this file manually using my script from https://github.com/mozilla-services/pageshot/issues/2380#issuecomment-286587433 &mdash; then I trimmed a couple of unwanted contributors (Bizarro and greenkeeperio-bot).

Not sure if the want to turn this into some ./bin/contributors script and hook into the Makefile so we can do this in a more automated way on each release (since we'll presumably have new contributors in the future).

Also noteworthy is that this only determines contributors based on Git commits, and doesn't include names of other people behind the scenes (such as our excellent UX team, UR team, legal team, l20n team, catering team, etc).

Fixes #2380

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,20 @@
+Abel 'Akronix' Serrano Juste
+Ben Ennis
+Bram Pitoyo
+Chuck Harmston
+Cory Price
+Daniel Thorn
+Danny Coates
+Donovan Preston
+Erik Vold
+Ian Bicking
+Jared Hirsch
+John Gruen
+Kit Cambridge
+Mark Banner
+Michael R. Bernstein
+niharikak101
+Peter deHaan
+Sadika Sumanapala
+Wil Clouser
+YFdyh000


wresuolc:
We should err on the side of being too broad here.  I suggest we give up hope of this being automated and spend some time collecting names from everyone involved.  I can do a first pass.

CONTRIBUTORS.md
```diff
@@ -0,0 +1,20 @@
+Abel 'Akronix' Serrano Juste
```
wresuolc: These should probably be a bulleted list?  GH render is showing them all on one line

ianb:
Some of the people just did one-line README fixes and stuff.  But if you run `git shortlog` you see the commits and it's pretty obvious.

As a release procedure we could do `git shortlog stable..master` to see if we need to make additions.

wresuolc:
first draft @ https://public.etherpad-mozilla.org/p/pwkFIoEL4e .  please edit as needed

CONTRIBUTORS.md
johngruen: @wresuolc markdown needs two spaces to represent a RETURN so
`Ben Ennis  ` not  
`Ben Ennis `

pdehaan:
I just noticed we have an empty [/CREDITS.md](https://github.com/mozilla-services/screenshots/blob/master/CREDITS.md) file in the root folder. Not sure if this is the same general concept as CONTRIBUTORS, or if we intended CREDITS.md to have something else: https://github.com/mozilla-services/screenshots/blob/master/CREDITS.md

CONTRIBUTORS.md
pdehaan: Since we aren't using Markdown at all in this file, should we just rename it to CONTRIBUTORS (sans extension) so it renders as plain text?

ianb:
I guess CREDITS.md used to refer to a library that we no longer use.  I guess we should remove that file.

ianb:
@pdehaan I'll land this one

ianb:
I did this in PR #2492

PR #2389, opened 2017-03-15
added selector ui toggling

dannycoates:
I apologize for the size of the diff. I ran into a few other bugs as I went. More comments in the diff...


addon/webextension/background/communication.js
```diff
@@ -18,6 +18,7 @@ window.communication = (function () {
     let func = registeredFunctions[req.funcName];
     let result;
     try {
+      req.args.push(sender);
```
dannycoates: I found I needed the `tab.id` in the `closeSelector` handler. It's probably generally useful to make this available, and I didn't find a case where it broke anything.

addon/webextension/README.md
```diff
@@ -8,7 +8,7 @@ This contains all the files for the WebExtension.  Most files are not "built", b
   2. The selector content worker, in `selector/`
   3. The site helper that communicates with the Page Shot website, in `site-helper.js`
 - The background process is loaded according to a list in `manifest.json.template` – if there are load order dependencies, they must be manually managed with that list
-- The selector content worker is loaded with a list in `background/loadSelector.js`.  Again this must be manually maintained.
+- The selector content worker is loaded with a list in `background/selectorLoader.js`.  Again this must be manually maintained.
```
dannycoates: I combined the functionality of `loadSelector.js` and `lifecycle.js` into `selectorLoader.js`. It was a bit easier for me to reason through it that way.

addon/webextension/background/main.js
```diff
@@ -25,6 +26,17 @@ window.main = (function () {
     }
   }
 
+  function setIconActive(active, tabId) {
+    const path = active ? "icons/pageshot-icon-green-38.png" : "icons/pageshot-icon-38.png";
+    browser.browserAction.setIcon({path, tabId});
```
dannycoates: Changes the icon to green when the selector is visible. @johngruen I know the color is bad, please change it.

addon/webextension/background/main.js
```diff
@@ -82,6 +94,10 @@ window.main = (function () {
     }
   });
 
+  communication.register("closeSelector", sender => {
```
dannycoates: I needed this to signal that the icon should go back to grey since there's more ways to get out of the selector ui than getting in. Originally this did more but I narrowed it down to this.

addon/webextension/background/selectorLoader.js
```diff
@@ -0,0 +1,74 @@
...
+    "selector/documentMetadata.js",
+    "selector/util.js",
+    "selector/ui.js",
+    "selector/shooter.js",
+    "selector/uicontrol.js"
+  ];
+
+  exports.unloadIfLoaded = function () {
+    return browser.tabs.executeScript({
+      code: "window.selectorLoader && window.selectorLoader.unloadModules()"
```
dannycoates: This replaces the code in `lifecycle.js` that unloaded the previous run. Since the toggle disables on the second press we don't want to load the scripts again after this.

addon/webextension/background/selectorLoader.js
```diff
@@ -0,0 +1,74 @@
...
+    }).then(result => {
+      return result && result.toString() === "true";
+    });
+  };
+
+  exports.loadModules = function () {
+    return scripts.reduce(
```
dannycoates: I went a little fancier here with the `reduce` but we can go back to the other way if preferred.

addon/webextension/background/selectorLoader.js
```diff
@@ -0,0 +1,74 @@
...
+        previous.then(() => browser.tabs.executeScript({file})),
+      Promise.resolve(null))
+      .catch(err => {
+        console.error(`Error loading script: ${err}`);
+        catcher.unhandled(err);
+        exports.unloadModules();
+      });
+  };
+
+  exports.unloadModules = function () {
```
dannycoates: This deletes everything including catcher (last). I think the unloading is clean enough that its safe to do it.

addon/webextension/selector/shooter.js
```diff
@@ -81,8 +77,7 @@ window.shooter = (function () { // eslint-disable-line no-unused-vars
     }).then((url) => {
       const copied = window.clipboard.copy(url);
       return callBackground("openShot", { url, copied });
-    }));
-    exports.deactivate();
+    }).then(() => uicontrol.deactivate()));
```
dannycoates: `deactivate` calls `selectorLoader.unloadModules` so it needs to wait until after the background call

addon/webextension/selector/shooter.js
```diff
@@ -104,11 +99,6 @@ window.shooter = (function () { // eslint-disable-line no-unused-vars
     }));
   };
 
-  /** Happens when the URL changes via window.history */
-  exports.popstate = catcher.watchFunction(function () {
```
dannycoates: popstate wasn't working so I removed this

addon/webextension/selector/ui.js
```diff
@@ -519,6 +519,7 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
       }
     }
     exports.iframe.remove();
+    exports.Box.remove();
```
dannycoates: moved from `uicontrol.deactivate`

addon/webextension/selector/uicontrol.js
```diff
@@ -353,8 +353,9 @@ window.uicontrol = (function () {
       this.cachedEl = null;
       ui.Box.remove();
       ui.WholePageOverlay.display(standardOverlayCallbacks);
-      document.addEventListener("keyup", watchFunction(keyupHandler), false);
```
dannycoates: this wasn't getting removed in `removeHandlers` since the wrapped function isn't the same as `keyupHandler`

addon/webextension/selector/uicontrol.js
```diff
@@ -778,12 +779,11 @@ window.uicontrol = (function () {
 
   exports.deactivate = function () {
     try {
-      ui.Box.remove();
-      ui.remove();
```
dannycoates: called as `unload` by `unloadModules`

addon/webextension/selector/uicontrol.js
```diff
@@ -778,12 +779,11 @@ window.uicontrol = (function () {
...
-      removeHandlers();
```
dannycoates: called on `unload`

addon/webextension/selector/uicontrol.js
```diff
@@ -822,6 +821,7 @@ window.uicontrol = (function () {
       primedDocumentHandlers.set(eventName, fn);
     });
     primedDocumentHandlers.set("keyup", keyupHandler);
+    window.addEventListener('beforeunload', beforeunloadHandler);
```
dannycoates: the `popstate` event wasn't being triggered by back/forward buttons. `beforeunload` is the best option I found. `unload` seems to be too late.

addon/webextension/selector/uicontrol.js
```diff
@@ -832,15 +832,19 @@ window.uicontrol = (function () {
...
+
   function keyupHandler(event) {
     if (event.ctrlKey || event.shiftKey || event.altKey || event.metaKey) {
       // Modified
       return;
     }
     if ((event.key || event.code) === "Escape") {
-      exports.deactivate();
       sendEvent("cancel-shot", "keyboard-escape");
-      shooter.deactivate();
```
dannycoates: `shooter.deactivate` was just calling `uicontrol.deactivate` so I removed it

ianb:
There should be a specific event for this, like `addon/cancel-selection/toolbar-pageshot-button`

This should also be noted in METRICS.md

addon/webextension/background/communication.js
ianb: We'll just have to be sure that we don't have any functions with implicitly optional last arguments.  Which, on consideration... we do!  `sendEvent` works this way.

We could overload `this` with the sender, but that's kind of a lame form of optional argument.  We could be very careful about how sendEvent works and look for any object arguments where we expect strings, though there's an optional options argument at the end of sendEvent that is an object.  We could also change sendEvent in the selector process to always send the expected number of arguments.  I'm not sure what the best approach is.

addon/webextension/background/main.js
ianb: We should have a followup bug, but it doesn't have to block this landing.

addon/webextension/background/selectorLoader.js
```diff
@@ -0,0 +1,74 @@
...
+    return scripts.reduce(
+      (previous, file) =>
+        previous.then(() => browser.tabs.executeScript({file})),
+      Promise.resolve(null))
+      .catch(err => {
+        console.error(`Error loading script: ${err}`);
+        catcher.unhandled(err);
+        exports.unloadModules();
+      });
+  };
```
ianb: I liked it the previous way, reduce always seems too clever to me.  Also I believe the previous logic included the script name when there was an error, and when there's an error the script name is very helpful (and not always available).

Also I think unloadModules is a good guard when there's an error, but we should let the caller handle the error and not do catcher.unhandled here.

addon/webextension/background/communication.js
dannycoates: I'm gonna move `sender` to be the first argument. The `register`ed callback will change signature to `(sender, ...args)`

dannycoates:
Thanks @ianb, new commit addresses those changes.

addon/webextension/background/main.js
dannycoates: #2403

Issue #2319, opened 2017-03-08
Document or script patch creation


ianb:
After copying files into the Firefox tree (#2318) we need to produce a patch and put the patch in Bugzilla.  This could be documented or scripted, or some combination of the two.

Standard8:
I'd suggest scripted. I had something for Loop that could do this, so I'll see if I can pull it out.

Standard8:
We've now got a scripted way of exporting to mozilla-central (#2407).

I don't have requirements for creating bugs, but it should be pretty easy to adapt https://github.com/mozilla/loop/blob/master/bin/create_release.py to file bugs automatically & post patches. 

The one thing that is slightly different is that the export_mc.py script created in #2407 already does the creating of a branch & commit from which to push the changes. Hence the first part of `releaseMozillaCentral` just needs to make a call to that script.

6a68:
Hmm. Looking at the create_release.py script, I think it would be good to create a bugzilla component for firefox release requests and security bugs.

johngruen:
Closing for now since current process seems to be working

PR #2393, opened 2017-03-15
Add flake8 lint checking of python files

Standard8:
For #2319 I'm creating a script that's going to use python (because that's currently the one language I can find a decent supported bugzilla API integration for).

Hence as part of that, we're going to need a .venv, and we might as well get flake8 set up as well to help with getting the scripts clean.


Makefile
```diff
@@ -103,6 +104,14 @@ else
   FIND_COMMAND := find -E $(GIT_EXPORT_DIR)
 endif
 
+$(VENV): bin/require.pip
+	virtualenv -p python2.7 $(VENV)
+	. $(VENV)/bin/activate && pip install -r bin/require.pip
```
Standard8: > This could be $(VENV)/bin/pip install -r bin/require.pip

Actually, that doesn't seem to work for pip install - looks like it needs the activation first.

PR #2397, opened 2017-03-15
Observe prefs and control webextension startup and shutdown via a pref

ianb:
For handoff to @6a68

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -1,38 +1,118 @@
-/* globals Components, AddonManager */
-/* eslint-disable no-unused-vars */
+/* globals AddonManager, Components, Services */
 
 const OLD_ADDON_PREF_NAME = "extensions.jid1-NeEaf3sAHdKHPA@jetpack.deviceIdInfo";
 const OLD_ADDON_ID = "jid1-NeEaf3sAHdKHPA@jetpack";
+const ADDON_ID = "pageshot@mozilla.org";
+const TELEMETRY_PREF = "toolkit.telemetry.enabled";
+const PREF_BRANCH = "extensions.pageshot.";
+const USER_DISABLE_PREF = "extensions.pageshot.disabled";
+const SYSTEM_DISABLE_PREF = "extensions.pageshot.system-disabled";
 
-const prefs = Components.classes["@mozilla.org/preferences-service;1"]
-  .getService(Components.interfaces.nsIPrefBranch);
-Components.utils.import("resource://gre/modules/AddonManager.jsm");
+const { interfaces: Ci, utils: Cu } = Components;
+Cu.import("resource://gre/modules/AddonManager.jsm");
+Cu.import("resource://gre/modules/Console.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+const { EmbeddedExtensionManager } = Cu.import("resource://gre/modules/LegacyExtensionsUtils.jsm");
 
-function startup(data, reason) {
-  data.webExtension.startup().then((api) => {
-    const {browser} = api;
-    browser.runtime.onMessage.addListener(handleMessage);
+const prefs = Services.prefs;
+const prefObserver = {
+  register: function() {
+    prefs.addObserver(PREF_BRANCH, this, false);
+  },
+
+  unregister: function() {
+    prefs.removeObserver(PREF_BRANCH, this);
+  },
+
+  observe: function(aSubject, aTopic, aData) {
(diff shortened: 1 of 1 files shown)


addon/bootstrap.js
```diff
@@ -3,15 +3,51 @@
...
+const USER_ENABLE_PREF = "extensions.pageshot.userdisabled";
+const SYSTEM_ENABLE_PREF = "extensions.pageshot.systemdisabled";
 
-const prefs = Components.classes["@mozilla.org/preferences-service;1"]
-  .getService(Components.interfaces.nsIPrefBranch);
 Components.utils.import("resource://gre/modules/AddonManager.jsm");
+Components.utils.import("resource://gre/modules/Services.jsm");
+const { EmbeddedExtensionManager } = Components.utils.import("resource://gre/modules/LegacyExtensionsUtils.jsm");
+
+const prefs = Services.prefs;
```
pdehaan:
```
/home/travis/build/mozilla-services/pageshot/addon/bootstrap.js
  15:15  error  'Services' is not defined  no-undef
  78:19  error  'Services' is not defined  no-undef
```

Not sure if we need to tweak the ESLint envs, or just manually add another `globals` to appease the ESLint gods.

ianb: I've been generous adding `/* globals */` to my code

pdehaan: > @ianb: _"I've been generous adding `/* globals */` to my code"_

I've noticed. We can also add them to the root .eslintrc.yml file, or even add nested .eslintrc.yml files, so you don't have to shotgun a small manifesto of global vars at the top of each file.

ianb: They are like pretend import statements!

6a68: Ah yeah, thanks @pdehaan. Updated globals.

6a68:
@ianb Would you mind giving this branch a try? It seems to be totally working for me.

Cases to try out:

- start without the `extensions.pageshot.userdisabled` pref set, then toggle it on and off a few times
- start with userdisabled set to false (should check the pref, then init the webextension)
- start with userdisabled set to true, verify it's not loaded, then toggle the pref and verify it works

Then, same three cases for `extensions.pageshot.systemdisabled`.

If this seems to be working for you as well, feel free to review now, and I'll plan to squash before merging.

addon/bootstrap.js
```diff
@@ -1,17 +1,65 @@
-/* globals Components, AddonManager */
+/* globals AddonManager, Components, Services */
 /* eslint-disable no-unused-vars */
 
 const OLD_ADDON_PREF_NAME = "extensions.jid1-NeEaf3sAHdKHPA@jetpack.deviceIdInfo";
 const OLD_ADDON_ID = "jid1-NeEaf3sAHdKHPA@jetpack";
+const ADDON_ID = "pageshot@mozilla.org";
+const TELEMETRY_PREF = "toolkit.telemetry.enabled";
+const USER_ENABLE_PREF = "extensions.pageshot.userdisabled";
+const SYSTEM_ENABLE_PREF = "extensions.pageshot.systemdisabled";
```
ianb: Well, this is my fault, but these variables have a really bad name – ENABLE when in fact they are DISABLE

addon/bootstrap.js
```diff
@@ -1,17 +1,65 @@
...
+
+function shouldStartWebExtension() {
+  return ! (
+    prefs.prefHasUserValue(USER_ENABLE_PREF) &&
+    prefs.getBoolPref(USER_ENABLE_PREF)
+  ) || (
+    prefs.getPrefType(SYSTEM_ENABLE_PREF) &&
+    prefs.getBoolPref(SYSTEM_ENABLE_PREF)
+  );
+}
```
ianb: This is probably my fault too, but the expression here is incorrect.  I think it's `(! USER_ENABLE_PREF) || SYSTEM_ENABLE_PREF` – as a result setting `extensions.pageshot.systemdisabled` to true forces Page Shot *on* rather than off.

6a68:
thanks for looking! I'll fixup the logic

6a68:
@ianb Mind taking another look? Fixed up the disable check logic, did a general cleanup pass, and squashed everything into one commit. Hopefully bootstrap.js will get through Firefox review without too many problems :-P

Note that I renamed the user pref to just `extensions.pageshot.disabled`, because the `system-disabled` pref will be gone within a few weeks of launch, so calling it `user-disabled` seemed unnecessary.

PR #2404, opened 2017-03-16
WIP: use window.crypto.getRandomValues

garbados:
This is a WIP solution to https://github.com/mozilla-services/pageshot/issues/2299

I found [this pastebin](http://pastebin.com/JSgC7eej) describing a method of creating UUIDs using `window.crypto.getRandomValues` and added it to two locations. I haven't added tests because I haven't found testing guidelines (ex: what files detail tests for which environments, server or client) but I would feel irresponsible merging this without them.


garbados:
The pastebin method, though lacking a known author besides "A guest, on December 7, 2017", makes sense to me, and matches the spec as I understand it. Still it irks my paranoia, since we're fundamentally discussing a security issue.

ianb:
The function in shot.js may be harder to change – that file is used on both the server and client, so `window` shouldn't be used.  OTOH, looking at how that function is used there's no real security concern, and Math.random would be fine for that case.  Also I notice while the function in shot.js is distractingly called `makeUuid()` it actually returns a random string.  I think you could just leave that function alone.

The function that *does* matter in terms of security is in randomString.

For makeUuid.js you could still use the string substitution method (which reads kind of nicely), and just use `var r = u[index]; index++`

garbados:
Understood. I'll make those changes. Thanks for the feedback!

garbados:
OK, I restored `shared/shot.js` to match the version on master, and modified `makeUuid.js` to use getRandomValues with the pre-existing 'xxxxx...'.replace approach. How's it look?

PR #2407, opened 2017-03-16
chore: Simplify exporting to mozilla-central by adding a script which exports and creates branches and commits.

Standard8:
This is at least a partial fix for #2319 - it creates a branch and a commit of the exported add-on. At the moment you still need bug 1346825 applied on the default branch for this to work.

It works slightly better in git (or at least, its slightly easier to recover from) - for Mercurial it uses bookmarks, which I've found a little confusing (it creates a bookmark ok, it is getting back to 'default' and killing the commits that is the issue there).

In any case, it should be reasonably adaptable if we need to change it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,7 +253,11 @@ all: addon server
 .PHONY: clean
 clean:
 	rm -rf build/ addon/webextension/build/ addon/webextension/manifest.json addon/webextension/_locales/
+
+.PHONY: distclean
+distclean: clean
 	rm -rf $(VENV)
+	rm -rf ./node_modules
 
 .PHONY: help
 help:
--- a/bin/export_mc.py
+++ b/bin/export_mc.py
@@ -1,0 +1,185 @@
+#!.venv/bin/python
+
+import argparse
+import os
+import subprocess
+import sys
+from git import Repo
+
+DEFAULTS = {
+    "local": "http://localhost:10080/",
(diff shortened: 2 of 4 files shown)


bin/export_mc.py
```diff
@@ -0,0 +1,185 @@
...
+import os
+import subprocess
+import sys
+from git import Repo
+
+DEFAULTS = {
+    "local": "http://localhost:10080/",
+    "dev": "https://pageshot.dev.mozaws.net/",
+    "stage": "https://pageshot.stage.mozaws.net/",
+    "prod": "https://pageshot.net"
```
pdehaan: not sure if it matters, but prod doesnt have a trailing `/`, but the other envs do.

Makefile
```diff
@@ -253,6 +253,9 @@ all: addon server
 .PHONY: clean
 clean:
 	rm -rf build/ addon/webextension/build/ addon/webextension/manifest.json addon/webextension/_locales/
+
+.PHONY: distclean
+distclean: clean
```
ianb: Maybe this should also rm -rf node_modules/ ?  But I'm not entirely sure what distclean should do compared to clean

Standard8: Distclean generally gets rid of the support files/dependencies as well as the build stuff. The simple clean is just for the build (or that's how we did it in Loop)

Standard8:
@ianb: Updated for your comments.

PR #2413, opened 2017-03-17
Sitehelper login

ianb:
This has a couple cleanup commits that can be looked at separately from the main commit

Testing this is not terribly easy.  You have to start run-addon and create a shot, then save the URL of that shot (not as a bookmark).  Then shut the browser, start it up again, and try to visit the shot page.  It should briefly appear unauthenticated, and then when the add-on auth fires it will change the page to be authenticated and you should be the owner.  In theory if you delete cookies and revisit a page it will also re-authenticate.

This doesn't handle My Shots, I'm making a followup for that.


addon/webextension/sitehelper.js
```diff
@@ -0,0 +1,41 @@
...
+
+window.sitehelper = (function () {
+
+  catcher.registerHandler((errorObj) => {
+    callBackground("reportError", errorObj);
+  });
+
+
+  function sendCustomEvent(name, detail) {
+    if (typeof detail == "object") {
```
6a68: I guess this is okay: it's not an accurate check for objecthood, but it's unlikely that `null`, an array, or some oddball object like `Math` would ever get passed in here.

ianb: The type coercion here is a little sloppy, but objects in particular (as opposed to strings and numbers and null) trigger permission problems when transferring them between the add-on and content.  I forgot about null==object though, that's an annoying case, but at least undefined (as when the argument isn't given) will be left alone.

PR #2448, opened 2017-03-21
provisional addon strings

johngruen:
@6a68 this is still provisional but probably worth merging. I kind of want to review some of the old keys b/c they're not very didactic.

### DIFF
--- a/locales/en-US/webextension.properties
+++ b/locales/en-US/webextension.properties
@@ -1,7 +1,7 @@
-addonDescription = Firefox Screenshots takes clips and screenshots from pages, and can save a permanent copy of a page.
-addonAuthorsList = Ian Bicking, Donovan Preston, and Bram Pitoyo
+addonDescription = Firefox Screenshots takes clips and screenshots from the Web and lets you save them temporarily or permanently.
+addonAuthorsList = Mozilla <screenshots-feedback@mozilla.org>
 toolbarButtonLabel = Take a shot
-contextMenuLabel = Create screenshot
+contextMenuLabel = Take a Screenshot
 myShotsLink = My Shots
 screenshotInstructions = Drag or click on the page to select a region. Press ESC to cancel.
 saveScreenshotSelectedArea = Save
@@ -15,14 +15,24 @@ notificationLinkCopiedTitle = Link Copied
 # modifier key used in the shortcut (do not translate it): for example, Ctrl-V
 # on Windows systems.
 notificationLinkCopiedDetails = The link to your shot has been copied to the clipboard. Press {meta_key}-V to paste.
+# Error Strings
 requestErrorTitle = Firefox Screenshots is out of order.
-requestErrorDetails = Your shot was not saved. We apologize for the inconvenience. Try again soon.
-connectionErrorTitle = Cannot connect to the Firefox Screenshots server.
-connectionErrorDetails = There may be a problem with the service or with your network connection.
-loginErrorDetails = Your shot was not saved. There was an error authenticating with the server.
-loginConnectionErrorDetails = There may be a problem with the service or your network connection.
-unshootablePageErrorTitle = Page cannot be screenshotted.
-unshootablePageErrorDetails = This is not a normal web page, and Firefox cannot capture screenshots from it.
+requestErrorDetails = Sorry! We couldn’t save your shot. Please try again later.
+connectionErrorTitle = We can’t connect to the Firefox Screenshots server.
+connectionErrorDetails = Please check your network connection. If you are able to connect to the Internet, there may be a temporary problem with the Firefox Screenshots service.
+loginErrorDetails = We couldn’t save your shot because we couldn’t access your account on the Firefox Screenshots server.
+unshootablePageErrorTitle = We can’t screenshot this page.
+unshootablePageErrorDetails = This isn’t a standard Web page, so Firefox Screenshots can’t take a screenshot of it.
 selfScreenshotErrorTitle = You can’t take a shot of a Firefox Screenshots page!
-genericErrorTitle = Firefox Screenshots went haywire.
-genericErrorDetails = Try again or take a shot on another page?
+genericErrorTitle = Whoa! Screenshots went haywire.
+genericErrorDetails = We’re not sure what just happened. Care to try again or take a shot of a different page?
+# Onboarding Strings
+tourHeaderOne = Firefox Screenshots
+tourHeaderOneSuper = Beta
+tourBodyOne = Take, save, and share screenshots without leaving Firefox.
+tourHeaderTwo = Capture Just What You Want
+tourBodyTwo = Click and drag to capture just a portion of a page. You can also hover to highlight your selection.
+tourHeaderThree = As You Like it
+tourBodyThree = Save your cropped shots to the Web for easier sharing, or download them to your computer. You also can click on the My Shots button to find all the shots you’ve taken.
+tourHeaderFour = Capture Windows or Entire Pages
+tourBodyFour = Select the buttons in the upper right to capture the visible area in the window or to capture an entire page.
--- a/addon/webextension/background/senderror.js
+++ b/addon/webextension/background/senderror.js
@@ -18,7 +18,7 @@ window.errorpopup = (function () {
     },
     LOGIN_CONNECTION_ERROR: {
       title: browser.i18n.getMessage("connectionErrorTitle"),
-      info: browser.i18n.getMessage("loginConnectionErrorDetails")
+      info: browser.i18n.getMessage("connectionErrorDetails")
     },
     UNSHOOTABLE_PAGE: {
       title: browser.i18n.getMessage("unshootablePageErrorTitle"),


locales/en-US/webextension.properties
```diff
@@ -1,9 +1,10 @@
-addonDescription = Firefox Screenshots takes clips and screenshots from pages, and can save a permanent copy of a page.
-addonAuthorsList = Ian Bicking, Donovan Preston, and Bram Pitoyo
+addonDescription = Firefox Screenshots takes clips and screenshots from the Web and lets you save them temporarily or permanently.
+addonAuthorsList = Ian Bicking, Donovan Preston, Jared Hirsch, Danny Coates, Morpheus Chen, Helen Huang, Mark Liang, Fang Shih, John Gruen, Bram Pitoyo, Sharon Bautista and Sean Martell, Michelle Heubusch
```
6a68: What about just `Mozilla <https://mozilla.org>`, like we do with Test Pilot's package.json?

locales/en-US/webextension.properties
```diff
@@ -1,9 +1,10 @@
...
 toolbarButtonLabel = Take a shot
-contextMenuLabel = Create screenshot
+contextMenuLabel = Take a Screenshot
 myShotsLink = My Shots
-screenshotInstructions = Drag or click on the page to select a region. Press ESC to cancel.
+screenshotInstructionOne = Drag or click on the page to select a region.
+screenshotIntrouctionTwo = Press ESC to cancel.
```
6a68: typo

locales/en-US/webextension.properties
ianb: We could use screenshots-feedback@mozilla.org (which exists)

locales/en-US/webextension.properties
```diff
@@ -15,14 +16,26 @@ notificationLinkCopiedTitle = Link Copied
...
-connectionErrorDetails = There may be a problem with the service or with your network connection.
-loginErrorDetails = Your shot was not saved. There was an error authenticating with the server.
-loginConnectionErrorDetails = There may be a problem with the service or your network connection.
-unshootablePageErrorTitle = Page cannot be screenshotted.
-unshootablePageErrorDetails = This is not a normal web page, and Firefox cannot capture screenshots from it.
+requestErrorDetails = Sorry! We couldn’t save your shot. Please try again later.
+connectionErrorTitle = We can’t connect to the Firefox Screenshots server.
+connectionErrorDetails = Please check your network connection. If you are able to connect to the Internet, there may be a temporary problem with the Firefox Screenshots service.
+loginErrorDetails = We couldn’t save your shot because we couldn’t access your account on the Firefox Screenshots server.
+loginConnectionErrorDetails = Please check your network connection. If you are able to connect to the Internet, there may be a temporary problem with the Firefox Screenshots service.
```
6a68: I am thinking we would not keep the separate string ID if the string itself is identical, it's wasted work for the translators. @flodolo, thoughts?

locales/en-US/webextension.properties
johngruen: yeah that could work

locales/en-US/webextension.properties
```diff
@@ -15,14 +16,26 @@ notificationLinkCopiedTitle = Link Copied
...
+unshootablePageErrorTitle = We can’t screenshot this page.
+unshootablePageErrorDetails = his isn’t a standard Web page, so Firefox Screenshots can’t take a screenshot of it.
```
6a68: typo: `This`, not `his`

locales/en-US/webextension.properties
6a68: Oh also, the template that renders this string will have to get updated to include both keys.

6a68:
@johngruen The new strings look good; I added some comments.

The /webextension/* files look like build output, so they should be removed.

locales/en-US/webextension.properties
flodolo: Yes, it doesn't make a lot of sense to have two separate IDs for the same string. `connectionErrorDetails` sounds good for both.

locales/en-US/webextension.properties
```diff
@@ -15,14 +16,26 @@ notificationLinkCopiedTitle = Link Copied
...
 selfScreenshotErrorTitle = You can’t take a shot of a Firefox Screenshots page!
-genericErrorTitle = Firefox Screenshots went haywire.
-genericErrorDetails = Try again or take a shot on another page?
+genericErrorTitle = Whoa! Screenshots went haywire.
+genericErrorDetails = We’re not sure what just happened. Care to try again or take a shot of a different page?
+## Onboarding Strings
```
flodolo: One # is enough

locales/en-US/webextension.properties
```diff
@@ -15,14 +16,26 @@ notificationLinkCopiedTitle = Link Copied
...
+tourHeaderOne = Firefox Screenshots
+tourHeaderOneSuper = Beta
+tourBodyOne = Take, save, and share screenshots without leaving Firefox.
+tourHeaderTwo = Capture Just What You Want
+tourBodyTwo = Click and drag to capture just a portion of a page.  You can also hover to highlight your your selection.
```
flodolo: No double spaces after periods, double your

locales/en-US/webextension.properties
```diff
@@ -15,14 +16,26 @@ notificationLinkCopiedTitle = Link Copied
...
+tourHeaderThree = As You Like it
+tourBodyThree = Save your cropped shots to the Web for easier sharing, or download them to your computer. You also can click on the My Shots button to find all the shots you’ve taken.
+tourHeaderFour = Capture Windows or Entire Pages
+tourBodyFour = Select the buttons in the upper right to capture the visible area in the window or to capture an entire page. Shot Taken Push Notification
+>>>>>>> provisional addon strings
```
flodolo: Git config leftover?

locales/en-US/webextension.properties
```diff
@@ -15,14 +16,26 @@ notificationLinkCopiedTitle = Link Copied
...
+## Onboarding Strings
+tourHeaderOne = Firefox Screenshots
+tourHeaderOneSuper = Beta
+tourBodyOne = Take, save, and share screenshots without leaving Firefox.
+tourHeaderTwo = Capture Just What You Want
+tourBodyTwo = Click and drag to capture just a portion of a page.  You can also hover to highlight your your selection.
+tourHeaderThree = As You Like it
+tourBodyThree = Save your cropped shots to the Web for easier sharing, or download them to your computer. You also can click on the My Shots button to find all the shots you’ve taken.
+tourHeaderFour = Capture Windows or Entire Pages
+tourBodyFour = Select the buttons in the upper right to capture the visible area in the window or to capture an entire page. Shot Taken Push Notification
```
flodolo: What's `Shot Taken Push Notification`?

johngruen:
this is what happens when i try to do stuff while in a meeting

johngruen:
@6a68 @flod, thanks for the review. Sorry about the distracted coding! LMK if there are any more issues here.

locales/en-US/webextension.properties
```diff
@@ -15,14 +16,24 @@ notificationLinkCopiedTitle = Link Copied
...
-genericErrorTitle = Firefox Screenshots went haywire.
-genericErrorDetails = Try again or take a shot on another page?
+genericErrorTitle = Whoa! Screenshots went haywire.
+genericErrorDetails = We’re not sure what just happened. Care to try again or take a shot of a different page?
+# Onboarding Strings
+tourHeaderOne = Firefox Screenshots
+tourHeaderOneSuper = Beta
+tourBodyOne = Take, save, and share screenshots without leaving Firefox.
+tourHeaderTwo = Capture Just What You Want
+tourBodyTwo = Click and drag to capture just a portion of a page. You can also hover to highlight your your selection.
```
flodolo: Still double your at the end

johngruen:
Thanks!

addon/webextension/selector/ui.js
```diff
@@ -222,7 +225,8 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
         <div class="full-page"></div>
       </div>
       `;
-      this.el.querySelector(".preview-instructions").textContent = browser.i18n.getMessage("screenshotInstructions");
+      this.el.querySelector(".preview-instructions-start").textContent = browser.i18n.getMessage("previewInstructionsStart");
+        this.el.querySelector(".preview-instructions-end").textContent = browser.i18n.getMessage("previewInstructionsEnd");
```
flodolo: Funny indentation?

6a68:
@johngruen No worries, I figured you were doing three other things at the same time :+1:

I'll take another look and rebase this branch real quick

locales/en-US/webextension.properties
6a68: Pretty annoying that github showed all this typos as fixed. Going through them now.

6a68:
@johngruen I added the remaining fixes, so you are off the hook here, except for one question: the string `tourBodyFour` had the awkward phrase `Shot Taken Push Notification` at the end (see [earlier comment](https://github.com/mozilla-services/screenshots/pull/2448#discussion_r107340829)). Could you confirm that phrase should have been removed?

We can do another PR if any last-minute changes arrive tomorrow, then I'll bump the bugzilla bug to get localizers started.

addon/webextension/selector/ui.js
```diff
@@ -231,7 +231,11 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
                <body>
                  <div class="preview-overlay">
                    <div class="fixed-container">
-                     <div class="preview-instructions"></div>
+                     <div class="preview-instructions">
+                       <span class="preview-instructions-step-one"></span>
+                       &nbsp;
+                       <span class="preview-instructions-step-two"></span>
```
6a68: Hey @stasm (or @flodolo), I'm not sure if this is the right way to put two strings on the same visual line such that it renders correctly in RTL languages: `span` + `&nbsp;` + `span`, and setting the `textContent` of each `span` to a different localized string.

Here's how this looks in regular English:

screen shot 2017-03-22 at 1 41 22 pm

If I set `dir="rtl"` in the `body` tag on the iframe, English RTL looks like this:

screen shot 2017-03-22 at 1 42 10 pm

Does that look correct?

Finally, the one annoyance about this approach is that, when the screen is narrow, the strings wrap a bit awkwardly:

screen shot 2017-03-22 at 1 41 31 pm

Any suggestions there? I could concatenate the strings and shove them into a single element, but wasn't sure that would work for RTL languages. @johngruen may have ideas for nicer wrapping in that situation, too.

6a68:
I'll wait to merge this till I get feedback on the string concatenation approach in the selector overlay, but I think it's otherwise ready.

addon/webextension/selector/ui.js
flodolo: The English result with forced RTL direction is expected, period is a weak character and follows RTL, the rest is LTR.

The real question is, why do we need to split this string in two? Can't we just use one? That would also solve the awkward wrapping.

6a68: @flodolo Sure, great point. I'll reconstruct the string.

PR #2449, opened 2017-03-21
Rename addon id

ianb:


Makefile
```diff
@@ -126,15 +122,15 @@ export_addon: addon
 zip: addon
 	# FIXME: should remove web-ext-artifacts/*.zip first
```
6a68: I just hit the bug this FIXME intends to fix (multiple zip files inside web-ext-artifacts causes `mv` to bail) :-P

PR #2452, opened 2017-03-21
revise metrics.md for 54 launch

clouserw:
r? @johngruen and anyone else

@ianb I noticed a bunch of checkboxes are struck through.  Should I just whack those lines?

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -1,10 +1,14 @@
+
 ## Firefox Screenshots Metrics
+*Last Update: 2017-03-21*
 
-A summary of the metrics Firefox Screenshots will record, and what we're looking for in those metrics.
+This document is a summary of the metrics Firefox Screenshots will record, how we're recording them, and what we're looking for in those metrics.  There are two main areas we'll look at:
 
-### Opt-out
+0. Metrics to tell us if Firefox Screenshots is affecting Firefox.  For instance, do Firefox Screenshots users engage longer with Firefox?
 
-The add-on tracks the Telemetry opt-out preference (`toolkit.telemetry.enabled`) each time the user presses the Screenshots button.  If this preference is false, or if there is any issue trying to fetch the preference, then no data is sent to the Screenshots server.
+0. Help us understand how people use Firefox Screenshots, so that we can improve the experience of the tool itself.  A/B tests fall under this umbrella.
+
+Additionally, it's important to recognize that Firefox Screenshots is an add-on in a browser as well as a server-side service.
 
 ### Key Metrics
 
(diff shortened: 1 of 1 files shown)


docs/METRICS.md
```diff
@@ -1,47 +1,22 @@
+
 ## Firefox Screenshots Metrics
+*Last Update: 2017-03-21*
 
-A summary of the metrics Firefox Screenshots will record, and what we're looking for in those metrics.
+This document is a summary of the metrics Firefox Screenshots will record, how we're recording them, and what we're looking for in those metrics.  There are two main areas we'll look at:
 
-### Opt-out
+0. Metrics to tell us if Firefox Screenshots is affecting Firefox.  For instance, do Firefox Screenshots users engage longer with Firefox?
```
ianb: We did have a talk with Brendan Colloran about this, and he was pretty confident that we couldn't find any meaningful way to understand this, except through controlled experiment.  That is, Firefox Screenshots users are going to use Firefox more, because people who use any feature are more active than the general population.

The one option – which I'd like us to consider at some point – is to actually do a controlled experiment by leaving some set of users with Screenshots disabled.  And maybe we could see some differences for the brief period when we are rolling out Screenshots (but probably not).

docs/METRICS.md
```diff
@@ -57,19 +32,17 @@ This is primarily a count of `web/visit/owner`.  The tool always opens the page
...
-Continuing to create shots indicates overall value to the user.  Sharing and revisiting confirm that the value is actually obtained (it's possible to fantasize that you *would* find value in a shot, while never actually realizing that value).  Lastly, evidence that people find Firefox Screenshots attractive when they see a shot, or that people would refer each other to use Firefox Screenshots, indicates potential for organic growth.
-
-We do not collect Net Promoter Score.
+### Data Collection
 
-### Usage Metrics
+Firefox Screenshots assigns each user a random ID (associated with their profile) when the add-on is installed. This ID is associated with all shots the user makes.  For the purpose of Google Analytics (GA) the ID is hashed. The same hashed ID is used for website visits and events, and for add-on events.
 
-This information is intended to help us make Firefox Screenshots better.
+No metrics data is sent to any servers until a user interacts with Page Shot.
```
ianb: Oops, "Page Shot"

docs/METRICS.md
```diff
@@ -258,10 +229,18 @@ This is stuff we get from including ga.js on Screenshots pages.
...
 1. Look at cohorts of individuals, reviewing their shot creation patterns
 2. How many people make a couple shots?  How many make a lot of shots?
 3. See if people are making shots of mostly private or mostly public pages
 4. Identify the dimensions of shots
-5. Compare shot URLs to a domain whitelist, to determine broad categories of sites that are popular to view
+5. Compare shot URLs to a domain allowlist, to determine broad categories of sites that are popular to view
+
+### Opt-out
+
+The add-on reads the Telemetry opt-out preference (`toolkit.telemetry.enabled`) before sending any metrics data to the servers.  If this preference is false, or if there is any issue trying to fetch the preference, then no data is sent.
```
ianb: We could be more specific that the add-on checks the value of that preference each time you click the Screenshots button.

docs/METRICS.md
wresuolc: We've been talking about that.  It sounds like you don't need to leave the set of users with the feature disabled for a super long time to get a decent result (longer is better, of course, but hey...).  If our rollout lasts a couple weeks, that's already useful data.  We might be able to keep a small portion out longer, but that schedule isn't finalized.

docs/METRICS.md
wresuolc: I thought their might be other entry points. Like, if a person hits the keyboard combo or if we eventually provide some other link (a my shots button?)

ianb: It's slightly ad hoc, but at each entry point we try to run `analytics.refreshTelemetryPref()` (it ended up this way because each entry point also has its own event, so we want to refresh the pref before sending the first event).  But yes, we try to do it at the beginning of each session.

docs/METRICS.md
```diff
@@ -57,19 +32,17 @@ This is primarily a count of `web/visit/owner`.  The tool always opens the page
...
-This information is intended to help us make Firefox Screenshots better.
+No metrics data is sent to any servers until a user interacts with Firefox Screenshots.
 
-We record an event stream of interaction with the add-on and website.  The events:
-
-(**Note**: per [#1183](https://github.com/mozilla-services/screenshots/issues/1183) we need to figure out if we are actually collecting these events properly)
+The add-on does not communicate directly with GA, instead it POSTs an event to the Firefox Screenshots server which, in turn, sends the event to GA.  This model allows us to only send the data we need for analysis instead of the more comprehensive data collection that google-analytics.js does.  Besides the event information specified below we also send the User-Agent string, and the add-on provides:
+* Any A/B tests the user is in, or if the user is in the control for a test
+* The application (“firefox”)
+* The Page Shot add-on version (“6.0.0”)
```
ckprice: "Page Shot"

ckprice:
Do we need anything about Sentry in here?

wresuolc:
None of the other metrics docs talk about sentry, so I don't think so

wresuolc:
thanks

PR #2453, opened 2017-03-22
Split screenshot iframe into preselection and selection iframes

### LINKED ISSUE #2162: Separate pre-selection iframe from selection iframe
Before you select something we put up an iframe with controls on it.  This iframe should be `position: fixed` (though it currently is not, and we watch for scroll events to move elements around).

After selection we want an iframe that covers the entire document and holds the selection.  This makes scrolling while selecting work.  This is the iframe we have currently.

It would work better if we swap iframes after selection.

dannycoates:
closes #2162

### DIFF
--- a/addon/webextension/selector/ui.js
+++ b/addon/webextension/selector/ui.js
@@ -47,17 +47,26 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
   });
 
   function makeEl(tagName, className) {
-    if (! iframe.document) {
+    if (! iframe.document()) {
       throw new Error("Attempted makeEl before iframe was initialized");
     }
-    let el = iframe.document.createElement(tagName);
+    let el = iframe.document().createElement(tagName);
     if (className) {
       el.className = className;
     }
     return el;
   }
 
-  let iframe = exports.iframe = {
+  function onResize() {
+    if (this.sizeTracking.windowDelayer) {
+      clearTimeout(this.sizeTracking.windowDelayer);
+    }
+    this.sizeTracking.windowDelayer = setTimeout(watchFunction(() => {
+      this.updateElementSize(true);
+    }), 50);
+  }
+
+  let iframeSelection = exports.iframeSelection = {
     element: null,
     addClassName: "",
     sizeTracking: {
@@ -72,7 +81,8 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
       return new Promise((resolve, reject) => {
         if (! this.element) {
           this.element = document.createElement("iframe");
-          this.element.id = "firefox-screenshots-iframe";
+          this.element.id = "firefox-screenshots-selection-iframe";
+          this.element.style.display = "none";
           this.element.style.zIndex = "99999999999";
           this.element.style.border = "none";
           this.element.style.position = "absolute";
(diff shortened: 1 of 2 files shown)


addon/webextension/selector/ui.js
```diff
@@ -198,83 +201,193 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
...
+    updateElementSize: function (force) {
+      // Note: if someone sizes down the page, then the iframe will keep the
+      // document from naturally shrinking.  We use force to temporarily hide
+      // the element so that we can tell if the document shrinks
+      const visible = this.element.style.display !== "none";
+      if (force && visible) {
+        this.element.style.display = "none";
+      }
+      let height = Math.max(
+        document.documentElement.clientHeight,
```
ianb: `updateElementSize` can be simpler for this iframe.  It should always be exactly `window.innerHeight` (and width), and doesn't have to track document changes.  `force` also doesn't matter for this iframe.

addon/webextension/selector/ui.js
```diff
@@ -198,83 +201,193 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
...
-      let scrollX = window.scrollX;
-      let scrollY = window.scrollY;
-      this.movingEl.style.top = scrollY + "px";
-      this.movingEl.style.left = scrollX + "px";
-      this.movingEl.style.width = window.innerWidth + "px";
-      this.movingEl.style.height = window.innerHeight + "px";
+    },
+
+    hide: function () {
+      window.removeEventListener("scroll", this.onScroll, false);
```
ianb: Everything related to scroll tracking can be removed.  Neither iframe should require it now.

addon/webextension/selector/ui.js
```diff
@@ -198,83 +201,193 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
...
   };
 
-  WholePageOverlay.onScroll = watchFunction(WholePageOverlay.onScroll.bind(WholePageOverlay));
+  iframePreSelection.onResize = watchFunction(onResize.bind(iframePreSelection));
+
+  let iframe = exports.iframe = {
+    currentIframe: iframePreSelection,
+    display: function (installHandlerOnDocument, standardOverlayCallbacks) {
+      return iframeSelection.display(installHandlerOnDocument)
+        .then(() => iframePreSelection.display(installHandlerOnDocument, standardOverlayCallbacks))
```
ianb: `;`

addon/webextension/selector/uicontrol.js
```diff
@@ -636,8 +636,7 @@ window.uicontrol = (function () {
         stateHandlers.resizing.startResize(event, "move");
       } else if (! ui.Box.isControl(target)) {
         mousedownPos = new Pos(event.pageX, event.pageY);
-        mouseupNoAutoselect = true;
-        setState("draggingReady");
+        setState("crosshairs");
```
ianb: I suppose we should rename crosshairs as that mode hasn't displayed crosshairs for a long time.  But not really related to this PR.

addon/webextension/selector/ui.js
dannycoates: The hoverbox remains visible during the scroll without this. I think it looks bad, but if that's not an issue I'll remove it.

ianb: Ah, I didn't catch that part.  That makes sense then.

PR #2463, opened 2017-03-22
Firefox review: full source

ianb:
This is a PR intended to cover the entirety of the source.  This PR is not intended to be merged, but only to act as a holding place for comments.


ianb:
We're going to try a different approach for this.

PR #2469, opened 2017-03-23
Exclude some pages from the screenshot context menu item

dannycoates:
This partially addresses @SoftVision-CosminMuntean's comment https://github.com/mozilla-services/screenshots/pull/2368#issuecomment-288721037

It'll prevent the context menu item from being added to excluded pages except the screenshots site itself by way of the default schemes in [match patterns](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Match_patterns).

### DIFF
--- a/addon/webextension/background/main.js
+++ b/addon/webextension/background/main.js
@@ -63,7 +63,8 @@ window.main = (function () {
   browser.contextMenus.create({
     id: "create-screenshot",
     title: browser.i18n.getMessage("contextMenuLabel"),
-    contexts: ["page"]
+    contexts: ["page"],
+    documentUrlPatterns: ["<all_urls>"]
   }, () => {
     // Note: unlike most browser.* functions this one does not return a promise
     if (browser.runtime.lastError) {


SoftVision-CosminMuntean:
The "Take a Screenshot" option from context menu is no longer displayed on "about:*" pages.

But, I have noticed that on Screenshots pages (eg: "My Shots" page or on a saved shot preview page) the Firefox Screenshots button from browser toolbar is inactive, but the option is still available in context menu.
@dannycoates Shouldn't the option from context menu also be removed from Firefox Screenshots pages?

SoftVision-CosminMuntean:
I just noticed that the"Take a Screenshot" option is also available on context menu if you right click on a tab, even if is not focused.

If you right click on a tab which is not focused, and choose the "Take a Screenshot" option from context menu, the overlay is displayed on the focused tab and the Screenshots icon is changed in the green color on the unfocused tab. Here is a screen recording: https://goo.gl/8HDAun.
Should I log a different issue for this?

dannycoates:
> Shouldn't the option from context menu also be removed from Firefox Screenshots pages?

It should but it doesn't because it's difficult to do so in a low impact way.

dannycoates:
> I just noticed that the"Take a Screenshot" option is also available on context menu if you right click on a tab, even if is not focused.

Good find. New bug #2483

Issue #2261, opened 2017-03-02
Change uses of XMLHttpRequest to fetch


ianb:
The [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) is the more modern alternative to using XMLHttpRequest.  We still have some uses of XMLHttpRequest in `addon/webextension/` and these could be replaced.  (We have some fetch examples too that you can use for reference.)

pgadige:
I'd like to work on this issue. Could please assign it to me. Thank you.

ianb:
@pgadige : I can't seem to assign it to you, but I'll assign it to myself so someone else doesn't take it (but I'll leave it to you to make the patch!)

pgadige:
@ianb thank you very much for holding it up. Could you please try to assign it to me again, now?

ianb:
Seems like there isn't activity on this, resetting assignment

ianb:
I started this in the branch [xmlhttprequest-to-fetch](https://github.com/mozilla-services/screenshots/tree/xmlhttprequest-to-fetch), but after doing this I'm feeling uncomfortable with the amount of logic being moved around at this late date.  I don't think this work should be done until after our first release.

ianb:
This issue is being closed (without being fixed) because [we are discontinuing the Screenshots server](https://blog.mozilla.org/futurereleases/2019/01/24/clarifying-the-future-of-firefox-screenshots/).

PR #2471, opened 2017-03-23
Add all the add-on files for review

ianb:
This is a PR specifically intended to review the entirety of the add-on code.  This is not intended to be merged, only to collect comments


addon/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+function shutdown(data, reason) { // eslint-disable-line no-unused-vars
+  prefObserver.unregister();
+}
+
+function install(data, reason) {} // eslint-disable-line no-unused-vars
+
+function uninstall(data, reason) {} // eslint-disable-line no-unused-vars
+
+function getBoolPref(pref) {
+  return prefs.getPrefType(pref) && prefs.getBoolPref(pref);
```
Mossop: You can now pass a second argument to getBoolPref to specify a default if the pref doesn't exist.

addon/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+function shouldDisable() {
+  return getBoolPref(USER_DISABLE_PREF) || getBoolPref(SYSTEM_DISABLE_PREF);
+}
+
+function handleStartup() {
+  AddonManager.getAddonByID(ADDON_ID).then((addon) => {
```
Mossop: Seems like you mostly do this to get the resource URI for the add-on? That is passed on the data object to startup already so you could just pass it through to here and not need this getAddonByID call which can cause a startup performance hit.

addon/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+    console.error(err);
+    if (err.message !== "This embedded extension has already been started") {
+      // TODO: Should we send these errors to Sentry? #2420
+    }
+  });
+}
+
+function stop(webExtension) {
+  webExtension.shutdown().then(() => {
+    EmbeddedExtensionManager.untrackEmbeddedExtension(webExtension);
```
Mossop: Looks like the shutdown call does this already.

addon/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+  });
+}
+
+function handleMessage(msg, sender, sendReply) {
+  if (!msg) {
+    return;
+  }
+
+  if (msg.funcName === "getTelemetryPref") {
+    let enableTelemetry = getBoolPref(TELEMETRY_PREF);
```
Mossop: This had me blinking for a moment as I thought you were turning telemetry on. Can you rename this variable to telemetryEnabled? I think that makes more sense.

addon/bootstrap.js
```diff
@@ -0,0 +1,118 @@
...
+const TELEMETRY_PREF = "toolkit.telemetry.enabled";
+const PREF_BRANCH = "extensions.screenshots.";
+const USER_DISABLE_PREF = "extensions.screenshots.disabled";
+const SYSTEM_DISABLE_PREF = "extensions.screenshots.system-disabled";
+
+const { interfaces: Ci, utils: Cu } = Components;
+Cu.import("resource://gre/modules/AddonManager.jsm");
+Cu.import("resource://gre/modules/Console.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+const { EmbeddedExtensionManager } = Cu.import("resource://gre/modules/LegacyExtensionsUtils.jsm");
```
Mossop: Can you just use the exposed LegacyExtensionsUtils instead since you don't need the untrack call below?

6a68:
@Mossop Thanks for the feedback. I'll get a PR out today that addresses the points you mentioned above.

addon/bootstrap.js
ianb: I was hoping we could keep Firefox 52 compatibility (for ESR), and @Standard8 mentioned that this was a 54 API change... except, if we distribute something for Firefox 52 is might be webextension/ only and then compatibility of bootstrap.js won't matter.

Mossop: That's fine then. You might also consider using Preferences.jsm which gives you the ability to do this even on 52 and earlier.

6a68:
I've landed the requested bootstrap.js changes in master, and cherry-picked that commit into this branch.

6a68:
Question for reviewers: It looks like `window.dump()` is available to WebExtensions. Should we switch to that instead of console.* throughout?

addon/webextension/background/main.js
```diff
@@ -0,0 +1,150 @@
...
+    backend = newBackend;
+    backend = backend.replace(/\/*$/, "");
+  };
+
+  exports.getBackend = function () {
+    return backend;
+  };
+
+  for (let permission of manifest.permissions) {
+    if (permission.search(/^https?:\/\//i) != -1) {
```
kmaglione: Nit: /^https?:\/\//.test(permission)`, please

addon/webextension/background/main.js
```diff
@@ -0,0 +1,150 @@
...
+  }));
+
+  browser.contextMenus.create({
+    id: "create-screenshot",
+    title: browser.i18n.getMessage("contextMenuLabel"),
+    contexts: ["page"]
+  }, () => {
+    // Note: unlike most browser.* functions this one does not return a promise
+    if (browser.runtime.lastError) {
+      catcher.unhandled(new Error(browser.runtime.lastError.message));
```
kmaglione: This shouldn't be necessary. This call is guaranteed to succeed in this case.

addon/webextension/background/main.js
```diff
@@ -0,0 +1,150 @@
...
+      return false;
+    }
+    return true;
+  }
+
+
+  browser.tabs.onUpdated.addListener(catcher.watchFunction((id, info, tab) => {
+    if (info.url && tab.selected) {
+      if (urlEnabled(info.url)) {
+        browser.browserAction.enable(tab.id);
```
kmaglione: We should probably use a pageAction for this, and call `show()` rather than `enable()`. We wouldn't have to worry about disabling in that case.

addon/webextension/background/main.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+    if (url.startsWith(backend) || /^(?:about|data|moz-extension):/i.test(url)) {
+      return false;
+    }
+    return true;
+  }
+
+
+  browser.tabs.onUpdated.addListener(catcher.watchFunction((id, info, tab) => {
+    if (info.url && tab.selected) {
```
kmaglione: This won't handle switching between tabs with different expected enabled states.

kmaglione: Also, `selected` isn't what you want here. You want `active`. Although there's not actually a distinction between the two in Firefox.

addon/webextension/background/selectorLoader.js
```diff
@@ -0,0 +1,83 @@
...
+    "selector/documentMetadata.js",
+    "selector/util.js",
+    "selector/ui.js",
+    "selector/shooter.js",
+    "selector/uicontrol.js"
+  ];
+
+  exports.unloadIfLoaded = function () {
+    return browser.tabs.executeScript({
+      code: "window.selectorLoader && window.selectorLoader.unloadModules()"
```
kmaglione: Can we just send a message to the tab instead? If the script is already loaded, it can add a message listener easily enough.

addon/webextension/background/selectorLoader.js
```diff
@@ -0,0 +1,83 @@
...
+    }).then(result => {
+      return result && result.toString() === "true";
```
kmaglione: Stringifying this is a bit weird...

addon/webextension/background/selectorLoader.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+  };
+
+  exports.loadModules = function () {
+    let lastPromise = Promise.resolve(null);
+    scripts.forEach((file) => {
+      lastPromise = lastPromise.then(() => {
```
kmaglione: An async function would make this much simpler. But passing an array for `file` rather than loading the scripts sequentially would make it much faster.

addon/webextension/background/selectorLoader.js
```diff
@@ -0,0 +1,83 @@
...
+        return browser.tabs.executeScript({file})
+          .catch((error) => {
+            console.error("error in script:", file, error);
+            error.scriptName = file;
+            throw error;
+          })
+      })
+    });
+    return lastPromise.then(() => {
+      console.log("finished loading scripts:", scripts.join(" "), "->", browser.runtime.lastError || "no error");
```
kmaglione: `lastError` is not used for promise handlers.

addon/webextension/background/takeshot.js
```diff
@@ -0,0 +1,123 @@
...
+              x: selectedPos.right - selectedPos.left,
+              y: selectedPos.bottom - selectedPos.top
+            }
+          }
+        });
+      });
+    }
+    let shotAbTests = {};
+    let abTests = auth.getAbTests();
+    for (let testName in abTests) {
```
kmaglione: for-of please.

addon/webextension/domainFromUrl.js
```diff
@@ -0,0 +1,30 @@
+/** Returns the domain of a URL, but safely and in ASCII; URLs without domains
+    (such as about:blank) return the scheme, Unicode domains get stripped down
+    to ASCII */
+window.domainFromUrl = (function () {
+
+  return function urlDomainForId(location) { // eslint-disable-line no-unused-vars
+    let domain = location.hostname;
+    if (domain) {
+      if (domain.indexOf(":") !== -1) {
+        domain = domain.replace(/:.*/, "");
```
kmaglione: `hostname` should never include a port part, so this isn't necessary.

addon/webextension/domainFromUrl.js
```diff
@@ -0,0 +1,30 @@
...
+      }
+    } else {
+      domain = location.href.split(":")[0];
```
kmaglione: Should probably be `location.origin`

addon/webextension/selector/ui.js
```diff
@@ -0,0 +1,627 @@
...
+      timer: null,
+      windowDelayer: null,
+      lastHeight: null,
+      lastWidth: null
+    },
+    document: null,
+    display: function (installHandlerOnDocument) {
+      return new Promise((resolve, reject) => {
+        if (! this.element) {
+          this.element = document.createElement("iframe");
```
kmaglione: The contents of this iframe will be accessible to web content, which worries me. Can we load a web-accessible extension URL into them?

addon/webextension/sitehelper.js
```diff
@@ -0,0 +1,41 @@
+/* globals catcher, callBackground */
+/** This is a content script added to all screenshots.firefox.com pages, and allows the site to
+    communicate with the add-on */
+
+window.sitehelper = (function () {
```
kmaglione: I'd rather we use export helpers (`exportFunction`, `cloneInto`, ...) than custom events for this.

addon/webextension/sitehelper.js
```diff
@@ -0,0 +1,41 @@
...
+  catcher.registerHandler((errorObj) => {
+    callBackground("reportError", errorObj);
+  });
+
+
+  function sendCustomEvent(name, detail) {
+    if (typeof detail == "object") {
+      // Note sending an object can lead to security problems, while a string
+      // is safe to transfer:
+      detail = JSON.stringify(detail);
```
kmaglione: You would be better off with sendMessage than custom events, since messages are structured cloned by default. However, if you need to use custom events, you can use `cloneInto` to make the object accessible to content.

shared/shot.js
```diff
@@ -0,0 +1,726 @@
+// Note: in this library we can't use any "system" dependencies because this can be used from multiple
+// environments
+/* globals console */
+
+/** Throws an error if the condition isn't true.  Any extra arguments after the condition
+    are used as console.error() arguments. */
+function assert(condition) {
+  if (! condition) {
+    console.error.apply(console, ["Failed assertion:"].concat(Array.prototype.slice.call(arguments, 1)));
```
kmaglione: `console.error("Failed assertion", ...args)` along with `...args` as a second argument to `assert`

shared/shot.js
```diff
@@ -0,0 +1,726 @@
...
+
+  }
+
+  /** Returns a JSON version of this shot */
+  asJson() {
+    let result = {};
+    for (let attr of this.REGULAR_ATTRS) {
+      var val = this[attr];
+      if (val && val.asJson) {
+        val = val.asJson();
```
kmaglione: The `toJSON` method is the standard way of doing this, and is automatically called by `JSON.stringify`

6a68:
@kmaglione Thanks for the review! I'll get started addressing your feedback in the `webextension-review-changes` branch and ping you with any questions.

addon/webextension/selector/ui.js
ianb: We took this approach from uBlock Origin's element picker.  In general in my experiments it seemed like the content worker has no special ability to look into iframes, though I never specifically tried a moz-extension URL for the iframe; is that different?

shared/shot.js
```diff
@@ -0,0 +1,726 @@
...
+  if ((/^data:.*$/i).test(url)) {
+    return true;
+  }
+  if ((/^chrome:.*/i).test(url)) {
+    return true;
+  }
+  if ((/^view-source:/i).test(url)) {
+    return isUrl(url.substr("view-source:".length));
+  }
+  return (/^https?:\/\/[a-z0-9\.\-]+[a-z0-9](:[0-9]+)?\/?/i).test(url);
```
ianb: This same code runs in Node, which doesn't have `URL`.  A quick google gives me this library: https://www.npmjs.com/package/dom-urls – but the package doesn't give me a lot of confidence, especially around the corner cases of potentially invalid URLs, and the functionality is fairly small.

shared/shot.js
```diff
@@ -0,0 +1,726 @@
...
+    if (! seen[attr]) {
+      if (! deepEqual(a[attr], b[attr])) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
+function makeUuid() {
```
ianb: I should probably rename this function, as it's really just a unique slug and not a UUID – in this particular instance it's just used to make unique IDs for multiple clips in a single shot.  To avoid collision we use random IDs in an object instead of putting them in an array.

addon/webextension/background/analytics.js
```diff
@@ -0,0 +1,74 @@
...
+  let telemetryPref;
+
+  exports.sendEvent = function (action, label, options) {
+    let eventCategory = "addon";
+    if (! telemetryPrefKnown) {
+      console.warn("sendEvent called before we were able to refresh");
+      return Promise.resolve();
+    }
+    if (! telemetryPref) {
+      console.info(`Cancelled sendEvent ${eventCategory}/${action}/${label || 'none'} ${JSON.stringify(options)}`);
```
ianb: What's a good way to handle a pref in a webextension?  In the SDK version we used about:config prefs for some things like this, but there doesn't seem to be a webextension equivalent, so either I'd need a whole options page (which would be bad since these aren't options that are meaningful to users), or build the pref into the source itself.  Or hopefully there's a better way I'm not aware of...?

addon/webextension/background/analytics.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+    if (typeof label == "object" && (! options)) {
+      options = label;
+      label = undefined;
+    }
+    options = options || {};
+    let di = deviceInfo();
+    return new Promise((resolve, reject) => {
+      let url = main.getBackend() + "/event";
+      let req = new XMLHttpRequest();
```
ianb: https://github.com/mozilla-services/screenshots/issues/2261 – we on the team have avoided taking good first bugs while Outreachy applications were incoming, but now that the application period is over we can probably take these ourselves.

addon/webextension/catcher.js
```diff
@@ -0,0 +1,74 @@
...
+      result = exc;
+    } else {
+      result = {
+        fromMakeError: true,
+        name: exc.name || "ERROR",
+        message: exc+"",
+        stack: exc.stack
+      };
+      for (let attr in exc) {
+        result[attr] = exc[attr];
```
ianb: I'm relying on the serialization actually doing that filtering, since we'd have to do a deep inspection to really be sure.  As far as I know anything that causes a problem with serialization will simply be omitted?

addon/webextension/catcher.js
```diff
@@ -0,0 +1,74 @@
...
+        exports.unhandled(e);
+        throw e;
+      }
+      return result;
+    };
+  };
+
+  exports.watchPromise = function watchPromise(promise) {
+    return promise.catch((e) => {
+      console.error("------Error in promise:", e+"");
```
ianb: I think this was an SDK-ism, as at times the SDK would show console.log(exceptionObject) as `{}`

addon/webextension/background/communication.js
```diff
@@ -0,0 +1,67 @@
...
+      will *not* fail, and will return a value of exports.NO_BOOTSTRAP  */
+  exports.sendToBootstrap = function (funcName, ...args) {
+    return browser.runtime.sendMessage({funcName, args}).then((result) => {
+      if (result.type === "success") {
+        return result.value;
+      } else {
+        throw new Error(`Error in ${funcName}: ${result.name || 'unknown'}`);
+      }
+    }, (error) => {
+      if (error && error.message === "Could not establish connection. Receiving end does not exist.") {
```
ianb: Is there another approach we could use?  I'd like to maintain the ability to run with or without bootstrap.js, but that means sending messages that might be rejected in this way.

addon/webextension/sitehelper.js
ianb: I created #2573 for this – the refactoring worries me right now, so I'd like to save that work for later.

addon/webextension/background/selectorLoader.js
ianb: The [docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/executeScript) don't seem to mention the ability to pass in a list.  I think the initial version of this used `chrome.tabs.executeScript()` which didn't allow a list when I was experimenting with it.

addon/webextension/background/selectorLoader.js
ianb: @dannycoates : do you remember why you did this?  I didn't notice it at the time.

addon/webextension/background/auth.js
```diff
@@ -0,0 +1,208 @@
...
+        secret: registrationInfo.secret,
+        deviceInfo: JSON.stringify(deviceInfo())
+      }));
+    });
+  }
+
+  function login(options) {
+    let { ownershipCheck, noRegister } = options || {};
+    return new Promise((resolve, reject) => {
+      let loginUrl = main.getBackend() + "/api/login";
```
ianb: We have a plan in the future to associate devices with identities, but this per-device automatic identity let's us onboard people with no gateways.  Probably the identity association will be done entirely through the website, and the add-on will only ever know about its own identity.

Though there was an idea of using sync, and syncing all these per-device identities across all devices.  The unpairing gets a little complicated in that case, as does keeping the server in sync with the add-on.

addon/webextension/background/selectorLoader.js
dannycoates: It was the simplest thing I could get to work. I couldn't figure out what `result` actually was. `typeof result` is `object` and `result == true` is `false`. I notice `console.log(result)` was the string `"true"` so I went with that. I rather do something less weird too.

addon/webextension/background/selectorLoader.js
dannycoates: We need to be able to get a result whether the script is loaded or not.

addon/webextension/background/selectorLoader.js
ianb: `new Boolean(true)` ? That's == true, but not === true

dannycoates: did you mean `new Boolean(result)`? which is **always** == true here. :man_shrugging:

addon/webextension/selector/callBackground.js
```diff
@@ -0,0 +1,17 @@
+/* globals browser */
```
6a68: @Standard8 are you working on an .eslintrc for the webextension?

shared/shot.js
6a68: @ianb What about using https://www.npmjs.com/package/whatwg-url as a polyfill on the server? Then we could use the URL constructor, keep zero dependencies in the addon, and the server should work just fine. Maybe this could be filed as a separate issue.

addon/webextension/background/analytics.js
6a68: Seems like [`window.dump()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/dump) would be a good alternative to pref-setting here.

addon/webextension/background/auth.js
```diff
@@ -0,0 +1,208 @@
...
+  catcher.watchPromise(browser.storage.local.get(["registrationInfo", "abTests"]).then((result) => {
+    if (result.abTests) {
+      abTests = result.abTests;
+    }
+    if (result.registrationInfo) {
+      registrationInfo = result.registrationInfo;
+    } else {
+      registrationInfo = generateRegistrationInfo();
+      console.info("Generating new device authentication ID", registrationInfo);
+      return browser.storage.local.set({registrationInfo});
```
6a68: This particular promise chain only runs when auth.js is initially loaded. I don't think we would ever encounter a race condition here.

addon/webextension/background/main.js
6a68: Wouldn't a pageAction insert an icon in the urlbar, not the toolbar?

addon/webextension/background/main.js
6a68: > This won't handle switching between tabs with different expected enabled states.

Cool, I guess adding a separate `onActivated` listener would be the way to go, where we disable the button if the tab doesn't (yet) have a URL, or if `urlEnabled` returns false.

addon/webextension/background/selectorLoader.js
6a68: `Boolean(result)`, omitting the `new`, actually === true

addon/webextension/background/auth.js
```diff
@@ -0,0 +1,208 @@
...
+      req.send(uriEncode({
+        deviceId: registrationInfo.deviceId,
+        secret: registrationInfo.secret,
+        deviceInfo: JSON.stringify(deviceInfo()),
+        ownershipCheck
+      }));
+    });
+  }
+
+  function saveAuthInfo(responseJson) {
```
ianb: So... here's how I think about what we're doing here:

1. The real canonical location for this information is in memory in the background page.  We initialize the values eagerly (at the module level), and update them in-place while also recording updates in storage.  
2. We're relying on a simple dumb store (i.e., dropping in `storage.sync` would break some assumptions).  That is, nothing ever changes behind our back.
3. `registrationInfo` is a compound object, so there's some danger we're overwriting something that we meant to keep – except for the fact that its canonical location is in memory.
4. `abTests` and `registrationInfo` aren't really related.  Really `abTests` is canonically stored on the server, but we can't fetch it eagerly, so the storage here is more like a cache.
5. The real concern I see is that we overwrite registrationInfo when it has already been used.  I think we're pretty safe here, assuming that `storage.local` is consistent and reliable.  We only set `registrationInfo` at module-load time, so there's no real opportunity to overwrite it so long as `storage.local` is reliable.  *Except...*
6. We also set the info in `setDeviceInfoFromOldAddon()`.  Or rather we do it wrong, ugh.  Filed as #2588 –
 anyway, that also happens at module load time based on communication with bootstrap.js

addon/webextension/background/selectorLoader.js
6a68: Ah, yeah. Looking at the [docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/executeScript#Return_value) for executeScript, the returned value is what's returned by `unloadModules`...which should be the boolean `true`, not the string "true", but maybe that's a side effect of Promisification

dannycoates: I think we've established that `result` is sufficiently weird here. I propose we just keep it like this since it does what we want.

addon/webextension/background/main.js
6a68: Interestingly, I've found just one case where `onActivated` correctly toggles the button state, but `onUpdated` doesn't: when a tab with a non-shootable URL is torn off a window.

addon/webextension/selector/callBackground.js
ianb: I believe @Standard8 was still working through some install issues with the Firefox standard eslint package, but that got shelved for now since there are higher priority things.

I'd like to adopt the Firefox rules, and fix up these declarations along the way, but it seems easiest and least intrusive to do it in one big batch.

addon/webextension/background/analytics.js
```diff
@@ -0,0 +1,74 @@
+/* globals main, auth, catcher, deviceInfo, communication */
+
+window.analytics = (function () {
```
ianb: François Marier is reviewing this now

addon/webextension/background/analytics.js
ianb: Do we want Screenshots messages to show up for someone who isn't developing Screenshots?

addon/webextension/background/main.js
ianb: I'll ask @johngruen about this, but pageAction leads to a UX change and I'd want to defer to him on that.

addon/webextension/selector/callBackground.js
Standard8: It'll likely be just after we land screenshots in the tree, I'll be able to get back to fixing up the eslint package & getting the rules applied.

addon/webextension/selector/ui.js
ianb: I opened #2605 to investigate this.  I'm not confident it's going to work, though it would be a security improvement if we could do it.

addon/webextension/background/auth.js
```diff
@@ -0,0 +1,208 @@
...
+    return new Promise((resolve, reject) => {
+      let registerUrl = main.getBackend() + "/api/register";
+      let req = new XMLHttpRequest();
+      req.open("POST", registerUrl);
+      req.setRequestHeader("content-type", "application/x-www-form-urlencoded");
+      req.onload = catcher.watchFunction(() => {
+        if (req.status == 200) {
+          console.info("Registered login");
+          initialized = true;
+          saveAuthInfo(JSON.parse(req.responseText));
```
6a68: filed as #2261

addon/webextension/background/auth.js
```diff
@@ -0,0 +1,208 @@
...
+          resolve(true);
+          analytics.sendEvent("registered");
+        } else {
+          console.warn("Error in response:", req.responseText);
+          reject(new Error("Bad response: " + req.status));
+        }
+      });
+      req.send(uriEncode({
```
6a68: filed as #2261

addon/webextension/background/auth.js
```diff
@@ -0,0 +1,208 @@
...
+        }
+      });
+      req.onerror = catcher.watchFunction(() => {
+        let exc = new Error("Connection failed");
+        exc.url = loginUrl;
+        exc.popupMessage = "CONNECTION_ERROR";
+        reject(exc);
+      });
+      req.setRequestHeader("content-type", "application/x-www-form-urlencoded");
+      req.send(uriEncode({
```
6a68: also filed as #2261

addon/webextension/background/selectorLoader.js
kmaglione: If the script isn't loaded, you'll get an error that there is no receiver.

addon/webextension/selector/ui.js
kmaglione: If the iframe has a moz-extension: URL, your content script will be able to access its contents, since it has an expanded principal with a moz-extension: URL, but the content page won't.

addon/webextension/background/selectorLoader.js
kmaglione: Result is an array of results from each frame that the script was executed in, and [true] stringifies to "true".

addon/webextension/background/selectorLoader.js
kmaglione: Hm. You're right, we should change that to accept an array like `content_script` declarations do. I thought it already did.

addon/webextension/background/main.js
kmaglione: I'll obviously leave UX decisions to someone more qualified, but we generally use pageActions for things that are only available on certain URLs. And the API is designed to automatically track which specific pages the action is active for, and automatically reset on navigation, so it would make this code much simpler.

addon/webextension/catcher.js
kmaglione: For the most part, yes, except that cyclic object values will cause an error, and some objects lead to weird/confusing/potentially unsafe serialization.

shared/shot.js
ianb: We'll push this off for now.  Followups in #2660 and #2659

addon/webextension/background/analytics.js
ianb: We'll do #2661 before landing, and #2261 later

addon/webextension/background/main.js
ianb: Followup in #2662

addon/webextension/sitehelper.js
ianb: Let's switch to cloneInto and leave postMessage for now

ianb:
I'm going to close this PR – we've discussed all the comments, fixed a bunch, made followup bugs for all the rest.  Thanks @kmaglione and @Mossop !

PR #2473, opened 2017-03-23
ensure auth when visiting /shots from addon

### LINKED ISSUE #2415: Ensure authentication before navigating to My Shots
In #2414 we want to lazily log you in if you visit My Shots, but we have two patterns that get you to My Shots from the add-on, and we can ensure a login before visiting.

The first case is the My Shots button on the selection screen.  The second is hitting the Page Shot button from about:newtab.  Calling `auth.authHeaders().then(...)` in these cases will ensure the request is authenticated.

#2414 will still allow a user to bookmark My Shots or revisit it from the awesomebar and be authenticated.

dannycoates:
fixes #2415 

IDK if we want to add the cookies permission, but if we do this works.

### DIFF
--- a/addon/webextension/background/main.js
+++ b/addon/webextension/background/main.js
@@ -49,7 +49,9 @@ window.main = (function () {
       catcher.watchPromise(analytics.refreshTelemetryPref().then(() => {
         sendEvent("goto-myshots", "about-newtab");
       }));
-      catcher.watchPromise(browser.tabs.update({url: backend + "/shots"}));
+      catcher.watchPromise(
+        auth.authHeaders()
+        .then(() => browser.tabs.update({url: backend + "/shots"})));
     } else {
       catcher.watchPromise(
         toggleSelector(tab)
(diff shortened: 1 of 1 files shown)


addon/webextension/background/auth.js
```diff
@@ -157,6 +157,19 @@ window.auth = (function () {
     });
   };
 
+  exports.ensureLoggedIn = function () {
+    return browser.cookies.get({
```
ianb: I don't think we need to worry about the cookies specifically, if we call `auth.authHeaders()` it should cover us.

dannycoates: it doesn't. clear your cookies and try it

ianb: they both just call login(), so don't they have to act the same? – it's true it can't handle explicitly clearing cookies after login occurred, but that seems like a fairly obscure case, and the lazy login in the other PR should handle that situation anyway.  Unless there's something I'm missing?

dannycoates:
updated

PR #2474, opened 2017-03-23
Fix #2414, lazily check auth state on My Shots page

### LINKED ISSUE #2414: Add wantsauth handling to My Shots
In #2413 there is handling for authenticating requests after page load.  But that ticket only works for individual shot pages, so if you haven't saved from Page Shot in a session and you visit My Shots you will appear unauthenticated.

To fix this we need to wire My Shots authentication failures to `wantsauth`.  Right now My Shots redirects, but I would propose we change it to a 404.  Then the 404 page should include `wantsauth` and try to request authentication.  If it does receive authentication then probably we could just reload the page, since the cookie is now set.

6a68:


server/src/pages/shotindex/controller.js
```diff
@@ -2,16 +2,34 @@ const sendEvent = require("../../browser-send-event.js");
...
+    model = m;
+    render();
+    return;
+  }
+  if (window.wantsauth) {
+    if (window.wantsauth.getAuthData()) {
+      location.reload();
+    } else {
+      let authTimeout = setTimeout(() => {
+        location.pathname = "";
```
6a68: I dunno if we will still want to redirect to `/?extension-not-installed` on the production site, but if so, I can add a `location.search = ` bit here.

server/src/pages/shotindex/view.js
```diff
@@ -68,6 +71,15 @@ class Body extends React.Component {
     );
   }
 
+  renderNoDeviceId() {
+    return (
+      <div className="large-icon-message-container" key="no-shots-found">
+        <div className="large-icon logo-no-shots" />
```
6a68: Would be nice to have some kind of a loading GIF for this step, maybe?

server/src/pages/shotindex/controller.js
ianb: Eh... this is a weird case anyway, we don't know why someone goes to /shots without the extension installed.

Though maybe we want to consider someone going there before they've used Screenshots ever, and thus they aren't registered?  That's not so hard to do, you could just hit the Screenshots button from about:newtab.

server/src/pages/shotindex/view.js
ianb: `/creating` has one that's in CSS, so it'd probably be easy to use.  Could be followup Stretch bug.

server/src/pages/shotindex/view.js
```diff
@@ -10,8 +10,9 @@ class Head extends React.Component {
   render() {
     return (
       <reactruntime.HeadTemplate {...this.props}>
-        <script src={this.props.staticLink("/static/js/shotindex-bundle.js")} async></script>
-        <link rel="stylesheet" href={this.props.staticLink("/static/css/shot-index.css")} />
+        <script src={ this.props.staticLink("/static/js/wantsauth.js") } />
```
ianb: This should be conditional on `this.props.deviceId`, like:

    { this.props.deviceId ? null : <script src={ this.props.staticLink("/static/js/wantsauth.js") } /> }

i.e., it's kind of aggressive, and you only wantauth if you don't haveauth.

server/src/pages/shotindex/server.js
```diff
@@ -8,16 +8,20 @@ exports.app = app;
...
   }).catch((err) => {
     res.type("txt").status(500).send("Error rendering page: " + err);
     console.error("Error rendering page:", err);
   });
+
+  function _render(shots) {
+    req.shots = shots || [];
+    const page = require("./page").page;
+    reactrender.render(req, res, page);
+  }
```
ianb: I'd probably have done this like:

```js
let getShots;
if (req.deviceId) {
  getShots = Shot.getShotsForDevice(req.backend, req.deviceId, query);
} else {
  getShots = Promise.resolve([]);
} 
getShots.then((shots) => { /* same as before */ });
```

but whichever is fine

6a68: oh, that's nice. I'll switch to that approach

server/src/pages/shotindex/controller.js
6a68: Yeah, I wonder if another pref for "user has never registered" might be in order. maybe `extensions.screenshots.new-user` or `first-run` or something? I can file a needs:product bug to smooth that pathway into the product.

6a68: Hmm, actually, in that case I think the 10 second timeout _should_ bump a totally new user to the screenshots homepage. The homepage can walk them through onboarding.

ianb: I think you are right about bumping the user to the homepage.  (We do keep track of whether you are registered in the add-on, but don't need to pay attention to that if we're bumping)

6a68:
@ianb Updated. I animated the 'go forth and take shots' icon during the loading step, what do you think?

6a68:
I think I've addressed all the concerns, merging

PR #2487, opened 2017-03-27
Focus pre-selection iframe when it is displayed

ianb:
Add tabindex to the buttons on that page


ianb:
Tab works, but enter still does not!

Issue #2270, opened 2017-03-03
After removing all Page Shot data, the save functionality no longer works 


SoftVision-CosminMuntean:
[Affected versions]:
- Nightly 54.0a1 (Build ID: 20170303030202)	
- Page Shot 5.2.6242

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" is set to "false" and latest Page Shot (5.2.6242) custom build is installed on a new clean profile.
- Have multiple saved shots on the created profile.

[Steps to reproduce]:
1. Open Firefox with the profile from prerequisites.
2. Go to "My Shots" page (https://pageshot.dev.mozaws.net/shots).
2. From the footer of the page click the "Remove All Data" link.
3. Click the red "Proceed" button.
4. Navigate to any website, performed a selection and save it.
5. Observe the browser behavior.

[Expected result]:
- The selection is successfully saved and you are redirected to the saved shot preview page

[Actual result]:
- After clicking the "Save" button the Page Shot overlay is dismissed and a "Error: Error: response failed" is displayed.

[Addition Notes:]
- If the browser is restarted the issue is no longer reproducible.
- Here is the screen recording with the issue: https://goo.gl/KC8I7H
- Here is a screen shot with browser console errors:

![after removing all page shot data](https://cloud.githubusercontent.com/assets/19226534/23559326/485824ae-003f-11e7-8782-57bc8af498f7.png)

ckprice:
Danny's fix deletes everything but leaves the device ID in place. Let's land Danny's fix, then field any data retention questions later.

ckprice:
Closed in #2498

PR #2498, opened 2017-03-27
delete data instead of device on delete-all

dannycoates:
A proposal for #2270

Instead of deleting the device just delete the data. (I'm not all that familiar with the server yet)

This approach doesn't require addon changes.


ianb:
I think this is okay.  I like how concise it is.  My one concern is that removing deviceId from the device itself is appealing because it dissociates the device from analytics data or any other data that might not be easily deleted from the database.

ckprice:
Fixes #2270

PR #2503, opened 2017-03-28
Selectorloader notification

ianb:


addon/webextension/selector/ui.js
```diff
@@ -286,7 +286,9 @@ window.ui = (function () { // eslint-disable-line no-unused-vars
     hide: function () {
       window.removeEventListener("scroll", this.onScroll, false);
       window.removeEventListener("resize", this.onResize, true);
-      this.element.style.display = "none";
+      if (this.element) {
```
dannycoates: When does this happen?

6a68: I'd guess this could happen if the overlay is toggled on and off quickly, so that it's torn down before the elements were rendered into the DOM

ianb: It specifically happens if there's an error during setup, so that the unloader is invoked before the ui is setup.  To test you have to introduce a top-level error into some module.

PR #2504, opened 2017-03-28
Exception fixups

ianb:
(Got a lot bigger than I expected)

### DIFF
--- a/addon/webextension/catcher.js
+++ b/addon/webextension/catcher.js
@@ -58,6 +58,7 @@ window.catcher = (function () {
       console.error("------Error in promise:", e+"");
       console.error(e.stack);
       exports.unhandled(makeError(e));
+      throw e;
     });
   };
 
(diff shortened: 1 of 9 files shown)


addon/webextension/catcher.js
```diff
@@ -58,6 +58,7 @@ window.catcher = (function () {
       console.error("------Error in promise:", e+"");
       console.error(e.stack);
       exports.unhandled(makeError(e));
+      throw e;
```
dannycoates: Can you explain the effect of this? I'm not 100% clear on what'll happen at the top level if we always re-throw. Just an error in the browser console?

ianb: Usually yes.  In the case I worked through here it causes the failure to go back to selector/shooter.js

It would probably be better if we were more explicit about expectations in each case but without doing that work I thought this would be a better default.

PR #2522, opened 2017-03-29
WIP remove user urls from sentry data

### LINKED ISSUE #2510: Implement filtering on Raven
We want to be extra sure that we don't submit the URL the user is visiting to Sentry.  For this purpose we should scan the data before submission to Sentry.

I propose we scan for:

1. The full URL
2. The full URL minus the hash
3. Origin + any path (something like `/{origin}[^\s>"]*/`)

In each case we would replace the value with some specific dummy value.  If 3 is the same as `backend` then we don't need to filter it out.

This can be implemented with [dataCallback](https://docs.sentry.io/clients/javascript/config/).  It will have to scan the entire submitted object for those strings, without attempting to understand the object.

dannycoates:
fixes #2510

### DIFF
--- a/addon/webextension/background/senderror.js
+++ b/addon/webextension/background/senderror.js
@@ -65,6 +65,11 @@ window.errorpopup = (function () {
     });
   };
 
+  function regexpEscape(str) {
+    // http://stackoverflow.com/questions/3115150/how-to-escape-regular-expression-special-characters-using-javascript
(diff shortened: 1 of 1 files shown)


dannycoates:
I almost think we might want `if (window.location.href.length > something) // forget sending to sentry`

ianb:
Where `something` is a very large number, to avoid weird DoSishness?  (4000?)

ianb:
(I'm not too worried about that, and looks good as is)

Issue #2533, opened 2017-03-30
Don't onboard every session in development


ianb:
Onboarding is going to load each time you start the browser during development.  (If you hot reload the onboarding will be skipped.)  This seems annoying, but I haven't figured out a way to suppress it as it's not easy to communicate with the add-on from the command-line.

ianb:
Goes with #2604 (as a build setting)

ianb:
This seems fine as it is.

Issue #2532, opened 2017-03-30
Onboarding Flow if user first clicks from any non-web or unsupported page


ianb:
Now it goes to My Shots, but that doesn't make much sense.

johngruen:
@ianb changed the name of this issue to be more expansive

### Current
If a user clicks the Screenshots icon from any prohibited page (about:x, frameset etc). We either show an error notification or redirect them to "my Shots"

### Expected
The user should be onboarded

### Proposal
Maybe we should pop a new tab on the screenshots home page and onboard users on top of that page, that seems like the easiest thing to do.

ianb:
2e5ce26005dbcd1094acc877f6c81c97aaf7c610 sends the user to `https://screenshots.firefox.com/#hello` and starts the screenshotting process (which also invokes onboarding) when the user tries to enable Screenshots on an unshootable page

Issue #2307, opened 2017-03-07
Implement onboarding


johngruen:
In order to implement we will need: 
- [ ] Full spec with measurements, button states, colors, etc
- [ ] All icons provided as SVG
- [ ] Rough WIP strings for the spec (I'll cover string review separately)
- [ ] full set of state diagrams explaining various flows into and out of onboarding.

There's a stretch from acceptance that indicates we might want a secondary method of triggering onboarding if a user doesn't engage w/in 3 weeks. It would be nice to see this in the flow diagram for the spec.

fangshih:
For the onboarding visual spec, most of the spec are ready. 
We will need to provide the illustration part later on with Sean's support. 

But here are the measurements  :
https://www.dropbox.com/sh/9u9040bwuohedf4/AABhzuzB1yX82hn1_tu04iKqa?dl=0

icon assets: 
https://www.dropbox.com/sh/skfyxqt1i0b782v/AAA-0r5-SSNVSL5EBOpxiVlla?dl=0

Onboarding screens: https://www.dropbox.com/sh/zkb6xrkkv9wusoz/AAAyLrGVDnKDB2odTPxayKEga?dl=0

Please let me know if you need anything else. Thanks :D

ckprice:
> Onboarding screens: https://www.dropbox.com/sh/zkb6xrkkv9wusoz/AAAyLrGVDnKDB2odTPxayKEga?dl=0

Clarifying question:

- screen 1 shows the browser just after Firefox install with the Screenshots icon "starred"
- screen 2 shows the FTU modal _after_ the user has clicked on the Screenshots icon (e.g. it doesn't automatically pop up), correct?

fangshih:
Hey @ckprice, That is correct! 
For the "starred" icon we may redesign the icon itself later on. But everything else is the same. : )

6a68:
what needs to happen to close this bug?

ianb:
I think we can call it closed – we have some followup bugs and @johngruen did a UI review

Issue #2529, opened 2017-03-30
After canceling shot, cannot re-activate


ianb:
To reproduce:

1. Start Screenshots
2. Cancel via Escape, or make a selection and cancel
3. Click the Screenshots button to re-activate

You'll notice an event `addon/cancel-shot/toolbar-button`, so I presume that the add-on believes that Screenshots is still active.  Reloading the page resets this behavior.

PR #2535, opened 2017-03-30
Start #2307, implement onboarding slideshow

ianb:
Not complete, but working.

This will open on every browser startup if you are using `./bin/run-addon`, since the flag for whether you have seen the onboarding is reset with each run.  Hot reloading doesn't reset the flag.

Followups: #2532 #2533 

There's code duplication between slides.js and ui.js, with a FIXME to acknowledge it, but I haven't made a ticket.  It feels safer to get this complete before refactoring these.

This leaves #2307 open in anticipation of full design.

### DIFF
--- a/addon/webextension/background/selectorLoader.js
+++ b/addon/webextension/background/selectorLoader.js
@@ -7,7 +7,8 @@ window.selectorLoader = (function () {
   const standardScripts = [
     "catcher.js",
     "background/selectorLoader.js",
-    "selector/callBackground.js"
+    "selector/callBackground.js",
+    "selector/util.js"
   ];
 
   const selectorScripts = [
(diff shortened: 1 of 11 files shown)


dannycoates:
I don't know the cause yet but here's a bug:

1. click screenshot button (opens onboarding)
2. click screenshot button again (closes onboarding)
3. click screenshot button yet again (nothing happens)

ianb:
I opened #2529 but I didn't switch back to master to check if it was a regression

addon/webextension/background/selectorLoader.js
```diff
@@ -28,7 +39,16 @@ window.selectorLoader = (function () {
     });
   };
 
-  exports.loadModules = function (tabId) {
+  exports.loadModules = function (tabId, hasSeenOnboarding) {
+    console.log("hasSeenOnboarding", hasSeenOnboarding);
```
dannycoates: nit

Issue #2466, opened 2017-03-23
Force a PNG content type when serving data


pauljt:
In addition to validating file input (issue #2301 ), we need to sanitize file outputs . The server should only ever be serving PNG images, right?  (i.e. we don't have any other content types?) Enforcing this content-type will prevent us serving unexpected content types which may help mitigate XSS style attacks.

PR #2547, opened 2017-03-31
assert data: urls have a png header

### LINKED ISSUE #2301: Enforce PNG check when posting images
Only images in PNG format must be permitted through the pageshot app when a user posts a new shot. The app must check the magic number of the image.

`PNG image files begin with an 8-byte signature which identifies the file as a PNG file and allows detection of common file transfer problems: \211 P N G \r \n \032 \n (89 50 4E 47 0D 0A 1A 0A).`

dannycoates:
fixes #2301

### DIFF
--- a/server/src/servershot.js
+++ b/server/src/servershot.js
@@ -9,6 +9,26 @@ const mozlog = require("mozlog")("servershot");
 
 const SEARCHABLE_VERSION = 1;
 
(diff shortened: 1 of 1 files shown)


ianb:
This might fix #2466 – except I don't think it quite does, I think that one also requires looking at the `/image` route.

Issue #2175, opened 2017-02-16
Do not show alt text while loading image


ianb:
Right now the alt text is pretty noticeable while the image is loading. This is distracting.

This is part of what was referred to in #2164

niharikak101:
Working on this :)

PR #2559, opened 2017-04-04
Fix/2175 (Work in Progress)

niharikak101:
This is a fix for #2175 

As of now it shows 'loading...' text while the image loads, but we could replace that with a spinner. 
Thoughts on the approach?


niharikak101:
Found some loaders at http://loading.io/ and added one as a placeholder!

dannycoates:
Hi @niharikak101, I think you're on the right track. Your `state`, render function, and spinner look good. 

I think we're going to need to do more than check the image in `componentDidMount` for this to work consistently. It works now sometimes (maybe every time when the server is on your computer), but the tricky thing is that the image *DOM element* being rendered (which triggers `componentDidMount`) and the image *data* being loaded don't always happen around the same time. We don't get another call to `componentDidMount` once the image data is loaded, so we need something else to change the `state`.

I think this post explains what we need to do pretty well: http://andrewhfarmer.com/detect-image-load/

niharikak101:
@dannycoates understood! :) Thank you!

dannycoates:
The code looks great! I'll defer to @johngruen on the spinner...

spinner

ianb:
I opened #2576 as a follow-up to maybe animate the spinner.

PR #2562, opened 2017-04-05
Ensure correct initialization

6a68:
Some Try jobs failed on the line that QI'd the URI to a FileURI.
Guarding against the preferences observer firing before startup fixes
those errors.

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -58,6 +58,11 @@ function shouldDisable() {
 }
 
 function handleStartup() {
+  // If the preference observer fires before startup completes, the URI won't
+  // be available yet, so just bail until startup has run.
+  if (!addonResourceURI) {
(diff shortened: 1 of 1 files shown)


addon/bootstrap.js
```diff
@@ -58,6 +58,11 @@ function shouldDisable() {
 }
 
 function handleStartup() {
+  // If the preference observer fires before startup completes, the URI won't
+  // be available yet, so just bail until startup has run.
+  if (!addonResourceURI) {
```
Standard8: This isn't going to work, the error we're getting is that `addonResourceURI.QueryInterface(Ci.nsIFileURL)` is throwing `NS_NOINTERFACE` - so it can't be converted to an `nsIFileURL`.

I'm starting to look at the options here.

Issue #2575, opened 2017-04-05
Delay starting WebExtension until Firefox has started


6a68:
Per discussion in the [Intent to Implement](https://groups.google.com/d/topic/firefox-dev/Bcmya7ZAVTM/discussion) email thread, we want to change the bootstrap.js startup routine to:

- lazy load all modules (replace `Cu.import` calls with `XPCOMUtils.defineLazyModuleGetter` calls)
- wait until the `sessionstore-windows-restored` nsIObserver event before starting the WebExtension

PR #2577, opened 2017-04-05
Delay WebExtension startup

6a68:
See #2575 for background

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -9,12 +9,21 @@ const USER_DISABLE_PREF = "extensions.screenshots.disabled";
 const SYSTEM_DISABLE_PREF = "extensions.screenshots.system-disabled";
 
 const { interfaces: Ci, utils: Cu } = Components;
-Cu.import("resource://gre/modules/AddonManager.jsm");
-Cu.import("resource://gre/modules/Console.jsm");
-Cu.import("resource://gre/modules/Services.jsm");
-Cu.import("resource://gre/modules/LegacyExtensionsUtils.jsm");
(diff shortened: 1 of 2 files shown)


6a68:
Here's a Try job to run the browser chrome tests with this patch. We'll see if anything breaks that was green. https://treeherder.mozilla.org/#/jobs?repo=try&revision=351a6a0120b4ea0b3ef2d621af676a9d6e8a7452

addon/bootstrap.js
```diff
@@ -30,15 +39,31 @@ const prefObserver = {
...
+  },
+
+  observe: function() {
+    appStartupDone();
+    this.unregister();
+  }
+}
+
 function startup(data, reason) { // eslint-disable-line no-unused-vars
+  appStartupObserver.register();
```
ianb: Are we confident that there's no way this code could be called after sessionstore-windows-restored is fired?  Like... if the add-on is reloaded?

6a68:
The first try job was on central, here's one on inbound:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=7f60640f1ce9617d9ec00167a7eda75e9d647ffc

addon/bootstrap.js
6a68: Great question. I think I should add a check that the `reason` is startup, and not upgrade

ianb: Even that seems a little fragile.  Is there a flag that changes when that event fires, so we can tell if the event won't fire?  Kind of like a readyState?

6a68:
Thanks for the review, @ianb. Now resolving the promise unless the startup `reason` is APP_STARTUP. Merging

PR #2585, opened 2017-04-06
Starred button

ianb:
Also onboarding flow, and a small tweak to bin/release-version


addon/webextension/background/main.js
```diff
@@ -14,6 +14,9 @@ window.main = (function () {
 
   browser.storage.local.get(["hasSeenOnboarding"]).then((result) => {
     hasSeenOnboarding = !! result.hasSeenOnboarding;
+    if (! hasSeenOnboarding) {
+      setIconActive(false, null);
```
dannycoates: nit: maybe we could leave the `null` off of these calls and default `tabId` to null. later is fine

addon/webextension/background/main.js
```diff
@@ -58,6 +68,11 @@ window.main = (function () {
 
   browser.browserAction.onClicked.addListener(catcher.watchFunction((tab) => {
     if (shouldOpenMyShots(tab.url)) {
+      if (! hasSeenOnboarding) {
+        sendEvent("goto-onboarding", "selection-button");
```
dannycoates: do we need to check the telemetry pref first here?

addon/webextension/background/main.js
```diff
@@ -35,7 +42,10 @@ window.main = (function () {
   }
 
   function setIconActive(active, tabId) {
-    const path = active ? "icons/icon-highlight-38.png" : "icons/icon-38.png";
+    let path = active ? "icons/icon-highlight-38.png" : "icons/icon-38.png";
+    if ((! hasSeenOnboarding) && ! active) {
+      path = "icons/icon-38-starred.png";
```
pdehaan: nit: This naming feels off to me. icon-highlight-38.png vs icon-38-starred.png

addon/webextension/background/main.js
ianb: Technically we could leave it off and it'll default to undefined right now!  But I don't know, how can you know what's right in JavaScript, amiright?

PR #2590, opened 2017-04-06
set registrationInfo from old addon

### LINKED ISSUE #2588: auth.setDeviceInfoFromOldAddon never updates registrationInfo
It updates the info in storage, but doesn't update the `registrationInfo` value

dannycoates:
fixes #2588

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -138,5 +138,6 @@ function handleMessage(msg, sender, sendReply) {
       }
       sendReply({type: "success", value: !!addon});
     });
+    return true;
   }
 }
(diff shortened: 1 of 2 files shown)


ianb:
I found a bug while testing, but now this is working for me.  Somewhat tedious reproduction steps:

1. Use `./bin/run-addon --setup-profile`
2. Uninstall any Screenshots or Page Shot addons
3. Install Page Shot
4. Make a shot
5. Quit
6. Run `./bin/run-addon --bootstrap`
7. Go to My Shots to confirm shot from step 4 is there

This errored out before the patch (was more of a problem than not saving the auth info), and now works.  I added a patch to handle the async response.

PR #2591, opened 2017-04-06
Webextension review changes

6a68:
This seems to be working fine for me locally. The outerHTML commit is reverted, I'll work on fixing that right now.

Still unaddressed things from the big review:

- #2592 callBackground.js:
  - need to see if Mark has an .eslintrc (comment at line 1), then remove the eslint globals
  - look into rejecting the Promise from the parent side (comment on line 8)
- #2593 shared/shot.js: deepEqual function changes, makeUuid changes, asJson vs toJson (all kinda scary and not clear we _have_ to change them before uplift)
- #2594 background/communication.js: return a promise.resolve (line 34), don't parse error string (line 56, already an open question there)
- #2595 sitehelper.js: replace JSON.stringify with `cloneInto` (line 16, kinda seems optional IMO)
- (done) #2596 randomString.js: array indexing fix (line 8)

*edited to add bug numbers*

[Ian] Also added:
- #2603 to change `error.message` check
- #2604 to have a logging level
- #2605 to experiment with a `moz-extension:` iframe

### DIFF
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -1,5 +1,7 @@
 /* globals main, auth, catcher, deviceInfo, communication */
 
+"use strict";
+
 window.analytics = (function () {
   let exports = {};
 
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -2,7 +2,7 @@
 
 "use strict";
 
-window.analytics = (function () {
+var analytics = (function () {
   let exports = {};
 
   let telemetryPrefKnown = false;
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -42,8 +42,8 @@ var analytics = (function () {
       options.applicationName = di.appName;
       options.applicationVersion = di.addonVersion;
       let abTests = auth.getAbTests();
-      for (let testName in abTests) {
-        options[abTests[testName].gaField] = abTests[testName].value;
+      for (let [gaField, value] of Object.entries(abTests)) {
+        options[gaField] = value;
       }
       console.info(`sendEvent ${eventCategory}/${action}/${label || 'none'} ${JSON.stringify(options)}`);
       req.send(JSON.stringify({
--- a/addon/webextension/background/auth.js
+++ b/addon/webextension/background/auth.js
@@ -1,6 +1,8 @@
 /* globals browser */
 /* globals main, makeUuid, deviceInfo, analytics, catcher, defaultSentryDsn, communication */
 
+"use strict";
+
 window.auth = (function () {
   let exports = {};
 
--- a/addon/webextension/background/auth.js
+++ b/addon/webextension/background/auth.js
@@ -31,8 +31,8 @@ var auth = (function () {
 
   function generateRegistrationInfo() {
     let info = {
-      deviceId: "anon" + makeUuid() + "",
-      secret: makeUuid()+"",
+      deviceId: `anon ${makeUuid()}`,
+      secret: makeUuid(),
       registered: false
     };
     return info;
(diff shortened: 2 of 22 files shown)


6a68:
@ianb It seems like setting the iframe's contentDocument innerHTML works equivalently to the previous use of DOMParser (see most recent commit, [7f03043](https://github.com/mozilla-services/screenshots/pull/2591/commits/7f03043935e81077328880fcbdf740a8f5be0b35)). The html document inside the iframe is now not fully replaced when the overlay is rendered, but if I take multiple screenshots without reloading a content page, I don't see any errors related to multiple copies of scripts or handlers running in the iframe. Thoughts?

6a68:
Rebased against current master

addon/webextension/background/auth.js
```diff
@@ -29,8 +31,8 @@ window.auth = (function () {
 
   function generateRegistrationInfo() {
     let info = {
-      deviceId: "anon" + makeUuid() + "",
-      secret: makeUuid()+"",
+      deviceId: `anon ${makeUuid()}`,
```
ianb: This introduces a space into the string, so `anon UUID` instead of `anonUUID`

addon/webextension/background/analytics.js
```diff
@@ -13,7 +15,7 @@ window.analytics = (function () {
       return Promise.resolve();
     }
     if (! telemetryPref) {
-      console.info(`Cancelled sendEvent ${eventCategory}/${action}/${label || 'none'} ${JSON.stringify(options)}`);
+      dump(`Cancelled sendEvent ${eventCategory}/${action}/${label || 'none'} ${JSON.stringify(options)}`);
```
ianb: I opened #2604 for this, which I would prefer to `dump()`

addon/webextension/background/main.js
```diff
@@ -108,11 +110,6 @@ window.main = (function () {
     title: browser.i18n.getMessage("contextMenuLabel"),
     contexts: ["page"],
     documentUrlPatterns: ["<all_urls>"]
-  }, () => {
```
ianb: I think this catcher is the only way to catch an error from contextMenu, unlike some of the other uses of `lastError`.

addon/webextension/catcher.js
```diff
@@ -42,27 +44,29 @@ window.catcher = (function () {
...
+      console.error("------Error in promise:", e);
       console.error(e.stack);
       exports.unhandled(makeError(e));
       throw e;
     });
   };
 
   exports.registerHandler = function (h) {
+    if (handler) {
+      console.error("registerHandler called after handler was already registered");
```
ianb: We should have `catcher.unhandled(new Exception("registerHandler called after handler was already registered"))`

ianb: I take that back, catcher calling catcher will lead to bad things

shared/shot.js
```diff
@@ -181,16 +178,15 @@ function deepEqual(a, b) {
...
   let id = "";
   while (id.length < 12) {
-    // 46656 == Math.pow(36, 3)
     let num;
     if (! id) {
-      num = Date.now() % 46656;
+      num = Date.now() % Math.pow(36, 3);
     } else {
-      num = Math.floor(Math.random() * 46656);
+      num = Math.floor(Math.random() * Math.pow(36, 3));
```
dannycoates: Do these ids need to be globally unique or just within a single account? If global, can we use `crypto` here? `Date.now()` and `Math.random()` make me nervous here.

ianb: We only want the ids for clips generated within a single shot to not conflict.  This isn't used for any security-conscious situation.  Things like clip image URLs are generated with UUIDs on the server.

6a68:
@ianb Should I merge this, or do you want to leave it open till Kris finishes responding to the review?

ianb:
@6a68 I think Kris can still leave comments after it closes, and it's already bitrotting, let's get it landed

6a68:
:+1: I"ll squash, rebase and land and push the fixes commit to the PR branch Kris is looking at

Issue #2147, opened 2017-02-08
[Enhancement] Shots titles should have a character limit or they should be truncated after a specific numbers of characters


SoftVision-CosminMuntean:
[Affected versions]:
- Firefox 51.0.1

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Page Shot add-on was previously installed on the profile from "https://pageshot.dev.mozaws.net/homepage/install-test-local.html".

[Steps to reproduce]:
1. Open the profile from prerequisites and navigate to any website.
2. Click the Page Shot button from the browser toolbar.
3. Perform a selection an save it.
4. Click on the saved shot title and change it with a long string.
5. Observe the new title of the shot in preview page .
6. Click the "My Shots" button from the top right of the page.
7. Observe the new title of the shot.

[Expected result]:
Step 4: - A "character limit" warning is displayed for the long string. 
or
Step 5 & 7: - The long title is displayed as truncated on both pages. 

[Actual result]:
- There is no character limitation for shot rename and there is no truncation on the pages where it is displayed.

[Notes]:
- The large shot title also affects the shots from "https://pageshot.dev.mozaws.net/shots" page. All the thumbnails from the same row will gain the same dimension as the one with the long title.
- Here are some screen shots with the issue:

![title saved page shot](https://cloud.githubusercontent.com/assets/19226534/22749278/6a61c67a-ee35-11e6-8b89-0e38f5b13317.png)

![saved shot tile](https://cloud.githubusercontent.com/assets/19226534/22749283/6d6b6786-ee35-11e6-931c-856b3b5e8b28.png)

ianb:
Let's do a 140 character limit on display (we'll keep saving the full title).

SoftVision-CosminMuntean:
I have retested this issue and the long title is correctly displayed as truncated on "My Shots"page.

But, if you click the "trash" icon, the title is not truncated in the displayed "delete" dialog.
Here is a screen recording:
![shot title](https://cloud.githubusercontent.com/assets/19226534/25128647/121773b8-2443-11e7-9d37-c74a1c20cc4b.gif)

PR #2613, opened 2017-04-08
Sets character limit on shot title (WIP)

niharikak101:
FIx for #2147

### DIFF
--- a/server/src/pages/shotindex/view.js
+++ b/server/src/pages/shotindex/view.js
@@ -131,10 +131,9 @@ class Body extends React.Component {
   }
 
   displayTitle(title) {
-    if(title.length > 140) {
+    if (title.length > 140) {
       return (title.substring(0, 140) + "...");
-    }
-    else {
+    } else {
       return title;
     }
   }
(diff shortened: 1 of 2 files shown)


dannycoates:
This isn't really my area of expertise but might we use [text-overflow](https://developer.mozilla.org/en/docs/Web/CSS/text-overflow) CSS here instead?

ianb:
I think we have it working in server/src/pages/shot/view.js through truncating (doesn't look great, but it's better than when #2147 was opened).  But I think the issue with My Shots still remains.  

@niharikak101: can you change this to only due the truncation on My Shots?

As written `displayTitle(title)` is also duplicated across shot/view.js and shotindex/view.js though if this is removed from shot/view.js then that won't matter.

server/src/pages/shotindex/view.js
```diff
@@ -130,6 +130,15 @@ class Body extends React.Component {
     );
   }
 
+  displayTitle(title) {
+    if(title.length > 140) {
```
ianb: There should be a space after "if"

server/src/pages/shotindex/view.js
```diff
@@ -130,6 +130,15 @@ class Body extends React.Component {
...
+      return (title.substring(0, 140) + "...");
+    }
+    else {
```
ianb: "else" should be on the same line as "}", like: `} else {`

niharikak101:
@ianb @dannycoates 
Understood :) I'll make the changes!

ianb:
Thanks!

PR #2614, opened 2017-04-09
Improve localization comments

flodolo:
Based on some of the feedback received so far from localizers on potentially confusing strings.


ianb:
Thanks!

PR #2631, opened 2017-04-10
force content-type image/png for /images

### LINKED ISSUE #2466: Force a PNG content type when serving data

dannycoates:
Is this all we need here?

fixes #2466


ianb:
Yup, no rocket science here ;)

Issue #2633, opened 2017-04-11
unloadModules is broken


dannycoates:
moving from `window.foo = (function () {` to `var foo = (function () {` broke the `unloadModules` function, which breaks a lot of the addon.

dannycoates:
ref #2591

6a68:
:-(

PR #2634, opened 2017-04-11
go back to window.*

dannycoates:
the quick fix for #2633 is to go back to `window`. Otherwise feel free to grind on `unloadModules`


ianb:
Onward and upward, even when it's backward!

PR #2638, opened 2017-04-11
restrict req.backend to a config origin

dannycoates:


server/src/server.js
```diff
@@ -256,7 +256,8 @@ app.use(function (req, res, next) {
     }
   }
   req.abTests = authInfo.abTests || {};
-  req.backend = `${req.protocol}://${req.headers.host}`;
+  const host = req.headers.host === config.contentOrigin ? config.contentOrigin : config.siteOrigin;
```
ianb: Might be nice to log cases when we get an unusual host header, otherwise looks good

dannycoates: would that be actionable in any way? IME unactionable lines end up as noise

PR #2640, opened 2017-04-11
Update the server names in the export script and docs.

Standard8:


bin/export_mc.py
```diff
@@ -7,10 +7,10 @@
...
 DEFAULTS = {
-    "local": "http://localhost:10080/",
-    "dev": "https://pageshot.dev.mozaws.net/",
-    "stage": "https://pageshot.stage.mozaws.net/",
-    "prod": "https://pageshot.net/"
+    "local": "http://localhost:10080",
+    "dev": "https://screenshots.dev.mozaws.net",
+    "stage": "https://screenshots.stage.mozaws.net",
+    "prod": "https://screenshots.firefox.com"
 }
```
6a68: I confirmed with @relud that these are the correct URLs

PR #2645, opened 2017-04-11
Fix #2642: Improve exporting files to mozilla-central, avoid duplicate/empty L10n files

### LINKED ISSUE #2642: Stop exporting duplicate / empty message.json locale files.
Currently we're hitting a test failure (I disabled on some of my previous exports), that we're creating duplicate files with the message.json.

These are mainly ones with a body of just `{}`, so we should try to filter those out and not ship them.

Additionally we should check for any matching locales, e.g. en-US and en-GB.

Standard8:
The m-c build system protects against duplicate files, so we need to avoid those. Additionally, this patch makes it
so that we don't export empty l10n files as we just fallback to en-US.


6a68:
Not a huge deal right now, but the `./bin/export_mc.py` example incantation in the export-to-firefox markdown file fails if `--mozilla-central-repo` is set and `EXPORT_MC_LOCATION` is unset.

PR #2657, opened 2017-04-12
added csrf protection

### LINKED ISSUE #2331: Add CSRF to /api/delete-shot
I'm pretty sure `/api/delete-shot` has a danger of CSRF.  We should sign those requests in some fashion.

dannycoates:
fixes #2331 and other important routes

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "convict": "3.0.0",
     "cookies": "0.7.0",
     "core-js": "2.4.1",
+    "csurf": "1.9.0",
     "envc": "2.5.0",
     "escape-html": "1.0.3",
     "eslint-plugin-promise": "3.5.0",
--- a/server/src/pages/shot/controller.js
+++ b/server/src/pages/shot/controller.js
@@ -97,7 +97,7 @@ exports.changeShotExpiration = function (shot, expiration) {
       location.reload();
     };
   }
-  req.send(`id=${encodeURIComponent(shot.id)}&expiration=${encodeURIComponent(expiration)}`);
+  req.send(`id=${encodeURIComponent(shot.id)}&expiration=${encodeURIComponent(expiration)}&_csrf=${encodeURIComponent(model.csrfToken)}`);
 };
 
 exports.deleteShot = function (shot) {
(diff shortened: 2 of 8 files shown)


server/src/pages/shot/controller.js
```diff
@@ -97,7 +97,7 @@ exports.changeShotExpiration = function (shot, expiration) {
       location.reload();
     };
   }
-  req.send(`id=${encodeURIComponent(shot.id)}&expiration=${encodeURIComponent(expiration)}`);
+  req.send(`id=${encodeURIComponent(shot.id)}&expiration=${encodeURIComponent(expiration)}&_csrf=${model.csrfToken}`);
```
ianb: I like to use `encodeURIComponent()` everywhere even when I'm pretty sure it won't be needed.

server/src/reactrender.js
```diff
@@ -9,18 +9,21 @@ exports.render = function (req, res, page) {
     model.backend = req.backend;
     let jsonModel = model.jsonModel || model;
     let serverModel = model.serverModel || model;
+    let csrfToken = req.csrfToken && req.csrfToken();
```
ianb: When would this be not be there?

server/src/server.js
```diff
@@ -59,6 +59,7 @@ const {
 const dbschema = require("./dbschema");
 const express = require("express");
 const bodyParser = require('body-parser');
+const csrf = require("csurf");
```
ianb: Renaming it on import seems odd...?

package.json
```diff
@@ -13,6 +13,7 @@
     "convict": "3.0.0",
     "cookies": "0.7.0",
     "core-js": "2.4.1",
+    "csurf": "^1.9.0",
```
ianb: Should be exact version

server/src/reactrender.js
dannycoates: it won't be there on routes that don't use the `csrfProtection` middleware

server/src/server.js
dannycoates: I guess I did this cuz the csurf readme does

PR #2672, opened 2017-04-12
Log levels

ianb:


addon/webextension/log.js
```diff
@@ -0,0 +1,47 @@
+/* globals buildSettings */
+
+"use strict";
+
+this.log = (function () {
+  let exports = {};
+
+  const levels = ["debug", "info", "warn", "error"];
+  if (! levels.includes(buildSettings.logLevel)) {
+    console.warn("Invalid buildSettings.logLevel:", buildSettings.logLevel);
```
6a68: Probably should use a template string here, was mentioned in the webext review

addon/webextension/log.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+  let shouldLog = {};
+
+  {
+    let startLogging = false;
+    for (let level of levels) {
+      if (buildSettings.logLevel == level) {
```
6a68: `===` here?

addon/webextension/log.js
ianb: This isn't concatenation, it's just normal console.* multiple argument usage

addon/webextension/log.js
ianb: Mozilla style is unclear on this, e.g., [this](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style) recommends `==`.  The code itself is kind of all over the place.

I changed, but probably best to let some future eslint rule manage this.

PR #2673, opened 2017-04-12
use a moz-extension src for the overlay iframes

### LINKED ISSUE #2605: Investigate moz-extension iframe
Right now we are loading the iframes directly in content, and so the page can see the iframe and can interact with it.

We might be able to change these iframes to be hosted as web-accessible add-on resources, served up from `moz-extension:` URLs.  But I'm not confident that this will all work, so first we should see just what's possible before actually committing to the work.  We should specifically check this on a site like github.com that has CSP rules that would normally forbid these iframes.

dannycoates:
I tried this on a couple pages including github.com as Ian suggested. It seems to work as expected but I'm not familiar with the nitty-gritty of what's supposed to happen.

Before, running code like this in the console would work:

`window.frames[1].document.getElementsByClassName('preview-overlay')[0].style.display = 'none'`

After:

`SecurityError: Permission denied to access property "document" on cross-origin object`

fixes #2605


ianb:
There's another iframe in slides.js that should be updated.  Otherwise looks good.  I'm slightly worried that a zero-length file could get cleaned up in some build process, `<html></html>` in blank.html would protect against that.

PR #2680, opened 2017-04-13
Update metrics for deviceid

ianb:


ianb:
I don't know why there's that extra commit for the version bump, I thought I pushed that... but maybe I didn't?  Or maybe I pushed the tag and thought I had pushed the whole branch, but then because I merged to stable locally I pulled that change into stage.  I guess it needs to land eventually...

PR #2697, opened 2017-04-17
removed Test Pilot reference from /leave

dannycoates:


server/src/pages/leave-screenshots/view.js
```diff
@@ -50,10 +50,6 @@ class Body extends React.Component {
           <div className="large-icon-message-container">
             <div className="large-icon check" />
             <div className="large-icon-message-string">All of your Firefox Screenshots data has been erased.<br/>
```
pdehaan: Not sure if you want to remove the trailing `<br/>` too.

SoftVision-CosminMuntean:
@dannycoates I have noticed that on "https://screenshots.stage.mozaws.net/leave-screenshots/?complete" page, the Test Pilot references are still displayed. The Test Pilot references are also displayed on dev server.
Or the fix will land only after the deploy on production?

Here is a screenshot with the page:
image

dannycoates:
@SoftVision-CosminMuntean I don't think stage has this code yet

Issue #2655, opened 2017-04-12
Error from bootstrap.js


ianb:
I saw this error in the browser console:

```
DOMException [InvalidStateError: "A mutation operation was attempted on a database that did not allow mutations."
code: 11
nsresult: 0x80660006
location: resource://gre/modules/IndexedDB.jsm:294]  bootstrap.js:113

InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.  (unknown)
	transaction resource://gre/modules/IndexedDB.jsm:294:39
	objectStore resource://gre/modules/IndexedDB.jsm:322:23
	get resource://gre/modules/ExtensionUtils.jsm:191:7
	next self-hosted:1126:9
```

The relevant function:

```js
function start(webExtension) {
  webExtension.startup().then((api) => {
    api.browser.runtime.onMessage.addListener(handleMessage);
  }).catch((err) => {
    // The startup() promise will be rejected if the webExtension was
    // already started (a harmless error), or if initializing the
    // WebExtension failed and threw (an important error).
    console.error(err);
    if (err.message !== "This embedded extension has already been started") {
      // TODO: Should we send these errors to Sentry? #2420
    }
  });
}
```

The add-on seemed to have started up regardless.

ianb:
@kmaglione says:

```
2:10 PM <John-Galt> rhelmer: That sometimes happens when we flush the startup cache during add-on upgrades, because of all the racy async shutdown/startup stuff that the add-on manager doesn't know about.
2:11 PM <rhelmer> where is the racy async shutdown/startup stuff, webextensions code?
2:11 PM <John-Galt> But it basically doesn't matter, so we just ignore it and proceed as if the startup cache isn't there.
2:12 PM <John-Galt> No, AOM, mostly. It needs to learn how to wait for startup() and shutdown() to complete before going on to the next one.
2:13 PM <John-Galt> But there are other things we can do to deal with the WebExtension startup cache bits. I want to stop doing the whole blanket cache flush thing, at some point.
```

It might be nice to capture/ignore this specific error, or at least put a note in the browser console.

ianb:
We explored this in more depth in #2911 and after changes to Firefox believe it is no longer a problem.

PR #2704, opened 2017-04-18
Fix #2695, expand Selenium test

### LINKED ISSUE #2695: Expand Selenium test
The test in `test/test.js` should be updated to test the actual functionality.

We want to run a test that:

- [x] Invokes Screenshots
- [x] Makes a selection, or maybe just hits "Save Visible"
- [x] Uses the homepage as its page to make a selection from
- [x] Confirms that the tab opens, is updated to the shot page
- [ ] Notices any `console.error` or `console.warn` messages
- [ ] Maybe notices the exact sequence of `sendEvent` calls (many of the results can be confirmed this way, instead of trying to detect items in the interface itself)
- [ ] Extra points: load My Shots and make sure it is authenticated

ianb:
Implement a test that invokes Screenshots, skips onboarding, clicks the Visible save button, and confirms that a new tab is opened to a shot page.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,12 @@ zip: addon
 	# We'll try to remove this directory, but it's no big deal if we can't:
 	rmdir web-ext-artifacts || true
 
+.PHONY: bootstrap_zip
+bootstrap_zip: addon
+	@rm -f build/screenshots-bootstrap.zip
+	cd addon && zip -rq ../build/screenshots-bootstrap.zip .
+	# build/screenshots-bootstrap.js created
+
 .PHONY: signed_xpi
 signed_xpi: addon
 	rm -f web-ext-artifacts/*.xpi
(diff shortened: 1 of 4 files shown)


ianb:
Currently, on Nightly, the test installation process does not overwrite the built-in version of the add-on.  You can see this because of a failure that it opens a page on screenshots.firefox.com instead of localhost:10080.  I'm not sure why, though I see this error in the Browser Console:

image

It seems to claim that it is overwriting the add-on, but then there's the error.  I recall this error from #2655, and @kmaglione said it was not a problem there, but maybe it means something in this case.

Though CircleCI is *not* showing this error, it actually seems to be working okay, except that the server isn't available locally (maybe fixed by ce5f4469e78a85c7b14e421142b08ea070e2b8ea)

6a68:
I can't find the error now, but I'm pretty sure I saw an error in today's nightly about failing to update system addons. Can you repro this on dev edition or a slightly older Nightly?

6a68:
Hmm, here's a bug about system addons not getting updated, but it's labeled as fixed: https://bugzilla.mozilla.org/show_bug.cgi?id=1350064

Maybe ping rhelmer to see if this is a known issue?

rhelmer:
We chatted about this in IRC - the tl;dr is that bug 1350064 is unrelated (it's about updates sticking around after application upgrade) - the problem is that the WebExtension started by the built-in add-on is not shut down when the bootstrapped add-on is.

I tested this locally and stopping the WebExtension on bootstrap.js `shutdown()` seems to do the trick.

PR #2709, opened 2017-04-19
added DMCA notice to shot page

dannycoates:
This adds a dmca field to the `data` table and shows a notice on the shot page when that field is not null. It could also be used for a link to specific info but is currently only used as a flag for a general message.

### DIFF
--- a/server/db-patches/patch-15-16.sql
+++ b/server/db-patches/patch-15-16.sql
@@ -1,0 +1,2 @@
+CREATE TYPE shot_block_type AS ENUM ('none', 'dmca');
+ALTER TABLE data ADD COLUMN block_type shot_block_type DEFAULT 'none' NOT NULL;
--- a/server/db-patches/patch-16-15.sql
+++ b/server/db-patches/patch-16-15.sql
@@ -1,0 +1,2 @@
+ALTER TABLE data DROP COLUMN block_type;
+DROP TYPE shot_block_type;
--- a/server/src/dbschema.js
+++ b/server/src/dbschema.js
@@ -4,7 +4,7 @@ const pgpatcher = require("pg-patcher");
 const path = require("path");
 const mozlog = require("mozlog")("dbschema");
 
-const MAX_DB_LEVEL = exports.MAX_DB_LEVEL = 15;
+const MAX_DB_LEVEL = exports.MAX_DB_LEVEL = 16;
 
 exports.forceDbVersion = function(version) {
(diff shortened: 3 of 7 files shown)


ianb:
Something that we're planning on for new updates (that we haven't done before) is to make the code tolerant of different database versions.  So for our purposes, 6.3.0 is the current production and runs with database version 15.  We should ensure that 6.3.0 can run with database version 16, and that master can run with database version 15.  *Or*, if we can't do that, then we are supposed to release (to production) another version which can do that.  We haven't done this yet, so I don't actually know quite what the migrations should look like!

ianb:
After discussion, we're not going to worry about the migration/database compatibility stuff for now.

dannycoates:
> I'm pretty sure this is leaky, it'll send all the data but just hide it on render

It never actually fetches the image. The prerendered html has the block inlined. The json contains all the data, including the image url, but never fetches it. I updated the image url to 404 on blocked images (technically not "not found" but whatevs). So whatever is "leaked" isn't useful. It might be nice to refactor things in the future, but this keeps the changes pretty minimal for now. 

I went with a single `block_type` column and an enum to represent unblocked (none) and DMCA. We can add more with `ALTER TYPE shot_block_type ADD VALUE 'abuse'` for example. I didn't add a `blocked_reason` column yet because I'm not clear on its use. Is there already a place where dmca notices are hosted? If not I'd rather wait to add that column.

dannycoates:
@ckprice screenshots

Owner sees:

screen shot 2017-04-24 at 11 05 46 am

Others see:

screen shot 2017-04-24 at 11 04 53 am

ckprice:
@dannycoates thanks! Just passed this around to legal.

dannycoates:
@ckprice updated. I'm not exactly sure what "generic not available page" means so I used the 404 page as is shown for any unknown shot url. cool?

creator:

screen shot 2017-04-25 at 8 38 46 am

non-creator:

screen shot 2017-04-25 at 8 42 36 am

clouserw:
@dannycoates that's fine, thanks (cory is pto today)

server/db-patches/patch-15-16.sql
```diff
@@ -0,0 +1,2 @@
+CREATE TYPE shot_block_type AS ENUM ('none', 'dmca');
+ALTER TABLE data ADD COLUMN block_type shot_block_type DEFAULT 'none' NOT NULL;
```
ianb: I guess I would have used NULL as not-blocked, but I suppose this is fine too.

PR #2712, opened 2017-04-19
Shutdown the embedded webExtension when bootstrap is asked to shutdown

ianb:
@rhelmer or @6a68

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -76,6 +76,13 @@ function startup(data, reason) { // eslint-disable-line no-unused-vars
 
 function shutdown(data, reason) { // eslint-disable-line no-unused-vars
   prefObserver.unregister();
+  const webExtension = LegacyExtensionsUtils.getEmbeddedExtensionFor({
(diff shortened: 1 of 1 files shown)


rhelmer:
lgtm, this is almost exactly what I tested locally.

6a68:
@ianb This looks good. Sadly, I think we need to change the "if reason == APP_STARTUP" bit from the startup method--otherwise I think the pref observer will stop working when the addon is upgraded. Would you mind setting a flag when the observer is registered, then checking that flag, instead of the reason constant, inside startup? We always want exactly one listener.

ianb:
@6a68 I'm not sure I understand.  `if (reason === APP_STARTUP)` only protects `appStartupObserver.register()` (which seems to unregister itself).  But it doesn't stop the rest of startup() from being called, including the pref observer.  Unless I miss what you are thinking about here.

6a68:
@ianb Ah yeah, you're right! Sorry about that, merging

PR #2713, opened 2017-04-19
Add third argument to add/removeObserver, which is required in Firefox 54

ianb:


ianb:
I think the build failure is simply due to Screenshots being built into Nightly now and pref'd off

6a68:
@ianb do we need to uplift this with the initial round of patches?

ianb:
No, this was a regression from @Standard8's eslint patches, which aren't part of what's in Nightly now

PR #2726, opened 2017-04-21
Fix #2698, disable selection after a scroll event when the mouseup is…

### LINKED ISSUE #2698: After dragging the scrollbar, a selection is automatically performed when hovering the page
[Affected versions]:
- Nightly 55.0a1 (Build ID:20170418030220)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have latest Nightly build with Screenshots installed and a clean profile.
- The "extensions.screenshots.system-disabled" pref is set to false.

[Steps to reproduce]:
1. Open the browser and profile from prerequisites and navigate to any website (eg: imdb.com).
2. Click the Screenshots button from browser toolbar.
3. Click the Firefox scrollbar and scroll the page.
4. Release the mouse and hover the page content.
5. Observe the brower behavior.

[Expected result]:
- The Screenshots overlay is not dismissed and no selection is performed.

[Actual result]:
- Automatically a selection is performed when hovering the page content.

[Notes]:
- The issue is also reproducible on stage server.
selection vs scrollbar

ianb:
… in the scroll area

### DIFF
--- a/addon/webextension/selector/uicontrol.js
+++ b/addon/webextension/selector/uicontrol.js
@@ -57,6 +57,9 @@ this.uicontrol = (function() {
   // This is how close (in pixels) you can get to the edge of the window and then
   // it will scroll:
   const SCROLL_BY_EDGE = 20;
+  // This is how wide the inboard scrollbars are, generally 0 except on Mac
+  const SCROLLBAR_WIDTH = (window.navigator.platform.match(/Mac/i)) ? 17 : 0;
+
 
   const { sendEvent } = shooter;
 
@@ -386,7 +389,7 @@ this.uicontrol = (function() {
         return;
       }
       let el;
-      if (event.target.classList.contains("preview-overlay")) {
+      if (event.target.classList && event.target.classList.contains("preview-overlay")) {
         // The hover is on the overlay, so we need to figure out the real element
         el = ui.iframe.getElementFromPoint(
           event.pageX + window.scrollX - window.pageXOffset,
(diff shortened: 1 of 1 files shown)


dannycoates:
i can still trigger the bug... looking into why

dannycoates:
On my mac with "magic" scrollbars I need to scroll a bit to get the bar to show up, but once I do I seem to be lingering over the scrollbar for more than a second before I mousedown which causes me to be over `SCROLLBAR_SCROLL_TIME_LIMIT`. 

When I set scrollbars to be always on, I get the selection behavior in the "normal" case as well, i.e. the first scroll movement is after I mousedown on the scrollbar, which seems like it'd be more normal on Windows. 

This patch improves on the current result for some cases, but it seems like the (most?) common usage of this (IMHO outmoded) scrolling behavior still results in the undesired selector.

I personally would rather WONTFIX because it seems really difficult to do conclusively but I'm ok with this minor improvement too.

ianb:
Increasing `SCROLLBAR_SCROLL_TIME_LIMIT` would seem fine to me.  It's a heuristic to be sure, but it's also fairly harmless – it basically makes the very right part of the screen inactive for a period after scrolling.

dannycoates:
With scrollbars set always visible:
1. click screenshot button
2. click scrollbar and drag down
3. release scrollbar and move left
4. selector appears

spinner

6a68:
Rather than use a timeout, what if we just disable the next mouseup when a mousedown is detected to the right of the body element?

6a68:
I guess it'd be slightly more complex, since the following mouseup might not land inside the browser window:

```
on scrollbar-mousedown,
  set ignoreNextMouseUp = true;
if the next mouse event is a mouseup, ignore it;
if the next mouse event is a mousedown, handle as normal and unset ignoreNextMouseUp

ianb:
Oh, I see now, my logic was all backwards.  Because mousedown triggers the action, the scroll happens after it, not before.  It was only working for me because I initiate a scroll to see the scrollbars.  Pondering...

ianb:
Redone with no timer or scroll detection, just a dead zone.

6a68:
I think this works really nicely. Let's land it :+1:

PR #2727, opened 2017-04-21
Fix #2643, force onboarding when you visit /#hello

### LINKED ISSUE #2643: Direct link to the onboarding modals
The team running about:home is asking if it would be possible to direct link to a page that will start the onboarding for Screenshots.

It doesn't look like http://localhost:10080/#hello works (even if you haven't visited before). From my testing it only gets started if you click the button. It would be nice if /#hello always started the onboarding.

ianb:


dannycoates:
👍

PR #2735, opened 2017-04-24
WIP - Localize server

6a68:
Opening to get l10n feedback.

I have a few different questions with this code, I'll insert comments at the appropriate lines.


server/src/pages/shot/time-diff.js
```diff
@@ -24,6 +24,12 @@ exports.TimeDiff = class TimeDiff extends React.Component {
     }
   }
```
6a68: @flod or @stasm - This file contains a number of functions that create partial time-related strings, like "in 1 day", or "3 hours ago". These partial strings are then concatenated with phrases like "expires " or "expired ", or, in the case of the `shotPageRestoreButton` l10n ID, "restore for 30 days". 

What's the best way to handle these relative times, given that some languages are declined, so that we couldn't simply concatenate strings to render "restore for 1 day" vs. "expires in 1 day" vs. "expired 1 day ago"?

server/src/pages/shot/view.js
```diff
@@ -180,7 +180,7 @@ class Body extends React.Component {
 
   onClickDelete(e) {
     sendEvent("start-delete", "navbar", {useBeacon: true});
-    if (window.confirm("Are you sure you want to delete this shot permanently?")) {
+    if (window.confirm("Are you sure you want to delete this shot permanently?")) { // todo l10n: shotPageConfirmDeletion
```
6a68: We are planning to use fluent-react, but for cases like this string, where a string is passed into `window.confirm` or `window.alert`, what's the right approach? Should we create an empty Localized tag in the DOM, then pass that into window.confirm?

server/src/pages/shot/view.js
```diff
@@ -420,15 +420,15 @@ class ExpireWidget extends React.Component {
...
-        keep for <select ref="expireTime">
-          <option value="cancel">Select time:</option>
-          <option value="0">Indefinitely</option>
-          <option value={ 10 * minute }>10 Minutes</option>
-          <option value={ hour }>1 Hour</option>
-          <option value={ day }>1 Day</option>
-          <option value={ 7 * day }>1 Week</option>
-          <option value={ 14 * day }>2 Weeks</option>
-          <option value={ 31 * day }>1 Month</option>
+        keep for <select ref="expireTime"> <!-- todo l10n -->
```
6a68: Speaking of declensions, does it make sense to have "keep for: " separate from the time period?

server/src/pages/shotindex/server.js
```diff
@@ -18,7 +18,7 @@ app.get("/", function(req, res) {
   }
   getShots.then(_render)
     .catch((err) => {
-      res.type("txt").status(500).send("Error rendering page: " + err);
+      res.type("txt").status(500).send("Error rendering page: " + err); // todo l10n: shotIndexPageErrorRendering
```
6a68: Here is another edge case I'm not sure about: in some cases, we render a server error string directly into the page.

I've been thinking I could work around this by passing an l10n ID into the page, then running fluent-react on that page, to swap in the appropriate error message. Is that the best approach? Is there a fluent library that could be used instead, for spots like this?

server/src/pages/not-found/model.js
```diff
@@ -1,5 +1,5 @@
 exports.createModel = function(req) {
   return {
-    title: "Page Not Found"
+    title: "Page Not Found" // todo l10n: notFoundPageTitle
```
6a68: Here's another spot where strings are defined in non-view code.

I think the cleanest approach would be to set an l10n ID as the title, then keep all the string substitution in the view code. Thoughts?

locales/en-US/server.properties
```diff
@@ -0,0 +1,150 @@
...
+# TODO: do we need this separate description meta tag?
+homePageDescription2 = Intuitive screenshots baked right into the browser. Capture, save and share screenshots as you browse the Web using Firefox.
+homePageButtonMyShots = Go To My Shots
+homePageTitleBeta = Beta
+homePageTeaser = Coming Soon...
+
+# leave screenshots page
+
+leavePageConfirmDelete = Confirm account deletion
+leavePageErrorAddonRequired = You must have the addon installed to delete your account
```
pdehaan: global: Do we prefer "addon" or "add-on"? (https://addons.mozilla.org/en-US/firefox/)

server/src/pages/shot/time-diff.js
6a68: I'm also wondering if fluent-react supports the advanced options shown here: http://l20n.org/learn/advanced-selectors

server/src/pages/shot/view.js
```diff
@@ -448,15 +448,15 @@ class ExpireWidget extends React.Component {
...
-        &bull; keep for: <select ref="expireTime">
-          <option value="cancel">Select time</option>
-          <option value="0">Indefinitely</option>
-          <option value={ 10 * minute }>10 Minutes</option>
-          <option value={ hour }>1 Hour</option>
-          <option value={ day }>1 Day</option>
-          <option value={ 7 * day }>1 Week</option>
-          <option value={ 14 * day }>2 Weeks</option>
-          <option value={ 31 * day }>1 Month</option>
+        &bull; keep for: <select ref="expireTime"> <!-- todo l10n -->
```
6a68: Speaking of declensions, does it make sense to have "keep for: " separate from the time period?

6a68:
(Rebased against current master to pick up john's string changes)

server/src/pages/homepage/view.js
```diff
@@ -9,14 +9,14 @@ class Head extends React.Component {
       <reactruntime.HeadTemplate {...this.props}>
         <script src={this.props.staticLink("/static/js/homepage-bundle.js")} async></script>
         <meta name="viewport" content="width=320, initial-scale=1.0, maximum-scale=1.0, user-scalable=0" />
-        <meta name="description" content="Share anything on the web with anyone using Firefox Screenshots." />
+        <meta name="description" content="Share anything on the web with anyone using Firefox Screenshots." /> <!-- todo l10n: homePageDescription -->
```
6a68: Another question: how do we localize attributes using fluent-react?

6a68:
Summarizing my questions at this point:
- How should I localize attributes, like `<button title=“Download”>`?
- How should I localize strings defined in models, not in elements:
```js
// server/src/pages/leave-screenshots/model.js
  let complete = "complete" in req.query;
  return {
    title: "Confirm account deletion", // how do we localize this?
    complete
  };
};

```
- How should I localize strings passed directly into `window.alert` or `window.confirm` calls?

server/src/pages/not-found/view.js
```diff
@@ -24,13 +25,19 @@ class Body extends React.Component {
...
+                <a href="/shots">My Shots</a>
+              </Localized>
+            </h1>
           </div>
           <div className="responsive-wrapper flex-1">
             <h2>{this.props.title}</h2>
-            <p>
-              The page was not found.
-            </p>
+            <Localized id="notFoundPageDescription>
```
ianb: syntax error?

server/src/pages/shot/time-diff.js
stasm: @zbraniecki -- what's our best options right now for relative dates?

stasm: The `LEN` function is not available in `fluent`, but selectors still work: http://projectfluent.io/fluent/guide/selectors.html

server/src/pages/shot/view.js
stasm: I think it would be helpful to have an imperative API just for these uses-cases. I imagine something like `formatString`. I'm not sure where to put it, however. The translations are managed by the `LocalizationProvider` component and are made available to `Localized` components via React's context.


What would be the best way to expose a method for retrieving translations from `LocalizationProvider`? Perhaps a HOC which connects a given component with the `LocalizationProvider` and passes a prop: the `formatString` method.

server/src/pages/not-found/view.js
6a68: thx

locales/en-US/server.properties
6a68: good question for @johngruen maybe

server/src/pages/shotindex/server.js
stasm: Communicating l10n ids between the server and the clients was our best practice recommendation in Firefox OS. I'll follow it here as well. This means there's no need to store translations nor negotiate languages on the server at all.

server/src/pages/not-found/model.js
stasm: Sounds good to me. And then just use the title l10n id in `<Localized id="…" /> in the view code.

server/src/pages/shot/time-diff.js
6a68: > I wonder if we should consider using some library

It would be cool if we could either contribute back to that library, or export translations out of mozilla-central into a helpful localized string toolkit (common stuff like "Download" or "Cancel" in many languages)

server/src/pages/homepage/view.js
stasm: Ah, good question! Sorry that the README is lacking. Your questions are actually super helpful in making me understand what the docs should include. :)

There's an example app demonstrating attribute localization at https://github.com/projectfluent/fluent.js/tree/master/fluent-react/examples/text-input

If you wrap an element with `Localized` like this:

        <Localized id="type-name">
          <input
            type="text"
            placeholder="Your name"
            onChange={evt => this.handleNameChange(evt.target.value)}
            value={name}
          />
        </Localized>

You can define an attribute translation in FTL:

    type-name
        .placeholder = Your name

More on attributes: http://projectfluent.io/fluent/guide/attributes.html This should work to `meta` elements, too.

locales/en-US/server.properties
```diff
@@ -0,0 +1,146 @@
...
+
+# home page
+
+# Note: do not translate the phrase 'Firefox Screenshots'.
+homePageDescription = Share anything on the web with anyone using Firefox Screenshots.
+# TODO: do we need this separate description meta tag?
+homePageDescription2 = Intuitive screenshots baked right into the browser. Capture, save and share screenshots as you browse the Web using Firefox.
+homePageButtonMyShots = Go To My Shots
+homePageTitleBeta = Beta
+homePageTeaser = Coming Soon...
```
flodolo: @... -> … (unicode character), same in the rest of the doc, unless there's a very good reason for not doing it

locales/en-US/server.properties
flodolo: Not a matter of preference, addon is the wrong spelling ;-)
https://transvision.mozfr.org/?recherche=add-on&repo=central&sourcelocale=en-US&locale=it&search_type=strings_entities

locales/en-US/server.properties
```diff
@@ -0,0 +1,146 @@
...
+# leave screenshots page
+
+leavePageConfirmDelete = Confirm account deletion
+leavePageErrorAddonRequired = You must have the addon installed to delete your account
+# TODO: do we need this separate error message?
+leavePageErrorAddonRequired2 = You must have the addon installed to leave
+leavePageErrorGeneric = An error occurred
+leavePageWarning = This will permanently erase all of your Firefox Screenshots data.
+leavePageButtonProceed = Proceed
+leavePageButtonCancel = cancel
```
flodolo: Is the case correct here?

locales/en-US/server.properties
```diff
@@ -0,0 +1,146 @@
+# global phrases shared across pages, prefixed with 'g'
+
+# TODO: replace each page's reference to this with the global
+gMyShots = My Shots
+
+# creating page
+
+# Localization note: {title} is a placeholder for the title of the web page
```
flodolo: No need for "Localization note", "Note", just add the comment.

locales/en-US/server.properties
```diff
@@ -0,0 +1,146 @@
...
+metricsPageShotsQueryTitle = Shots By Day
+metricsPageShotsQueryDescription = Number of shots created each day (for the last 30 days)
+metricsPageShotsQueryCount = Number of shots
+metricsPageShotsQueryDay = Day
+metricsPageUsersQueryTitle = Users By Day
+metricsPageUsersQueryDescription = Number of users who created at least one shot, by day (last 30 days)
+metricsPageUsersQueryCount = Number of users
+metricsPageUsersQueryDay = Day
+metricsPageUserShotsQueryTitle = Number of Shots per User
+metricsPageUserShotsQueryDescription = The number of users who have about N total shots
```
flodolo: N?

locales/en-US/server.properties
```diff
@@ -0,0 +1,146 @@
...
+metricsPageUserShotsQueryCount = Number of users
+# TODO: I think we'll want to replace the tilde with something like 'Approximate number of shots'
+metricsPageUserShotsQueryShots = ~Number of shots
+metricsPageRetentionQueryTitle = Retention By Week
+metricsPageRetentionQueryDescription = Length of time users have been creating shots, grouped by week
+metricsPageRetentionQueryUsers = Number of users
+metricsPageRetentionQueryDays = Days the user has been creating shots
```
flodolo: Uhm, I'm not sure I get the meaning of this one

locales/en-US/server.properties
6a68: :+1:

package.json
```diff
@@ -18,6 +18,10 @@
     "envc": "2.5.0",
     "escape-html": "1.0.3",
     "express": "4.15.2",
+    "fluent": "^0.3.1",
+    "fluent-intl-polyfill": "^0.1.0",
+    "fluent-langneg": "0.0.2",
+    "fluent-react": "^0.3.0",
```
pdehaan: nit: Not sure if these should be exact versions, for consistency.

locales/en-US/server.properties
pdehaan: I too am confused by these strings. Is the metrics page user-facing?
What's the URL, I'm not sure if I've seen it.

6a68: See https://screenshots.firefox.com/metrics

locales/en-US/server.properties
6a68: Yup, where N is an integer. The wording for these metrics strings could definitely use some help from non-engineers :-)

server/src/pages/homepage/view.js
6a68: @stasm Thanks for the info! I'm glad my newbie questions can help improve the docs :-)

Is FTL attribute translation supported in Pontoon? I looked at the Pontoon BMO bugs and Github repo a bit today, but couldn't figure out the extent of FTL support.

server/src/pages/shot/time-diff.js
flodolo: I've replied on IRC, but putting it here too. There's some standardization happening, but I think it won't be suitable for a website any time soon (zibi would know for sure): https://bugzilla.mozilla.org/show_bug.cgi?id=1270140

Using external libraries is not ideal, there are concerns about quality and control about the localization living there.

Concatenating is very frequently a path to disaster, especially when it comes to plural, gender, and declensions.
Do we have an idea of how many variations we have? e.g. "expired X seconds|minutes|hours|days|weeks|months|years ago", "expires in X seconds|minutes|hours|days|weeks|months|years"? 

I know it's a step backward in terms of usability, but is there any chance to have "expire(s|d) on X", with X being a localized date?

server/src/pages/homepage/view.js
stasm: Attribute support i currently in review and will land together with https://github.com/mozilla/testpilot/pull/2399. See https://github.com/mozilla/pontoon/pull/518/.

6a68: @stasm The fluent/fluent-react code seems like it's still evolving pretty rapidly. Is there a recommended, more stable library we could use instead, until the fluent APIs have some time to bake?

server/src/pages/shot/time-diff.js
johngruen: FWIW, I modified the strings in question here so the substrings are `expires in` `x days` rather than 'expires``in x days`  since `expires in` is a single part of speech (compound verb). Seems like this should make things easier to localize.

flodolo: We still need to deal with proper plural forms with "x days".
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals

server/src/pages/homepage/view.js
stasm: I've come across [react-intl][1] and [react-globalize][2] before. I haven't used them extensively in my projects. I'd like to test them in depth and write a comparison with fluent-react in the future. They both use MessageFormat for storing translations which isn't supported out of the box by Pontoon and would require some kind of build-time transformation to get into Pontoon.

If you're worried about attributes, Pontoon already supports some FTL and its support is constantly improving :)

[1]: https://github.com/yahoo/react-intl
[2]: https://github.com/globalizejs/react-globalize

6a68: Ah, right - I forgot that react will fight other libraries for control of the DOM.

I'm thinking I will just have the l10n.js file export a getText function, for situations where imperative localization is needed.

6a68:
I've gotten good feedback here. Closing, will reopen when the branch is ready to land. Thanks all!

PR #2747, opened 2017-04-25
Fix #2652, avoid error popups when selector is torn down during save

### LINKED ISSUE #2652: An error is displayed after the Screenshots overlay is dismissed while a shot is being created
[Affected versions]:
- Firefox 55.0a1 (Build ID:20170411153633)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Nightly build with Screenshots installed and a clean profile.
- The "extensions.screenshots.system-disabled" pref is set to false.

[Steps to reproduce]:
1. Open the browser and profile from prerequisites and navigate to any website.
2. Click the Screenshots button from browser toolbar and save a selection.
3. Quickly return to the tab where the shot was saved and dismiss the Screenshots overlay.
4. Observe the browser behavior. 

[Expected result]:
- The overlay is successfully dismissed.

[Actual result]:
- The overlay is dismissed but you get a "Whoa! Firefox Screenshots went haywire. ReferenceError: clipboard is not defined" error.

[Notes]:
- The shot is successfully saved even if the error is displayed.
- Here is screenshot with browser console errors:
errors
- Here is a screen recording with the issue: https://goo.gl/CMVKDp

ianb:
Also bind several modules so that catcher works even after selectorLoader.unloadModules has been called

Note that an error is still signalled when there is a premature teardown, but only to the console

To reproduce the original error, run the server like this:

```
$ TEST_SLOW_RESPONSE=3000 ./bin/run-server
```

This gives enough time to switch tabs and hit Escape to close the selector.

There was another bug that caused the popup to stop displaying when there was an error that happened after teardown.  This also fixes that, while avoiding any popup for this particular situation.

### DIFF
--- a/addon/webextension/catcher.js
+++ b/addon/webextension/catcher.js
@@ -1,4 +1,4 @@
-/* globals log */
+/* globals global */
 
 "use strict";
 
@@ -8,6 +8,8 @@ this.catcher = (function() {
   let handler;
 
   let queue = [];
+
+  let log = global.log;
 
   exports.unhandled = function(error, info) {
     log.error("Unhandled error:", error, info);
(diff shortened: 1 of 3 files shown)


ianb:
This isn't working right, will re-open later.

PR #2761, opened 2017-04-27
Update modules

ianb:


ianb:
Note an update to eslint-plugin-mozilla got held back due to a (packaging?) bug

PR #2762, opened 2017-04-27
Fix testIfLoaded so it can see when a load is in progress

ianb:
Per @johngruen's bug report


6a68:
For future reference:

Steps to reproduce:
- Open FF at about:blank in a fresh profile (the screenshots button should be displayed with a pink star shown in the toolbar button)
  - Incantation I used: `./bin/run-addon --scratch --bootstrap --binary /Applications/FirefoxDevEdition`
- Click the Screenshots button
- The My Shots page opens, onboarding is displayed, and an error notification is shown

With this patch applied, no error notification is generated.

PR #2765, opened 2017-04-27
NOT FOR MERGE: Updates for review

ianb:
Styles and images not in this commit (but included in a separate commit)
Updates files through commit c521008371bc8e0e842b7a9d7a21255787b9950d

This includes all updates to master since the last review was made.  It also includes changes made to address comments from that review.  (It wasn't easy to separate these, since several commits landed between the initial review and commits to address comments.)


addon/webextension/background/main.js
```diff
@@ -85,22 +151,75 @@ window.main = (function () {
...
     }
     return true;
   }
 
+  function isShotOrMyShotPage(url) {
+    // It's okay to take a shot of any pages except shot pages and My Shots
+    if (!url.startsWith(backend)) {
+      return false;
+    }
+    let path = url.substr(backend.length).replace(/^\/*/, "").replace(/#.*/, "").replace(/\?.*/, "");
```
kmaglione: It would be nice to use a `URL` object for this... But failing that, the second two regexps can be consolidated into `.replace(/[#?].*/, "")`

addon/webextension/background/main.js
```diff
@@ -110,27 +229,41 @@ window.main = (function () {
...
   });
 
+  communication.register("downloadShot", (sender, info) => {
+    // 'data:' urls don't work directly, let's use a Blob
+    // see http://stackoverflow.com/questions/40269862/save-data-uri-as-file-using-downloads-download-api
+    const binary = atob(info.url.split(',')[1]); // just the base64 data
+    const data = Uint8Array.from(binary, char => char.charCodeAt(0))
+    const blob = new Blob([data], {type: "image/png"})
+    return browser.downloads.download({
+      url: URL.createObjectURL(blob),
```
kmaglione: This blob URL needs to be destroyed after the download finishes.

addon/webextension/background/selectorLoader.js
```diff
@@ -1,78 +1,134 @@
...
     });
   };
 
-  exports.loadModules = function () {
+  function executeModules(tabId, scripts) {
     let lastPromise = Promise.resolve(null);
     scripts.forEach((file) => {
       lastPromise = lastPromise.then(() => {
-        return browser.tabs.executeScript({file})
+        return browser.tabs.executeScript(tabId, {
```
kmaglione: Is there a reason you're still not using an async function for this?

addon/webextension/makeUuid.js
```diff
@@ -1,10 +1,17 @@
...
   return function makeUuid() { // eslint-disable-line no-unused-vars
+    // get sixteen unsigned 8 bit random values
+    var randomValues = window
+      .crypto
+      .getRandomValues(new Uint8Array(36));
+
     return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
-      var r = Math.random()*16 | 0,
-        v = c === 'x' ? r : (r & 0x3 | 0x8);
+      var i = Array.prototype.slice.call(arguments).slice(-2)[0]; // grab the `offset` parameter
```
kmaglione: If you really want to do the Array.prototype.slice thing, this should be written as `Array.prototype.slice.call(arguments, -2)`. But I'm not sure why you aren't just explicitly specifying this in the arguments list. Failing that, you can use a rest arg `function(c, ...args)` and then `args.slice(-2)[0]`

addon/webextension/selector/shooter.js
```diff
@@ -1,28 +1,47 @@
...
+    // http://stackoverflow.com/questions/3115150/how-to-escape-regular-expression-special-characters-using-javascript
+    return str.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&");
+  }
+
+  function sanitizeError(data) {
+    const href = new RegExp(regexpEscape(window.location.href), 'g');
+    const origin = new RegExp(`${regexpEscape(window.location.origin)}[^\s",>]*`, 'g');
+    const json = JSON.stringify(data)
+      .replace(href, 'REDACTED_HREF')
+      .replace(origin, 'REDACTED_URL');
```
kmaglione: I assume you've gotten a separate data review for this?

ianb: This change came out of the legal/data review, though not an engineering data review.  @fmarier was looking at our data practices, but I don't think we called this part out during that, I'll ping him about this.

addon/webextension/background/selectorLoader.js
ianb: I don't think I understand what you mean...?

ianb: (clarified to refer to async/await) – we haven't used async/await at all yet in the codebase, that's a refactoring I'd like to put off until a later date with less time pressure

addon/webextension/background/main.js
ianb: Bug in #2776, fix in #2777

addon/webextension/background/main.js
ianb: There's another case in main.js too, I created a followup in #2778

ianb:
Closing this as we've gone through the review and filed followups

PR #2781, opened 2017-04-28
Fix #2779, add Sentry information to METRICS.md

### LINKED ISSUE #2779: Add Sentry information to METRICS.md

ianb:


docs/METRICS.md
```diff
@@ -274,6 +274,18 @@ We can do queries directly on the database of Firefox Screenshots.  These might
...
+
+In most cases if an exception report has been sent you will also see a popup notification of an error in Screenshots.  Exceptions use [catcher.unhandled](https://github.com/mozilla-services/screenshots/search?q=catcher.unhandled) or [noPopup](https://github.com/mozilla-services/screenshots/search?q=noPopup), and these are typically unusual cases where a popup would intrude on an otherwise recoverable error.
+
+In addition to the exception message and a stack trace, the error information includes information about your browser (User-Agent), platform, the add-on version, and IP address.  We use [Raven.js](https://docs.sentry.io/clients/javascript/) to collect the error reports.
+
+If an exception occurs in the context of screenshotting a specific page, we specifically filter out any reference to the URL or domain of that page.
+
+Error reporting in the add-on can be opted out of in the same way as other metrics (using the general Telemetry preference, see below).  Error reporting on the website cannot be opted out of.
```
fmarier: Is that because Sentry doesn't honour DNT?

ianb: We have specific code for suppressing metrics with DNT.  Sentry reporting would seem to be outside the scope of what DNT is intended for:

image

fmarier: Maybe we should also point that out in this sentence of the opt-out section:

> The website reads the DNT Header (`navigator.doNotTrack`) and if it is present and set to *1* the website will not send metrics data.

fmarier: something like "will not send metrics data though it may still send error reports."

fmarier: I'm not saying that DNT should block these, rather I'm saying that it's a little unclear whether or not DNT will opt you out of error reports too. One section says there's no opt out, the other seems to imply that DNT is the key to opting out of all data collection from the web service.

ianb: Updated in 709b3eecc8f856cbcc8238d0a0037aa779cac961

ianb:
@fmarier : do you have any more comments or feedback, or is this good to go?

fmarier:
This pull request is good to go.

I'll get back to you regarding the data review. I'm discussing this with Benjamin.

Issue #2210, opened 2017-02-23
Host static assets in CDN


ianb:
This probably isn't too hard, as `staticLink()` is already used on all these links, so we'd just make that one function CDN-configuration-aware.  @relud is going to give a suggestion on how this is configured from ops (the variable to look for, and how we use that variable to construct a URL).

relud:
I'm really flexible on this, so my suggestion doesn't need to be set in stone.

I would have `STATIC_URL=https://pageshot.cdn.mozilla.net/static/` or `STATIC_URL=https://pageshot-cdn.net/static/` and construct the urls as `staticLink(a){return config.STATIC_URL + a}; link = staticLink('minified.js')`. The cdn would then use the server as an upstream, so `https://pageshot.cdn.mozilla.net/static/minified.js` would be backed by `https://pageshot.net/static/minified.js`.

relud:
`pageshot.cdn.mozilla.net` has the advantage that we already have an ssl cert for it, and it's already available in all our CDNs, so if we don't use cloudfront (because of the 35KB average request goal), then it wouldn't cost any extra for ssl when hosted on akamai or edgecast.

pdehaan:
Ref: https://www.webpagetest.org/result/170307_MR_1DRW/1/performance_optimization/

![](https://pageshot.net/images/b28cb3c3-39e9-491e-84ce-aa43ec76eba3.png)

PR #2786, opened 2017-05-01
extended direct-view event with ownership

### LINKED ISSUE #2143: Implement event label to distinguish between owner and non-owner direct-views
currently we have `web/visit/direct-view`, let's split this out into `web/visit/direct-view-owner` and `web/visit/direct-view-non-owner`

dannycoates:
closes #2143

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -208,6 +208,8 @@ These are events that an add-on user can encounter on a shot they own
 32. [x] Click Feedback/mailto button `start-feedback/footer`
 31. [x] Click on clip `web/goto-clip/content`
 32. [x] Click the download button `web/download/navbar`
+33. [x] Visit an image directly, when the image isn't embedded directly in a Screenshots shot page, `web/visit/direct-view-owner`
+34. [x] View an image directly, when the image is being shown as part of a Facebook/Twitter style preview (the og:image or twitter:image), `web/visit/direct-view-embedded-owner`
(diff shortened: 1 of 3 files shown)


dannycoates:
Are we planning on using a CDN for screenshots? I was under the impression that we'll only be using one for site assets.

ianb:
Yes, we were planning on using a CDN (under a separate domain name) for individual images, not just site assets.

dannycoates:
> we were planning on using a CDN (under a separate domain name) for individual images

I think there's more work for this that doesn't have tracking issues like invalidating the CDN cache when the conditions used in the query for [getRawBytesForClip](https://github.com/mozilla-services/screenshots/blob/master/server/src/servershot.js#L303) change (block, delete, expire).

Secondarily, do we have evidence that CDNifying the screenshots will be worth the expense? My uneducated guess is that most shots aren't generally accessed by a large distribution of users. Maybe a simple http cache (nginx, varnish) would be enough. I guess I'm saying I don't understand the expectations / requirements.

cc #2210

ianb:
You're right, there are important implications to the CDN.  @relud did some calculations that using a CDN wouldn't be any more expensive than serving the images directly, even if the images weren't accessed in any volume.

dannycoates:
There's also the complexity cost. I'm not against it if it makes sense, I just want to understand the cost / benefits.

PR #2787, opened 2017-05-01
inline style/image on /#hello

### LINKED ISSUE #2703: Flashes of older content in the onboarding flow
@johngruen reported this, to confirm.

dannycoates:
I can't repro that video, but i guess we could embed the style and image in the html.

fixes #2703


dannycoates:
or maybe its already fixed and we can just close

ianb:
I think the missing styles referred to the slides, not the homepage

dannycoates:
> I think the missing styles referred to the slides, not the homepage

according to john's video in #2703 it was the page

ianb:
Oh huh.  I don't know why/how/if styles are normally applied to pages – it seems like this would happen often, but I never notice it?  If we really wanted to inline styles then we'd want to use fs.readFileSync to load the content, and not inline it into view.js.

dannycoates:
@johngruen or anyone. Can you reproduce the unstyled content since we updated the default page? (I can't) https://screenshots.firefox.com

I'd prefer to just close this.

johngruen:
pulling down now!

johngruen:
LGTM!

PR #2802, opened 2017-05-03
Fix #2792, make clip image.onload resilient

### LINKED ISSUE #2651: An infinite spinner is displayed when opening a non-owner shot taken from the latest Nightly Try build if uBlock Origin is installed
[Notes]:
- The issue is not reproducible if the first ever shot is shared ([Browser Console errors](https://drive.google.com/open?id=0B8ICzPJynbchOGdfc2lyZTNqZDQ)). However it it reproducible with all subsequent shots.
- The issue is also reproducible on Chrome with uBlock Origin installed.
- If I disable uBlock Origin and refresh the page, the image is loaded correctly.
- The issue is not reproducible if the page is shared from a custom build using the Page Shot dev server.
- The issue is not reproducible on Chrome or Firefox with AdBlock Plus installed.

[Affected versions]:
- Firefox 52 and up.

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have uBlock Origin installed on a clean profile.

[Steps to reproduce]:
1. Open the browser and profile from prerequisites.
2. Open a non-owner shot taken with the Nightly try build. (eg. https://screenshots.firefox.com/8JSV1y1KERYowIT3/www.imdb.com)
3. Observe the browser behavior.

[Expected result]:
- The page is loaded and the image is displayed.

[Actual result]:
- The page is loaded, but the spinner loops infinitely and the image is not displayed.

[Additional notes]:
- Here is a screen shot of the issue:
image
- Here are the errors shown in the browser console:
image

### LINKED ISSUE #2792: Make clip image.onload resilient to JavaScript failures
It used to be if JavaScript failed on the page, the page still was viewable and pleasant.  Now it doesn't work, because the image is loaded off-screen and moved on-screen in image.onload.

I propose we do something like move the onload function into an attribute of the image itself, and make sure it doesn't depend on anything else, so that it can execute even when nothing else does.  Also I suspect that we've put an unnecessary block into the image displaying, because now the entire JavaScript bundle has to load and React has to re-hydrate before anyone can see an image.

I am not sure if React will be cool with us creating a string `onload` attribute, but let's find out!  Otherwise we can do an inline script in the page.  Or we can hack it on the server using dangerouslySetInnerHTML (which is only required on the server).

ianb:
This duplicates some of the React logic that shows the clip image when it loads, but will run even if the bundle doesn't load or there's other Javascript errors.
Incidentally fixes #2651

### DIFF
--- a/server/src/pages/shot/page.js
+++ b/server/src/pages/shot/page.js
@@ -2,5 +2,19 @@ const { Page } = require("../../reactruntime");
 
 exports.page = new Page({
   dir: __dirname,
-  viewModule: require("./view.js")
+  viewModule: require("./view.js"),
+  extraBodyJavascript: `
+(function () {
+  var el = document.getElementById("clipImage");
+  if (el) {
+    el.addEventListener("load", function () {
+      el.style.display = 'inline';
+      var spinner = document.getElementById("spinner");
+      if (spinner) {
+        spinner.style.display = 'none';
+      }
+    }, false);
+  }
+})();
+`
 });
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -42,7 +42,10 @@ class Clip extends React.Component {
       console.warn("Somehow there's a shot without an image");
       return null;
     }
-    let node = <img style={{height: "auto", width: clip.image.dimensions.x + "px", maxWidth: "100%", display: this.state.imageDisplay}} ref="clipImage" src={ clip.image.url } alt={ clip.image.text } onLoad = { this.onImageLoaded.bind(this) } />;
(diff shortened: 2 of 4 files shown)


server/src/pages/shot/page.js
```diff
@@ -2,5 +2,20 @@ const { Page } = require("../../reactruntime");
...
 exports.page = new Page({
   dir: __dirname,
-  viewModule: require("./view.js")
+  viewModule: require("./view.js"),
+  extraBodyJavascript: `
+(function () {
+  var el = document.getElementById("clipImage");
+  if (el) {
+    el.addEventListener("load", function () {
+      console.log("fired the fire baby");
```
pdehaan: bonus `console.log()`?

ianb: Oops!  I could have sworn I removed that.  Opening #2804

PR #2803, opened 2017-05-03
Actually fix #2796, disable all active A/B tests

### LINKED ISSUE #2796: Disable all active A/B experiments
We have some active experiments set, but aren't using them.

ianb:
(Do it right this time)

* Document tests are finished
* Remove code referring to autoOpenSharePanel
* Remove code and styles referring to styleMyShotsButton
* Put all old tests in deprecatedTests

### DIFF
--- a/addon/webextension/selector/uicontrol.js
+++ b/addon/webextension/selector/uicontrol.js
@@ -815,10 +815,6 @@ this.uicontrol = (function() {
       return;
     }
     addHandlers();
-    // FIXME: self.options is gone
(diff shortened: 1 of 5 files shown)


addon/webextension/selector/uicontrol.js
```diff
@@ -816,9 +816,6 @@ this.uicontrol = (function() {
     }
     addHandlers();
     // FIXME: self.options is gone
```
pdehaan: Is this comment still valid with the freshly deleted code?

PR #2820, opened 2017-05-05
add landing page ui

johngruen:


dannycoates:
@johngruen are we planning on keeping the style in view.js?

johngruen:
@dannycoates no, i'm gonna break this up at some point. there's some state that needs to get added.

PR #2825, opened 2017-05-05
Fix another perf issue in latest Firefox export

6a68:
Remove the code that toggles the button state when a tab's URL changes. This will hopefully address perf regressions preventing our latest code bundle from landing in Firefox.

Also pull in the updated SVGs per jgruen's request


6a68:
got r+ from ian in IRC, merging once tests are green

PR #2832, opened 2017-05-09
Add Firefox release checklist to export docs

6a68:


6a68:
inspired in part by https://github.com/mozilla-services/screenshots/issues/2097

we could have a similar template checklist for server releases, too

PR #2847, opened 2017-05-16
DO NOT MERGE Fix #2843, lazily load code into the background page

### LINKED ISSUE #2843: Lazy load our background page code
From https://bugzilla.mozilla.org/show_bug.cgi?id=1361792#c31 :

> After some further investigation, it looks like a lot of the startup responsiveness issues are caused by loading a huge amount of code into the Screenshots background page at startup. It would help a lot if the add-on lazily loaded its background page code when first needed (something like require.js might be a good idea), and avoided touching extension APIs before absolutely necessary.

It looks like this is taking 150-200ms at startup

ianb:
This takes a minimal approach, loading scripts using the script tag on the first button click.

### DIFF
--- a/addon/webextension/README.md
+++ b/addon/webextension/README.md
@@ -7,7 +7,7 @@ This contains all the files for the WebExtension.  Most files are not "built", b
   1. The background page, in `background/`
   2. The selector content worker, in `selector/`
   3. The site helper that communicates with the Firefox Screenshots website, in `site-helper.js`
-- The background process is loaded according to a list in `manifest.json.template` – if there are load order dependencies, they must be manually managed with that list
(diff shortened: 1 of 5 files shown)


ianb:
It occurs to me that this might integrate badly with wantsauth and the sitehelper – sitehelper could be loaded when the background page is not, and so if it tries to communicate that might fail.

6a68:
> It occurs to me that this might integrate badly with wantsauth and the sitehelper – sitehelper could be loaded when the background page is not, and so if it tries to communicate that might fail.

@ianb what additional steps would be needed to make this fully functional?

ianb:
I could make the background page load itself on any message. It wouldn't be too hard to do.

ianb:
Closing, as this does not seem to help much

PR #2857, opened 2017-05-18
Fix #2837, add some more details about Sentry to the metrics documentation

### LINKED ISSUE #2837: Make METRICS.md more explicit about Sentry
There's some information in METRICS about Sentry, but I'd like to expand it to say exactly what information we collect, give an example, and some language about data we avoid.  I think [this](https://sentry.prod.mozaws.net/operations/pageshot-dev/issues/394500/) is a reasonable example of what the WebExtension catches.

ianb:
/cc @fmarier @bsmedberg


fmarier:
Thanks for documenting this explicitly.

6a68:
@bsmedberg Haven't gotten a review from you yet; if you'd like to suggest additional changes, please open a bug.

PR #2864, opened 2017-05-19
Fix #2734, make protocol configurable with EXPECT_PROTOCOL=https

### LINKED ISSUE #2734: Make protocol configurable
Right now we're picking the protocol up from `X-Forwarded-Proto`.  This generally works fine, but we've seen at least one problem in production, and in production we really the protocol to be hardcoded.

I would propose that the configuration for this would force a protocol, but we leave the detection in place when the configuration isn't set.  I think the detection is helpful for development and ad hoc deployments.

ianb:
I don't know if this will actually fix our problems, but from everything I can tell it should.  `req.protocol` is a getter, but its [logic is fairly simple](https://github.com/expressjs/express/blob/c087a45b9cc3eb69c777e260ee880758b6e03a40/lib/request.js#L306-L320).  `trust` is created [here](https://github.com/expressjs/express/blob/c087a45b9cc3eb69c777e260ee880758b6e03a40/lib/utils.js#L232-L251) with a `val` of `true`. 

Once applied we need to update our environmental config.

### DIFF
--- a/server/src/config.js
+++ b/server/src/config.js
@@ -28,6 +28,13 @@ var conf = convict({
     env: "CONTENT_ORIGIN",
     arg: "contentOrigin"
   },
+  expectProtocol: {
+    doc: "Treat all incoming requests as using this protocol, instead of defaulting to http: or detecting from X-Forwarded-Proto",
+    format: String,
+    default: "",
+    env: "EXPECT_PROTOCOL",
+    arg: "expectProtocol"
+  },
   localhostSsl: {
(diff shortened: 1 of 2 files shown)


6a68:
@relud note that we'll need to add EXPECT_PROTOCOL=https to our environment config

relud:
would it break anything if i added it right now, and a version before this patch was released?

6a68:
@relud Good idea! I can't imagine adding an unused env var could hurt anything

relud:
okay, i'll add it right away, so we don't have to worry about it

Issue #2877, opened 2017-05-23
lazybutton metabug


6a68:
Our WebExtension performance workaround for Firefox 55 involves moving UI code (toolbar button, context menu) into bootstrap.js, and only initializing the WebExtension after the initial click.

NB: there may be unexpected blockers not on this checklist

Here's a checklist of _known_ issues:
- [ ] implement l10n in bootstrap.js for button, context menu (PR #2869)
- [ ] create skeleton toolbar button in bootstrap & remove from webext
- [ ] figure out how to send over tab ID on click (or just show overlay in focused tab/window)
- [ ] check screenshots-enabled prefs before creating toolbar button, context menu
- [ ] context menu: restore URL filtering removed from the webextension
- [ ] toolbar button: restore first run button icon if not onboarded (need to move onboarding state into a pref)
- [ ] figure out if context menu needs `accesskey` attribute
- [ ] get a lazybutton review from Mossop or other Gecko expert/superreviewer

6a68:
Closing this one as well, looks like we aren't going with this approach after all

PR #2878, opened 2017-05-23
Move button and context menu code into bootstrap, inline icon svg

6a68:


addon/bootstrap.js
```diff
@@ -51,11 +57,21 @@ function startup(data, reason) { // eslint-disable-line no-unused-vars
   prefObserver.register();
   addonResourceURI = data.resourceURI;
   // eslint-disable-next-line promise/catch-or-return
-  appStartupPromise.then(handleStartup);
+  appStartupPromise.then(initUI);
+}
+
+function initUI() {
+  initButton();
+  // TODO: check the pref state, then either init or shutdown the context menu
```
6a68: added to todo list in #2877

addon/bootstrap.js
```diff
@@ -77,7 +93,21 @@ function shouldDisable() {
...
 
+// the postmessage channel opened once the webextension is started
+let webExtensionPort;
+function onConnect(port) {
+  webExtensionPort = port;
+}
+
+function onClick() {
+  if (!webExtensionPort) { return; } // just do nothing till the port is ready
+  // TODO: figure out how to send the tab ID over
```
6a68: added to todo list in #2877

addon/bootstrap.js
```diff
@@ -77,7 +93,21 @@ function shouldDisable() {
...
+  webExtensionPort.postMessage({ content: "click" });
+}
+
+
 function handleStartup() {
+  // TODO: check the pref _before_ calling into ExtensionsUtils.
```
6a68: this is an older todo, so it might make sense to fix directly in master. filed #2879

addon/bootstrap.js
```diff
@@ -130,3 +164,68 @@ function handleMessage(msg, sender, sendReply) {
     return true;
   }
 }
+
+let buttonStyleSheet;
+
+// TODO: deal with toggling the prefs on/off
```
6a68: duplicate todo, I guess. tracked in #2877

addon/bootstrap.js
```diff
@@ -130,3 +164,68 @@ function handleMessage(msg, sender, sendReply) {
...
+  },
+  observe: function(aSubject, aTopic, aData) {
+    const document = aSubject.wrappedJSObject.menu.ownerDocument;
+    let item = document.getElementById("context-screenshots");
+    if (!item) {
+      item = document.createElement("menuitem");
+      item.setAttribute("id", "context-screenshots");
+      item.setAttribute("label", "screenshots"); // TODO solve l10n without webextensions :-(
+      item.setAttribute("oncommand", "window.alert('clicked screenshots context menu')");
+      // TODO: add an "accesskey" attribute?
```
6a68: added to #2877

ianb: Would that be a hotkey like asked for in https://bugzilla.mozilla.org/show_bug.cgi?id=1365757 ? Or just a key specific to the context menu?

addon/webextension/background/main.js
```diff
@@ -295,5 +179,17 @@ this.main = (function() {
     return startSelectionWithOnboarding(sender.tab);
   });
 
+  // Open a port to let the embedding legacy addon tell us when a button is pressed
+  // TODO: do we need to specify the extensionID in the connect call?
```
6a68: stale todo - the console.log is fired when the button is clicked repeatedly, so this code does work

addon/webextension/background/main.js
```diff
@@ -84,113 +69,12 @@ this.main = (function() {
     });
   }
 
-  function shouldOpenMyShots(url) {
-    return /^about:(?:newtab|blank)/i.test(url) || /^resource:\/\/activity-streams\//i.test(url);
-  }
-
-  browser.browserAction.onClicked.addListener(catcher.watchFunction((tab) => {
```
ianb: In lazy-background I changed this to a function, `main.onClicked = ...` – then I just changed how it was invoked. I think the same would be appropriate here.

addon/webextension/background/main.js
```diff
@@ -295,5 +179,17 @@ this.main = (function() {
...
 
+  // Open a port to let the embedding legacy addon tell us when a button is pressed
+  // TODO: do we need to specify the extensionID in the connect call?
+  // TODO: as soon as the port exists, send a "hello" message over, so the embedding
+  // addon can get a reference to the port.
+  const port = browser.runtime.connect({name: "screenshots-legacy-connection"});
+  port.onMessage.addListener((message) => {
+    if (message.content === "click") {
+      console.log("got a click signal from the legacy addon");
+      // TODO: figure out how to get the tabID of the clicked tab
```
ianb: Something like `browser.tabs.query({active: true}).then(...)`

addon/webextension/background/main.js
```diff
@@ -84,113 +69,12 @@ this.main = (function() {
...
-    contexts: ["page"],
-    documentUrlPatterns: ["<all_urls>"]
-  }, () => {
-    // Note: unlike most browser.* functions this one does not return a promise
-    if (browser.runtime.lastError) {
-      catcher.unhandled(new Error(browser.runtime.lastError.message));
-    }
-  });
-
-  browser.contextMenus.onClicked.addListener(catcher.watchFunction((info, tab) => {
```
ianb: This also could become `main.onClickedContextMenu`

6a68:
The existing changes in the lazybutton branch were rebased into the one commit in this branch, so that's why git is showing conflicts. I can just force-push this branch when it's good to land, and no code should be lost (though existing branches will be orphaned)

addon/bootstrap.js
6a68: I noticed an 'accesskey' attribute in the pocket code i borrowed from. Looking at MDN, it seems this attribute just underlines a specific character in the string: https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/accesskey

6a68:
Yeah, I can bring back the deleted code. For some reason, commenting it out with multiline comments was causing the build to break, which was the original motivation to delete the handlers.

6a68:
looks like we are not going in this direction after all 😅

PR #2883, opened 2017-05-24
Add more traditional localization to Screenshots

ianb:
A different attempt. Simpler really.

Still TODO: don't modify chrome.manifest inline, maybe do the entire thing with jar.mn, and integrate with the build system in some fashion.  Works with `./bin/run-addon --bootstrap`


Pike:
This looks a lot less scary.

The properties parser is still brittle, though, and it won't parse all valid .properties files correctly.

Also, string bundles on startup are brittle, which locale you get depends on startup order. Bug 1005640 is an example.

6a68:
I think we can probably close this without merging, since the Talos numbers for the original webextension code are looking much better?

PR #2921, opened 2017-05-30
Update ignores

ianb:


6a68:
Works for me locally: the buildsettings.js.template and .eslintrc.js aren't showing up in a fresh export.

Issue #2897, opened 2017-05-28
[RTL] welcome presentation is misaligned


yarons:
I've clicked the Firefox Screenshots button on the top left side of the screen (Firefox RTL) and a welcome presentation appeared.

First thing I noticed was: It's not well aligned, the periods appear at the end of the line and the Hebrew and English mixed sentences are incomprehensible.

The arrow should be facing the other way around, the dots should be from right to left instead of the current order and the Skip button should be on the right.

Some screenshots attached:
<img width="768" alt="screen shot 2017-05-28 at 15 53 50" src="https://cloud.githubusercontent.com/assets/406826/26529171/d8645cf8-43c3-11e7-8664-c66fc7eebd70.png">

Next slide, only in Persian this time:
![photo_2017-05-28_16-40-41](https://cloud.githubusercontent.com/assets/406826/26529209/7efadee8-43c4-11e7-9152-7205e6713607.jpg)

Here, the window in the background does no reflect an RTL Firefox but an LTR:
<img width="795" alt="screen shot 2017-05-28 at 15 55 42" src="https://cloud.githubusercontent.com/assets/406826/26529172/dbb76cc4-43c3-11e7-83a6-74f6ee7bb9e5.png">

tomer:
I'd suggest having a different bug for graphics, as other locales (such as french) could benefit from this as well.

johngruen:
agree with @tomer since this seems like a different issue...I think this is now resolved lastest master: 


<img width="937" alt="screen shot 2017-05-31 at 2 44 35 pm" src="https://cloud.githubusercontent.com/assets/3323249/26632350/29ec53ca-4610-11e7-9e51-f58bd8369957.png">
<img width="905" alt="screen shot 2017-05-31 at 2 44 40 pm" src="https://cloud.githubusercontent.com/assets/3323249/26632340/1aa9f3ea-4610-11e7-86be-aacc7ae7f1ae.png">

johngruen:
@yarons, how does this look to you ^
<img width="816" alt="screen shot 2017-05-31 at 2 44 46 pm" src="https://cloud.githubusercontent.com/assets/3323249/26632417/6b4fe714-4610-11e7-8ed6-f4f4609e0973.png">
<img width="525" alt="screen shot 2017-05-31 at 2 47 47 pm" src="https://cloud.githubusercontent.com/assets/3323249/26632418/6b53b600-4610-11e7-9479-8ecace625bfc.png">

[...]

Issue #2898, opened 2017-05-28
[RTL] Inteface is reversed


yarons:
When trying to take screenshots with the tool I've noticed that the message that appears on the shade is reversed:
<img width="558" alt="screen shot 2017-05-28 at 15 56 22" src="https://cloud.githubusercontent.com/assets/406826/26529233/f5baad38-43c4-11e7-939b-d8290b538a28.png">

johngruen:
@yarons, can you clarify what's incorrect here? The RTL copy looks correct to me

arashm:
@johngruen The dot at the end of the sentence is on the right, that's a sign that the interface is LTR. It should be at the end of the sentence, on the left.

PR #2932, opened 2017-05-30
Set lang and dir to html tags

### LINKED ISSUE #2928: Set lang and dir in UI
We should set `<html lang="lang-code" dir="dir">` in the add-on.

The values are available from [predefined variables](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Predefined_messages).  To actually get them we need `browser.i18n.getMessage("@@bidi_dir")` and `browser.i18n.getMessage("@@ ui_locale")`.

There are three HTML documents that need to be updated: two in `ui.js` (for the selection and preselection iframe), and one in `slides.js`.

arashm:
Fixes #2928, for fixing #2897 we still need to reverse and move the arrows.
Not sure if this is the best way to do this since I'm copying the same code into multiple places, but I also couldn't find a utility file to add a function there maybe.

### DIFF
--- a/addon/webextension/onboarding/slides.js
+++ b/addon/webextension/onboarding/slides.js
@@ -46,6 +46,8 @@ this.slides = (function() {
           doc.documentElement
         );
         doc.addEventListener("keyup", onKeyUp);
+        doc.documentElement.dir = browser.i18n.getMessage("@@bidi_dir");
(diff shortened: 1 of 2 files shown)


arashm:
It also fixes the  #2898.

Issue #2880, opened 2017-05-24
"Download", "Share" and "Delete" buttons do not loose their highlight after each respective action is completed


SoftVision-CiprianMuresan:
[Affected versions]:
- Firefox 55 and up
- Screenshots Dev (7.0.11976)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" is set to "false" and latest Firefox Screenshots (7.0.11976) dev is installed on a new clean profile.
- At least one shot is saved.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to the My Shots page.
2. Hover one of the saved shots and click the "Delete" button.
3. Click the "Cancel" button and hover the same saved shot.
4. Observe the "Delete" button state.

[Expected result]:
- The button is no longer highlighted.

[Actual results]:
- The button is highlighted even though the mouse is not hovering it.

[Notes]:
- On Mac, the issue is reproducible only for the "Download" button.
- For the "Share" button the panel needs to be dismissed in order to reproduce the issue.
- Clicking anywhere else on the page will change the focus and the button will no longer be highlighted.
- Here is a screen recording with the issue:
![hover states still appear while buttons are not hovered](https://cloud.githubusercontent.com/assets/18304511/26397376/c3b43712-407e-11e7-87de-372883a062d2.gif)

johngruen:
@ianb yeah this is bound up with that other issue; i think we need to go ahead and componentize each card so we can pass state down to the buttons

PR #2940, opened 2017-05-31
Fix #2771, #2880

### LINKED ISSUE #2771: Make Share panels match spec/not overflow the window on shot index page
we need to write a bit of js logic to do the following:
1. ensure that share panels close onMouseOut of the parent shot
2. move the position of the panels around depending on where they are on the page so that they do not get hidden outside the window

niharikak101:
This is a work-in-progress for #2771 and #2880(?).

### DIFF
--- a/server/src/pages/shotindex/view.js
+++ b/server/src/pages/shotindex/view.js
@@ -52,8 +52,7 @@ class Body extends React.Component {
   renderShots() {
     let children = [];
     for (let shot of this.props.shots) {
-      let downloadUrl = this.props.downloadUrls[shot.id];
-      children.push(this.renderShot(shot, downloadUrl));
+      children.push(<Card shot={shot} downloadUrl={this.props.downloadUrls[shot.id]} abTests={this.props.abTests} clipUrl={shot.urlDisplay} isOwner={this.props.isOwner} staticLink={this.props.staticLink} isExtInstalled={this.props.isExtInstalled} />);
     }
 
     if (children.length === 0) {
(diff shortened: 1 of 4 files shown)


server/src/pages/shotindex/view.js
```diff
@@ -52,8 +52,7 @@ class Body extends React.Component {
   renderShots() {
     let children = [];
     for (let shot of this.props.shots) {
-      let downloadUrl = this.props.downloadUrls[shot.id];
-      children.push(this.renderShot(shot, downloadUrl));
+      children.push(<Card shot = {shot} downloadUrl = {this.props.downloadUrls[shot.id]} abTests={this.props.abTests} clipUrl={shot.urlDisplay} isOwner={this.props.isOwner} staticLink={this.props.staticLink} isExtInstalled={this.props.isExtInstalled} />);
```
ianb: you shouldn't have spaces around the `=` here, the style is more like HTML

server/src/pages/shotindex/view.js
```diff
@@ -110,19 +109,64 @@ class Body extends React.Component {
...
+      document.body.classList.remove("search-results-loading");
     }
-    return "square";
   }
 
-  renderShot(shot, downloadUrl) {
+}
+
+class Card extends React.Component {
+  constructor(props) {
```
ianb: You don't need any constructor (here you are just delegating to the superclass, but that's what happens automatically when a subclass doesn't define a method)

niharikak101:
Update:
1. Changes the position of Share Panels if they're not in the viewport.
2. Closes the share panels onMouseLeave of the share panels, but since we are keeping them open now, shall I remove this?

ianb:
Yes, I think remove onMouseLeave. The approach we talked about earlier should work – have the ShareButton inform its parent (via some handler passed through props) about its open/close state, then have the parent shot tile add or remove a class (e.g., `share-panel-open`) based on that state. Then change the `something:hover` selector to be `something:hover, something.share-panel-open`

niharikak101:
Update:
1. Changes the position of Share Panels if they aren't in the viewport (improved)
2. Makes sure that the opened share panels remain open when they aren't being hovered on

server/src/share-buttons.js
```diff
@@ -116,7 +117,21 @@ class ShareButtonPanel extends React.Component {
     </div>;
   }
 
+  changePanelPosition() {
+    let el = this.refs.share;
+    let rect = el.getBoundingClientRect();
+    if (!(rect.right <= (window.innerWidth || document.documentElement.clientWidth)) && !(rect.bottom <= (window.innerHeight || document.documentElement.clientHeight))) {
+      this.setState({left: -140, top: -385});
+    }
```
ianb: Don't the two conditionals below handle the same thing as this one check?

PR #2944, opened 2017-05-31
Consistent button label

6a68:
IIRC we decided not to localize the button label, but I hard-coded the string "Firefox Screenshots" in background/main, yet also localized it in the manifest (probably in a flurry of l10n edits).

Ian mentioned the browser test was inconsistent about the button label text; this is certainly why.

Browser test updated, seems to work consistently for me locally.


ianb:
I'd like to make sure this is the tooltip @johngruen wants (but r+ the code)

Issue #2957, opened 2017-06-02
No shots can be saved in the same session after "Out of order" error is displayed on Nightly and Beta


SoftVision-CarmenFat:
[Notes]:
- The error is not displayed anymore if a new screenshots session is started (ESC key or "Close button").
- On the Release version 53.0.3, this issue is not reproducible: the next error is displayed "Index or size is negative or greater than the allowed amount". After resizing the area and clicking the "Save" button, the shot is saved and page preview is opened in new tab.
- On the Pilot version, the area with height/width reduced to the maximum is saved and the page preview is displayed, even though there is no image displayed.

[Affected versions]:
- Firefox Nightly 55.0a1
- Firefox Beta 54.0b12
- Screenshots v. 8.2.1970
- Screenshots from latest Nightly.

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have the latest Nightly build installed and create a new profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to a website (www.wikipedia.org).
2. Click the Screenshots button from the main browser toolbar.
3. Select an area with height/width reduced to the maximum.
4. Click on "Save" button.
5. Resize the area horizontally/vertically.
6. Click the "Save" button and observe what happens next.

[Expected result]:
- The user is redirected to the saved shot preview page.

[Actual result]:
- "Out of order. Sorry! We couldn't save your shot. Please try again later" error is displayed.

[Additional Notes]:
- Please see the following screencast:

![out of order](https://cloud.githubusercontent.com/assets/26000308/26731907/460c01ae-47be-11e7-9741-21b9aea82707.gif)

ianb:
Very helpful bug report, this uncovered a number of issues!

SoftVision-CarmenFat:
I have retested this issue with latest Screenshots (dev build 06.06.2017) and the issue is no longer reproducible: 
- "Your selection is too small" error is displayed after clicking the "Save" button for a zero area selection;
- After displaying the error above, the "Save" button regains its functionality if the selected area is resized.

PR #2961, opened 2017-06-02
Save resilience

ianb:
This deals with several issues from #2957:

* Log weird data URLs better
* Fix the Save button staying stuck disabled
* Fix changing the clip after an unsuccessful save
* Give an error when you have zero-area selection

### DIFF
--- a/locales/en-US/webextension.properties
+++ b/locales/en-US/webextension.properties
@@ -27,6 +27,8 @@ loginErrorDetails = We couldn’t save your shot because there is a problem with
 unshootablePageErrorTitle = We can’t screenshot this page.
 unshootablePageErrorDetails = This isn’t a standard Web page, so you can’t take a screenshot of it.
 selfScreenshotErrorTitle = You can’t take a shot of a Firefox Screenshots page!
+# Fired when someone makes a zero-width or zero-height selection
+emptySelectionErrorTitle = Your selection is too small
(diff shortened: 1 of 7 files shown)


locales/en-US/webextension.properties
```diff
@@ -27,6 +27,8 @@ loginErrorDetails = We couldn’t save your shot because there is a problem with
 unshootablePageErrorTitle = We can’t screenshot this page.
 unshootablePageErrorDetails = This isn’t a standard Web page, so you can’t take a screenshot of it.
 selfScreenshotErrorTitle = You can’t take a shot of a Firefox Screenshots page!
+# Fired when someone makes a zero-width or zero-height selection
+emptySelectionErrorTitle = Your selection is too small
```
6a68: @flodolo We're aiming for 55, but have a new string to add. Is it possible to get strings uplifted after 55 graduates to Beta?

flodolo: Localization happens in GitHub, outside of mozilla-central, so you're not bound to trains: the content that you land in mozilla-central is not exposed to localizers, so no problem with uplifts.

Having said that, the sooner you land, the better. And if you're planning to enable in 55, website's localization is getting kind of late.

ianb: This is a very obscure error: the user has the resize the selection down to zero height or width to encounter it, so I think if it doesn't get localized the impact will be very small.

Issue #2918, opened 2017-05-30
cancel,download and save button is not visible


6a68:
Copied over from https://bugzilla.mozilla.org/show_bug.cgi?id=1366335

Build ID   : 20170518175637
User Agent : Mozilla/5.0 (X11; Linux x86_64; rv:54.0) Gecko/20100101 Firefox/54.0

Steps to reproduce :

1. Enable Firefox Screenshots from about:config
2. Select a small portion of the current window by mouse to take screenshot
3. Now drag the whole selected area to the right side of the window(drag as much as 
   you can)

Actual Result:    You will see that the cancel,download and save button is not 
                  visible

Expected Result : These buttons should always be present in-front of the eyes of 
                  the user

![](https://bug1366335.bmoattachments.org/attachment.cgi?id=8869532)

PR #2971, opened 2017-06-06
Change button position if not in viewport

niharikak101:
Fixed them to the top as a placeholder while I figure out the best way to reposition them.
Fix for #2918

### DIFF
--- a/addon/webextension/selector/ui.js
+++ b/addon/webextension/selector/ui.js
@@ -429,7 +429,6 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
       } else {
         this.el.classList.remove("left-selection");
       }
-
       this.el.style.top = (pos.top - bodyRect.top) + "px";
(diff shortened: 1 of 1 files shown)


addon/webextension/selector/ui.js
```diff
@@ -541,6 +553,13 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
       return false;
     },
 
+    isElementInViewport(el) {
+      let rect = el.getBoundingClientRect();
+      if (rect.right <= (window.innerWidth || document.documentElement.clientWidth)) {
```
niharikak101: Found a reference for clientWidth: https://www.w3.org/TR/cssom-view-1/#dom-element-clientwidth
I think `document.documentElement.clientWidth` is specefically only used for IE8 and earlier versions because they don't support `window.innerWidth`, so maybe we could remove it entirely?
Ref: https://www.w3schools.com/jsref/prop_win_innerheight.asp

PR #2973, opened 2017-06-06
Fix #2843, lazily load code into the background page

### LINKED ISSUE #2843: Lazy load our background page code

ianb:
This takes a minimal approach, loading scripts using the script tag on the first button click.
Also creates and intercepts the contextMenu action, and any incoming communication
Reads onboarding flag and changes icon as necessary

Close reading of this would be welcome! It's a pretty substantial change, more than the lines of code would imply.

### DIFF
--- a/addon/webextension/background/README.md
+++ b/addon/webextension/background/README.md
@@ -1,10 +1,11 @@
 These are all the files/modules for the background worker.  This is the higher-privileged WebExtension worker that has no interface.
 
+- `startBackground.js` handles the browserAction click event and context menu, and loads everything else on the first button of context menu click.
 - `analytics.js` handles sending events to the server for use with GA.
 - `auth.js` handles the initial authentication call, and handles subsequent requests using that authentication.  It also informs `analytics.js` of any A/B tests so those can be submitted.
 - `clipboard.js` handles copying to the clipboard.
 - `communication.js` handles communication with the content workers.
 - `deviceinfo.js` gets information about this device that is used in some logging and analytics.
 - `selectorLoader.js` handles loading the selector worker, and contains the list of files that the selector loads.
-- `main.js` loads and launches everything.
+- `main.js` launches everything
 - `takeshot.js` takes the shot the content worker has constructed, sends it to the server, and handles copying and opening the tab.
--- a/addon/webextension/background/startBackground.js
+++ b/addon/webextension/background/startBackground.js
@@ -1,0 +1,99 @@
+/* globals browser, main, communication */
+/* This file handles:
+     browser.browserAction.onClicked
+     browser.contextMenus.onClicked
+     browser.runtime.onMessage
+   and loads the rest of the background page in response to those events, forwarding
+   the events to main.onClicked, main.onClickedContextMenu, or communication.onMessage
+*/
+
+this.startBackground = (function() {
+  const backgroundScripts = [
+    "log.js",
+    "makeUuid.js",
+    "catcher.js",
+    "background/selectorLoader.js",
+    "background/communication.js",
+    "background/auth.js",
+    "background/senderror.js",
+    "build/raven.js",
+    "build/shot.js",
+    "background/analytics.js",
+    "background/deviceInfo.js",
+    "background/takeshot.js",
+    "background/main.js"
+  ];
+
+  browser.browserAction.onClicked.addListener((tab) => {
+    loadIfNecessary().then(() => {
+      main.onClicked(tab);
+    }).catch((error) => {
+      console.error("Error loading Screenshots:", error);
+    });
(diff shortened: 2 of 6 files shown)


addon/webextension/background/startBackground.js
```diff
@@ -0,0 +1,94 @@
...
+    "background/deviceInfo.js",
+    "background/takeshot.js",
+    "background/main.js"
+  ];
+
+  browser.browserAction.onClicked.addListener((tab) => {
+    loadIfNecessary().then(() => {
+      main.onClicked(tab);
+    }).catch((error) => {
+      console.error("Error loading Screenshots:", error);
```
6a68: Hmm. As of right now, if `loadIfNecessary` fails, the user will see "Error loading Screenshots" in the console, and we won't be able to send the failure to Sentry (since the raven.js script may or may not have loaded).

Is there any way we could inline a Sentry ping inside this file, so that we can detect worst-case failure scenarios while gradually rolling out a new release? If we take that approach, it would be nice to add a `catch` inside `loadIfNecessary`, to avoid copying the error handling code around.

Also: if loading scripts fails, does it even make sense to keep the Screenshots toolbar button and context menu item available? Should we just tear down the entire feature?

addon/webextension/background/startBackground.js
```diff
@@ -0,0 +1,94 @@
...
+  browser.contextMenus.onClicked.addListener((info, tab) => {
+    loadIfNecessary().then(() => {
+      main.onClickedContextMenu(info, tab);
+    }).catch((error) => {
+      console.error("Error loading Screenshots:", error);
+    });
+  });
+
+  // Note this duplicates functionality in main.js, but we need to change
+  // the onboarding icon before main.js loads up
```
6a68: Instead of copying the code around, what if main.js exposed a callback (um...`onOnboarding`? idk) that it called when the onboarding view was shown + dismissed? Would be nice to avoid copying code around

addon/webextension/background/startBackground.js
```diff
@@ -0,0 +1,94 @@
+/* globals browser, main, communication */
+/* This file handles browser.browser.onClicked, and loads the rest of the background
```
6a68: typo: `browser.browserAction.onClicked`

addon/webextension/background/README.md
```diff
@@ -1,10 +1,11 @@
 These are all the files/modules for the background worker.  This is the higher-privileged WebExtension worker that has no interface.
 
+- `startBackground.js` handles the click event and loads everything else on the first button click.
```
6a68: maybe "first button or context menu click"

addon/webextension/background/startBackground.js
6a68: Maybe also mention that the context menu click can initiate loading as well, just for consistency

addon/webextension/background/startBackground.js
```diff
@@ -0,0 +1,94 @@
...
+    return true;
+  });
+
+  let loadedPromise;
+
+  function loadIfNecessary() {
+    if (loadedPromise) {
+      return loadedPromise;
+    }
+    loadedPromise = Promise.resolve()
```
6a68: missing semicolon

addon/webextension/background/startBackground.js
ianb: If main.js exposed something, then startBackground would have to track that in its own persistent variable (so it could access the value on startup before loading main), and I think we'd mostly be back where we started

addon/webextension/background/startBackground.js
ianb: Raven.js is one of the larger things to load, I think (96k). So reporting something isn't easy. Or we try to load it while reporting... but then we're in a weird circular state where for some reason loading scripts isn't working, so we try to load a script to report it. It's a bummer, but I'm not sure what else to do.

Though it would be easy to at least pop up an error notification, even without Sentry reporting. That might be better than removing the button, or simply doing nothing. Removing the button could be weird: the button will be there, the person *clicks* on it, and then it's gone! Haha, tricked you! 

I don't think there's a ton of errors that will actually result in a rejected promise either. Script errors do not – unfortunately we have no good way of catching those (before or after this patch). I think only a missing script will trigger `onerror` and hence reject.

PR #2974, opened 2017-06-07
Sets limits in regexes

### LINKED ISSUE #2431: Set regex input matcher lengths
Per checklist APP-INPUTVAL

> * Defines a minimum and maximum length for the data (e.g. {1,25} ) 

https://wiki.mozilla.org/WebAppSec/Secure_Coding_Guidelines#Input_Validation

Replace unbound patterns like + and * with limits like `{1,255}` in validation regexes. The maximums can be set to the database field sizes.

niharikak101:
Fixes #2431

### DIFF
--- a/addon/webextension/selector/documentMetadata.js
+++ b/addon/webextension/selector/documentMetadata.js
@@ -18,7 +18,7 @@ this.documentMetadata = (function() {
   function getOpenGraph() {
     let openGraph = {};
     // If you update this, also update _OPENGRAPH_PROPERTIES in shot.js:
-    let forceSingle = `title type url`.split(/\s+/g);
+    let forceSingle = `title type url`.split(" ");
     let openGraphProperties = `
     title type url image audio description determiner locale site_name video
     image:secure_url image:type image:width image:height
--- a/server/src/ga-activation.js
+++ b/server/src/ga-activation.js
@@ -21,7 +21,7 @@ function _dntEnabled(dnt, ua) {
     // List of Windows versions known to not implement DNT according to the standard.
     var anomalousWinVersions = ['Windows NT 6.1', 'Windows NT 6.2', 'Windows NT 6.3'];
 
-    var fxMatch = ua.match(/Firefox\\/(\\d+)/);
+    var fxMatch = ua.match(/Firefox\\/(\\d{1,10})/);
     var ieRegEx = /MSIE|Trident/i;
     var isIE = ieRegEx.test(ua);
     // Matches from Windows up to the first occurance of ; un-greedily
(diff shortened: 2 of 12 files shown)


addon/webextension/selector/documentMetadata.js
```diff
@@ -18,7 +18,7 @@ this.documentMetadata = (function() {
   function getOpenGraph() {
     let openGraph = {};
     // If you update this, also update _OPENGRAPH_PROPERTIES in shot.js:
-    let forceSingle = `title type url`.split(/\s+/g);
+    let forceSingle = `title type url`.split(/\s{1,225}/g);
```
ianb: We should probably remove the regex here, just do `let forceSingle = "title type url".split(" ");`

addon/webextension/selector/documentMetadata.js
```diff
@@ -27,7 +27,7 @@ this.documentMetadata = (function() {
     article:published_time article:modified_time article:expiration_time article:author article:section article:tag
     book:author book:isbn book:release_date book:tag
     profile:first_name profile:last_name profile:username profile:gender
-    `.split(/\s+/g);
+    `.split(/\s{1,225}/g);
```
ianb: Since this is a static string being split I think we can forgo the limits

addon/webextension/selector/documentMetadata.js
```diff
@@ -61,7 +61,7 @@ this.documentMetadata = (function() {
     let properties = `
     card site title description image
     player player:width player:height player:stream player:stream:content_type
-    `.split(/\s+/g);
+    `.split(/\s{1,225}/g);
```
ianb: ditto

server/src/ga-activation.js
```diff
@@ -21,7 +21,7 @@ function _dntEnabled(dnt, ua) {
     // List of Windows versions known to not implement DNT according to the standard.
     var anomalousWinVersions = ['Windows NT 6.1', 'Windows NT 6.2', 'Windows NT 6.3'];
 
-    var fxMatch = ua.match(/Firefox\\/(\\d+)/);
+    var fxMatch = ua.match(/Firefox\\/(\\d{1,4000})/);
```
ianb: This is a Firefox version, so I think we can do something more limited like `\\d{1,10}`

server/src/server.js
```diff
@@ -284,7 +284,7 @@ function decodeAuthHeader(header) {
   /** Decode a string header in the format {deviceId}:{deviceIdSig};abtests={b64thing}:{sig} */
   // Since it's treated as opaque, we'll use a fragile regex
   let keygrip = dbschema.getKeygrip();
-  let match = /^([^:]+):([^;]+);abTests=([^:]+):(.*)$/.exec(header);
+  let match = /^([^:]{1,225}):([^;]{1,225});abTests=([^:]{1,225}):(.{0,225})$/.exec(header);
```
ianb: I suspect abTests could go over 255 characters (it's a base64'd JSON object), so the second-to-last one should be longer than 225. Also maybe you meant 255 instead of 225 ;) Though really they are both entirely arbitrary numbers.

Anyway, like:

`let match = /^([^:]{1,225}):([^;]{1,225});abTests=([^:]{1,1500}):(.{0,225})$/.exec(header);`

Issue #2988, opened 2017-06-07
Add-on UI NITS


johngruen:
- [ ] fix icon
- [ ] fix UI in slides
- [ ] add a bit of pop to the hover selection UI

PR #2989, opened 2017-06-07
update icon

johngruen:
Fixes pt 1 of  #2988


ianb:
I also couldn't see the new highlight icon when testing locally, but I'm pretty sure this is indeed some kind of caching issue. If I load up the URL of the icon (something like `"moz-extension://b4efb261-a824-014c-b1fc-d31f8aa5bf0d/icons/icon-starred-32.svg` – origin changes) then I do see the new icon.  Unfortunately we probably won't see the new icon until the import into Firefox happens.

PR #2992, opened 2017-06-08
hide search

### LINKED ISSUE #2949: Hide search
We should remove the *interface* for search, based on a config.js option.  Search should default to off.

This should only remove the search field from the HTML, not the other logic.  Note that `shotindex/model.js` will have to add a flag to the model based on the configuration, then `shotindex/view.js` will use that to show or hide the search field.

niharikak101:
Fixes #2949

### DIFF
--- a/server/src/pages/shotindex/view.js
+++ b/server/src/pages/shotindex/view.js
@@ -34,11 +34,7 @@ class Body extends React.Component {
         <div className="column-space full-height default-color-scheme">
           <div id="shot-index-header" className="header">
             <h1><a href="/shots">Firefox <strong>Screenshots</strong> <sup>Beta</sup></a></h1>
-            <form onSubmit={ this.onSubmitForm.bind(this) }>
-              <span className="search-label" />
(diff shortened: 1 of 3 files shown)


server/src/pages/shotindex/view.js
```diff
@@ -34,7 +34,7 @@ class Body extends React.Component {
         <div className="column-space full-height default-color-scheme">
           <div id="shot-index-header" className="header">
             <h1><a href="/shots">Firefox <strong>Screenshots</strong> <sup>Beta</sup></a></h1>
-            <form onSubmit={ this.onSubmitForm.bind(this) }>
+            <form onSubmit={ this.onSubmitForm.bind(this) } style={{display: this.state.displaySearch}}>
```
ianb: There's no need for state here, the props can be used alone. Also I'd rather the form just be left out, instead of rendered but hidden.

So if you move the search form to a method, it would just be like `{ this.props.disableSearch ? null : this.renderSearchForm() }`

PR #2993, opened 2017-06-08
hides 'copy image text' option if text capture is disabled

### LINKED ISSUE #2956: "Copy Image Text" option from context menu is available, even if the text capture is disabled
[Affected versions]:
- Nightly 55.0a1
- Sccreenshots 8.2.0

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have the Nightly try build with Screenshots 8.2.0 installed and create a new profile.
- Screenshots Onboarding tour was already visited. 

[Steps to reproduce]:
1. Open the Nightly try build with Screenshots 8.2.0 installed with the profile from prerequisites.
2. Navigate to any website and save a selection.
3. Right click on the image and observe the context menu.

[Expected result]:
- The "Copy Image Text" option is no longer available.

[Actual results]:
- "Copy Image Text" option is still available.

niharikak101:
Fixes #2956

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -47,9 +47,7 @@ class Clip extends React.Component {
     // that displays the image onload, as a backup to make sure the image always
     // gets displayed even if the bundle doesn't load
     return <div ref="clipContainer" className="clip-container">
-      <menu type="context" id="clip-image-context">
-        <menuitem label="Copy Image Text" onClick={this.copyImageText.bind(this)} ></menuitem>
-      </menu>
+      { this.copyTextContextMenu() }
       { this.renderLoader() }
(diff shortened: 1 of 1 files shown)


server/src/pages/shot/view.js
```diff
@@ -80,6 +78,16 @@ class Clip extends React.Component {
     // Allow default action to continue
   }
 
+  textCaptureEnabled() {
```
ianb: `textCaptureEnabled` sounds like a true/false function. `textCaptureIfEnabled` would be a little clearer, but I would just name this `copyTextContextMenu()`.

Also it should return null at the end instead of the implicit return undefined.

PR #3001, opened 2017-06-08
Fix #2991, update npm packages

### LINKED ISSUE #2991: Update npm packages
It's getting time to freshen up.  `npm run update_outdated` will attempt to do the update.

ianb:


package.json
```diff
@@ -17,53 +17,53 @@
...
+    "mozlog": "2.1.1",
     "node-statsd": "0.1.1",
     "nodemon": "1.11.0",
     "nodify-uuid": "0.0.1",
-    "pg": "6.1.5",
+    "pg": "6.2.3",
     "pg-patcher": "0.3.0",
-    "raven": "1.2.1",
-    "raven-js": "3.14.2",
+    "raven": "2.0.2",
```
6a68: Noticed this warning message in the [changelog](https://github.com/getsentry/raven-node/blob/master/History.md#200---5102017) for raven:

> Be aware when upgrading: it is possible that with raven-node 2.0, your node process will shut down from exceptions where it previously did not

Looking at the server code, I don't think this should affect us, but thought I'd mention it anyway

ianb: Good to note, we'll have to keep an eye out!

Issue #2976, opened 2017-06-07
Use Raven 'release' parameter


ianb:
Looking at the docs:

![image](https://user-images.githubusercontent.com/44390/26888281-2fe7b0fc-4b70-11e7-934d-03fa6d7d8611.png)

I think we're adding `version` somewhere, but that might be as useful as release.

ianb:
Fixed by #2975

PR #3002, opened 2017-06-08
2975 normalize sentry data

ianb:


addon/webextension/background/senderror.js
```diff
@@ -81,6 +81,16 @@ this.senderror = (function() {
     });
   };
 
+  // Every copy of Screenshots has its own unique moz-extension:// URL.
+  // Raven doesn't correctly group errors with different URLs, so we use the
+  // raven.js `dataCallback` option to turn the UUID into "screenshots-addon"
+  // before sending.
+  // TODO: do this here or somewhere else, though? like in reportError?
+  function normalizeURL(data) {
+    data.request.url = data.request.url.replace(/moz-extension:\/\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/, "resource://screenshots-addon");
```
ianb: If we were fancy we could actually rewrite it to the github URL, but maybe later. Mostly we need to jam the github revision in somewhere in the add-on, though I guess `https://raw.githubusercontent.com/mozilla-services/screenshots/{version}/addon/webextension/{file}` would work for our releases. I believe Sentry gives you nice links if you give it fetchable URLs.

addon/webextension/background/senderror.js
```diff
@@ -92,20 +102,26 @@ this.senderror = (function() {
       return;
     }
     if (!Raven.isSetup()) {
-      Raven.config(dsn, {allowSecretKey: true}).install();
+      Raven.config(dsn, {allowSecretKey: true, dataCallback: normalizeURL}).install();
     }
-    let exception = new Error(e.message);
+    // NOTE: don't use e.message, as that is a localized string
```
ianb: e.message shouldn't be localized.  popupMessage is used to get a localized string from elsewhere for the popup.

addon/webextension/background/senderror.js
```diff
@@ -92,20 +102,26 @@ this.senderror = (function() {
...
       if (!["name", "message", "stack", "multilineStack", "popupMessage", "version", "sentryPublicDSN", "help"].includes(attr)) {
         rest[attr] = e[attr];
       }
     }
-    rest.stack = e.multilineStack || e.stack;
+    rest.stack = exception.stack;
     Raven.captureException(exception, {
       logger: 'addon',
       tags: {version: manifest.version, category: e.popupMessage},
+      culprit: e.popupMessage,
```
ianb: This probably isn't needed

addon/webextension/background/senderror.js
```diff
@@ -92,20 +102,26 @@ this.senderror = (function() {
...
     for (let attr in e) {
       if (!["name", "message", "stack", "multilineStack", "popupMessage", "version", "sentryPublicDSN", "help"].includes(attr)) {
         rest[attr] = e[attr];
       }
     }
-    rest.stack = e.multilineStack || e.stack;
+    rest.stack = exception.stack;
     Raven.captureException(exception, {
       logger: 'addon',
       tags: {version: manifest.version, category: e.popupMessage},
```
ianb: I think if we change `version` to `release` (and don't put it in `tags`) then I think we also fix #2976

addon/webextension/background/senderror.js
```diff
@@ -92,20 +102,26 @@ this.senderror = (function() {
       return;
     }
     if (!Raven.isSetup()) {
-      Raven.config(dsn, {allowSecretKey: true}).install();
+      Raven.config(dsn, {allowSecretKey: true, dataCallback: normalizeURL}).install();
     }
-    let exception = new Error(e.message);
+    // NOTE: don't use e.message, as that is a localized string
+    let exception = new Error(e.popupMessage);
```
ianb: Also I noticed the errors had no description, which I realize it because we're using popupMessage.  Example: https://sentry.prod.mozaws.net/operations/screenshots-addon-dev/issues/612365/

addon/webextension/background/senderror.js
6a68: oddly the 'culprit' is always set to moz-extension.../_generated_background_page if we don't manually set this

ianb: That's probably location.href, which is always the same, but the top of the stack would be better

addon/webextension/background/senderror.js
6a68: Oh, nice! I had assumed the problem was that Sentry [can't parse](https://github.com/getsentry/raven-js/issues/974) moz-extension URLs, which would explain why it couldn't locate the source. I'll give the raw git approach a try

6a68: The github URL turns out to be more trouble than it's worth - Sentry infers that the module name is the entire path beyond the domain name, which makes it hard to parse the 'culprit' field. We can experiment with better integration later, though.

addon/webextension/background/senderror.js
6a68: Turns out the culprit isn't needed if the stack is transformed before passing to raven

6a68:
@ianb I think this moves things forward a good step, though Github integration isn't quite there yet. Do you want to take another pass, or should I just land it?

PR #3003, opened 2017-06-08
Sentry/investigative fixes

ianb:
A bunch of stuff that came out looking at Sentry reports


addon/webextension/background/senderror.js
```diff
@@ -88,7 +88,7 @@ this.senderror = (function() {
     }
     let dsn = auth.getSentryPublicDSN();
     if (!dsn) {
-      log.warn("Error:", e);
+      log.warn("Screenshots error:", e);
```
6a68: I changed this in my PR, we'll probably get conflicts depending on which lands first. Either message is fine with me.

PR #3011, opened 2017-06-09
2955 startup errors

6a68:


addon/webextension/background/senderror.js
```diff
@@ -73,12 +75,14 @@ this.senderror = (function() {
         message = error.message;
       }
     }
-    browser.notifications.create(id, {
-      type: "basic",
-      // FIXME: need iconUrl for an image, see #2239
-      title,
-      message
-    });
+    if (Date.now() - startTime > 5 * 1000) {
```
ianb: I added #3015 as a followup here

PR #3017, opened 2017-06-12
Fix #3007, start background page if migration is needed

### LINKED ISSUE #3007: lazy-background delays Page Shot migration
The Page Shot migration is done in main.js, so until you use Screenshots it doesn't migrate Page Shot

ianb:


ianb:
There's some server Raven compatibility issues (but only if you have Sentry on), but otherwise tests are running locally okay, CircleCI seems to be having some other issues

addon/webextension/background/startBackground.js
```diff
@@ -69,6 +71,30 @@ this.startBackground = (function() {
     return true;
   });
 
+  // We delay this check (by CHECK_MIGRATION_DELAY) just to avoid piling too
+  // many things onto browser/add-on startup
+  setTimeout(() => {
```
6a68: Instead of just blindly setting a timeout, let's wait for the browser to become idle, via `requestIdleCallback`

addon/webextension/background/startBackground.js
```diff
@@ -69,6 +71,30 @@ this.startBackground = (function() {
...
+        throw new Error("Got no result from getOldDeviceInfo");
+      }
+      if (result.type == "error") {
+        throw new Error(`Error from getOldDeviceInfo: ${result.name}`);
+      }
+    }).catch((error) => {
+      if (error && error.message == "Could not establish connection. Receiving end does not exist") {
+        // Just a missing bootstrap.js, ignore
+      } else {
+        console.error("Screenshots error checking for Page Shot migration:", error);
```
6a68: Should we log this to Sentry?

addon/webextension/background/startBackground.js
6a68: Note that `requestIdleCallback` provides a `timeout` parameter that we can use to guarantee execution by a particular point in time

addon/webextension/background/startBackground.js
ianb: No Sentry in startBackground, because we haven't loaded catcher, senderror, etc.

PR #3019, opened 2017-06-12
Fix 3005: Hides download firefox button

### LINKED ISSUE #3005: [Enhancement] The "Download Firefox" button should not be displayed on Onboarding page, if it's opened in Firefox
[Affected versions]:
- Nightly 55.0a1
- Sccreenshots 9.0.0

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have the latest Nightly build with Screenshots 9.0.0 installed and create a new profile.

[Steps to reproduce]:
1. Open the latest Nightly build with the profile from prerequisites.
2. Click the Screenshots button from browser toolbar.
3. Skip the "Onboarding" tour and observe the page.

[Expected result]:
- The "Download Firefox" button is not displayed.

[Actual results]:
- The "Download Firefox" button is displayed even if the Onboarding page it's opened in Firefox.

niharikak101:
fixes #3005

### DIFF
--- a/server/src/pages/homepage/model.js
+++ b/server/src/pages/homepage/model.js
@@ -1,7 +1,9 @@
 exports.createModel = function(req) {
+  let isFirefox = (/firefox\/\d{1,255}/i).test(req.headers['user-agent']);
   let model = {
     title: "Firefox Screenshots",
-    showMyShots: !!req.deviceId
+    showMyShots: !!req.deviceId,
+    isFirefox
   };
   return model;
 };
--- a/server/src/pages/homepage/view.js
+++ b/server/src/pages/homepage/view.js
@@ -558,6 +558,23 @@ class Body extends React.Component {
     window.location = "/shots";
   }
 
+  renderGetFirefox() {
+    if (this.props.isFirefox) {
+      return null;
+    }
+    return (
+      <a href="https://www.mozilla.org/firefox" className="button primary download-firefox">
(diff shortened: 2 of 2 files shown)


server/src/pages/homepage/model.js
```diff
@@ -1,7 +1,8 @@
 exports.createModel = function(req) {
   let model = {
     title: "Firefox Screenshots",
-    showMyShots: !!req.deviceId
+    showMyShots: !!req.deviceId,
+    userAgent: req.headers['user-agent']
```
ianb: I think it would be better here to do:

```js
let isFirefox = (/firefox\/\d{1,255}/i).test(req.headers['user-agent'])
```

and then put that in as a property of the model. Then the test is simpler in the view.

server/src/pages/homepage/view.js
```diff
@@ -558,6 +558,23 @@ class Body extends React.Component {
     window.location = "/shots";
   }
 
+  renderGetFirefox() {
+    if (this.props.userAgent.search(/firefox\/\d{1,255}/i) === -1) {
```
ianb: The logic would change based on the above comment, but generally I try to put exceptions first, so like:

```js
renderGetFirefox() {
  if (!this.props.isFirefox) {
    return null;
  }
  return ...
}
```

PR #3023, opened 2017-06-12
Bump version to 11.0.0 for dev

ianb:


6a68:
leaving this open so we can add the one string change without having to bump major versions all over again

ianb:
I got a branch going for v10, so master officially goes to 11!

PR #3034, opened 2017-06-14
modifies staticLink to host static assets in CDN

niharikak101:
Fix for #2210

### DIFF
--- a/server/src/config.js
+++ b/server/src/config.js
@@ -274,6 +274,13 @@ var conf = convict({
     default: true,
     env: "DISABLE_SEARCH",
     arg: "disable-search"
+  },
+  cdn: {
+    doc: "CDN URL prefix, e.g. 'https://somecdn.com/mysite'; links will be rewritten as https://somecdn.com/mysite/static/style.css",
+    format: String,
+    default: "",
+    env: "CDN",
+    arg: "cdn"
   }
 });
 
--- a/server/src/linker.js
+++ b/server/src/linker.js
@@ -14,11 +14,11 @@ exports.getGitRevision = function() {
   return gitRevision;
 };
 
-exports.staticLink = function(resource) {
+exports.staticLink = function(req, resource) {
   if (!resource.startsWith("/")) {
     resource = "/" + resource;
   }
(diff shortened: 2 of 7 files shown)


server/src/pages/homepage/model.js
```diff
@@ -1,7 +1,8 @@
 exports.createModel = function(req) {
   let model = {
     title: "Firefox Screenshots",
-    showMyShots: !!req.deviceId
+    showMyShots: !!req.deviceId,
+    staticUrl: req.config.staticUrl
```
ianb: These can all go in `reactrender` – the same way `gitRevision` is set there.

ianb: Also when you are setting it there, you can normalize the URL.  Specifically `staticUrl` shouldn't end with a `/`, and if it's set that way it should be removed (like `staticUrl.replace(/\/*$/, "")`).

server/src/config.js
```diff
@@ -274,6 +274,13 @@ var conf = convict({
     default: true,
     env: "DISABLE_SEARCH",
     arg: "disable-search"
+  },
+  staticUrl: {
+    doc: "URL to hose static assets on CDN",
```
ianb: This doesn't explain enough about the variable.  Maybe something more like "CDN URL prefix, like 'https://somecdn.com/mysite' and links will be rewritten like https://somecdn.com/mysite/static/style.css"

Calling it a CDN might be better, since this isn't where the static files are hosted, but where they are served up –the actual source is still this server.

server/src/linker.js
```diff
@@ -29,7 +29,7 @@ exports.staticLink.simple = function(resource) {
 };
 
 exports.staticLinkWithHost = function(req, resource) {
-  return req.protocol + "://" + req.headers.host + exports.staticLink(resource);
+  return req.protocol + "://" + req.headers.host + `${resource}?rev=${gitRevision}`;
```
ianb: Shouldn't it use `req.cdn` if it's available? With this patch `staticLinkWithHost` and `staticLink` aren't that much different anymore (except of course when the CDN isn't configured)

PR #3035, opened 2017-06-14
change cursor style

niharikak101:
Small change - Go To My Shots on the homepage has a text cursor and not a pointer.


ianb:
Good attention to detail!

PR #3037, opened 2017-06-14
Protect against an empty response

ianb:
This happened in some weird corner case while debugging

### DIFF
--- a/addon/webextension/selector/callBackground.js
+++ b/addon/webextension/selector/callBackground.js
@@ -4,10 +4,10 @@
 
 this.callBackground = function callBackground(funcName, ...args) {
   return browser.runtime.sendMessage({funcName, args}).then((result) => {
-    if (result.type === "success") {
+    if (result && result.type === "success") {
       return result.value;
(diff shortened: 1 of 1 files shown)


addon/webextension/selector/callBackground.js
```diff
@@ -4,10 +4,10 @@
...
   return browser.runtime.sendMessage({funcName, args}).then((result) => {
-    if (result.type === "success") {
+    if (result && result.type === "success") {
       return result.value;
-    } else if (result.type === "error") {
-      let exc = new Error(result.message);
+    } else if (result && result.type === "error") {
+      let exc = new Error(result.message || "Unknown error");
       exc.name = "BackgroundError";
       if ('errorCode' in result) {
```
6a68: I think you'll need to guard against `result` being undefined in these `"foo" in result` clauses, too. They'll throw if result is falsy.

6a68: oh, we're inside a result && check here. Sorry!

Issue #2157, opened 2017-02-09
Design: download for Save visible and Save full page


ianb:
Right now we have a noticeable inconsistency where you can download a selection, but not the other options.

johngruen:
I think this is a stretch based on the current flows i've seen. Would be nice to resolve, but i don't think it blocks launch.

johngruen:
This was my idea...asking design to go ahead and take a look
![select list shot button 1](https://cloud.githubusercontent.com/assets/3323249/24128888/4e2e5e60-0db5-11e7-93c1-85e0f9fd2bcc.png)
![select list shot button 2](https://cloud.githubusercontent.com/assets/3323249/24128889/4e307d94-0db5-11e7-962c-6903b6f63bce.png)

ianb:
I'm having a hard time putting my finger on why, but in this model it feels like "Save full page" is a modifier on the selection action, as opposed to an entirely separate action.  Maybe because it looks like a select box and so it's not obvious that each item is actually an action with immediate effects.

ChenMorpheus:
We tried several mockups and decided to go this proposal since it's more consistent to the current regional crop. In short, when users choose to have a full page shot, we offer them the preview of the final result to decide if download or crop. Please find the link below for the prototype and let us know your thoughts or the feasibility, thanks.

https://mozilla.invisionapp.com/share/BJB0YYB9W#/225402436_2-1_Shot_page_right_side_menu_-2-

![2 5 save fullpage](https://cloud.githubusercontent.com/assets/18019946/24339285/50f6a558-12dd-11e7-99bc-faad5adbc3b1.png)

dannycoates:
Should this still be in Milestone 54?

johngruen:
@ianb is this feasible? would be a good follow on in 56?

ianb:
Outside of the UI implementation itself, implementing download-full/visible-page is very simple. The only blocker to this has been the UI to actually ask for it to be downloaded.

johngruen:
@ianb i guess my question is does the proposal that @ChenMorpheus made in the screenshot above look feasible...if you click the invision link, you can see a stepwise version of the same. password = testpilot (password not intended to be cryptographically secure)

johngruen:
User Agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:55.0) Gecko/20100101 Firefox/55.0
Build ID: 20170529030204

Steps to reproduce:

[...]

Issue #3047, opened 2017-06-16
[9.0.0] Screenshot overlay doesn't work at all on ulmart.ru


SXZ1:
Hello!
Ulmart is some kind of russian Amazon, more info on wiki.
STR:
Nightly 16-06, Screenshot 9.0.0, create new profile, launch it
Open new tab, go to https://www.ulmart.ru/goods/3863051?jd=false
Wait until the web page stops loading
Click on  Screenshot button

ER: Screenshot overlay on top of the ulmart page
AR: No overlay, Screenshot completely broken

SXZ1:
These entries in the browser console seem to be related to Screenshot addon:
![image](https://user-images.githubusercontent.com/26294581/27225878-da8c46d6-52a4-11e7-9b2a-406e11427ea6.png)

6a68:
@SXZ1 Thanks for the bug report! Unfortunately, none of those console
messages were created by Screenshots code.

Looking at the page, there are already 19 (!) iframes in the page. I wonder
if the page has some rules that prevent unknown iframes (like the
Screenshots overlay) from displaying properly.

ianb:
OK, I figured this out.  On that page there's a CSS file with these rules:

```css
body > iframe {
  position: absolute !important;
  clip: rect(1px,1px,1px,1px);
}
```

Both of them render our iframe invisible.

@niharikak101 is in the middle of a PR that would be useful here, adding some rules to [this function](https://github.com/mozilla-services/screenshots/pull/3050/files#diff-88bd153ab5ec987fdce5d6c98f39f31dR79).  Specifically adding `!important` to the position and `clip: auto`.  I'm going to hand it off to her.

PR #3050, opened 2017-06-19
preview shot before saving/downloading full page/visible - WIP

niharikak101:
Still needs a metrics update and some fine-tuning.
#2157


addon/webextension/selector/ui.js
```diff
@@ -322,11 +322,98 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
 
   iframePreSelection.onResize = watchFunction(onResize.bind(iframePreSelection));
 
+  let iframePreview = exports.iframePreview = {
+    element: null,
+    document: null,
+    sizeTracking: {
```
ianb: Huh, I notice the `iframePreSelection` code has size tracking, but I wonder if it really needs it? Unlike selection, we just need an element that is the height and width of the viewport, position:fixed.  The selection iframe does something more complicated that this element won't have to do.

addon/webextension/selector/ui.js
```diff
@@ -322,11 +322,98 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
...
+                      <button class = "highlight-button-cancel"></button>
+                      <button class = "highlight-button-download"></button>
+                      <button class = "highlight-button-save"></button>
+                    </div>
+                  </div>
+                </div>
+              </body>`;
+            installHandlerOnDocument(this.document);
+            const overlay = this.document.querySelector(".preview-overlay");
+            overlay.querySelector(".highlight-button-save").textContent = browser.i18n.getMessage("saveScreenshotSelectedArea");
```
ianb: It might be good to use the code/approach from slides.js – the code that does the i18n substitution could get pulled out and used here.

addon/webextension/selector/ui.js
```diff
@@ -322,11 +322,98 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
...
+          this.element.onload = watchFunction(() => {
+            this.document = this.element.contentDocument;
+            this.document.documentElement.innerHTML = `
+              <head>
+                <style>${substitutedCss}</style>
+                <title></title>
+              </head>
+              <body>
+                <div class="preview-overlay">
+                  <div class = "preview-image">
```
ianb: unnecessary spaces around `=`

addon/webextension/selector/uicontrol.js
```diff
@@ -166,9 +175,18 @@ this.uicontrol = (function() {
       selectedPos = new Selection(
         0, 0,
         width, height);
-      shooter.takeShot("fullPage", selectedPos);
+      captureType = 'fullPage';
+      setState("previewing");
+    },
+    onSavePreview: () => {
+      sendEvent(`capture-preview`, "save-preview-button");
```
ianb: These probably should include `captureType` as part of the event (maybe a label)

ianb: Also of course a METRICS.md update

addon/webextension/selector/ui.js
```diff
@@ -322,11 +322,98 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
...
 
+  let iframePreview = exports.iframePreview = {
+    element: null,
+    document: null,
+    sizeTracking: {
+      windowDelayer: null
+    },
+    display(installHandlerOnDocument, standardOverlayCallbacks) {
+      return new Promise((resolve, reject) => {
+        if (!this.element) {
```
ianb: It would be really nice to put this stuff in a function – we really should have done this when we had 2 iframes, and then 3, and now 4...

OTOH, it might be better to do that separately with a commit that doesn't change any functionality.

niharikak101:
some changes, and a fix for #3047

youwenliang:
@niharikak101 
Just a quick suggestion, the bounce-in animation would look better if you use "ease-in-out" instead of "ease", and change the bounce-in keyframe from 0%->97%->100% to 0%->60%->100%.

something like this: https://codepen.io/youwenliang/pen/MozwRp

PR #3068, opened 2017-06-28
localize server

6a68:
Still debugging some build issues, but the .ftl file is mostly baked. @flodolo comments welcome! We can land the server.ftl file before the rest of the code, if the localizers need to get started now for the 55 release date.

Major outstanding todos:
- figure out what to do about timediffs (likely replace relative times with absolute dates)
- check these strings against the master approved list of strings

### DIFF
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -1,0 +1,214 @@
+// Global phrases shared across pages, prefixed with 'g'
+gMyShots = My Shots
+gHomeLink = Home
+gNoShots
+  .alt = No shots found
+gScreenshotsDescription = Screenshots made simple. Take, save, and share screenshots without leaving Firefox.
+
+// Creating page
+
+// Note: {$title} is a placeholder for the title of the web page
+// captured in the screenshot. The default, for pages without titles, is
+// creatingPageTitleDefault.
+creatingPageTitle = Creating {$title}
+creatingPageTitleDefault = page
+
+// Home page
+
+homePageDescription
+  .content = Intuitive screenshots baked right into the browser. Capture, save and share screenshots as you browse the Web using Firefox.
+homePageButtonMyShots = Go To My Shots
+homePageTeaser = Coming Soon…
+homePageDownloadFirefoxTitle = Firefox
+homePageDownloadFirefoxSubTitle = Free Download
+homePageGetStarted = Get Started
+// Note: do not translate 'Firefox Screenshots' when translating this string
+homePageHowScreenshotsWorks = How Firefox Screenshots Works
+homePageGetStarted = Get Started
+homePageGetStartedDescription = Find the new Screenshots icon on your toolbar. Select it, and the Screenshots menu will appear on top of your browser window.
+homePageCaptureRegion = Capture a Region
+homePageCaptureRegionDescription = Click and drag to select the area you want to capture. Or just hover and click — Screenshots will select the area for you. Like what you see? Select Save to access your screenshot online or the down arrow button to download it to your computer.
+homePageCapturePage = Capture a Page
+homePageCapturePageDescription = Use the buttons in the upper right to capture full pages. The Save Visible button will capture the area you can view without scrolling, and the Save Full Page will capture everything on the page.
+homePageSaveShare = Save and Share
+homePageSaveShareDescription = When you take a shot, Firefox posts your screenshot to your online Screenshots library and copies the link to your clipboard. We automatically store your screenshot for two weeks, but you can delete shots at any time or change the expiration date to keep them in your library for longer.
+homePageLegalLink = Legal
+homePagePrivacyLink = Privacy
+homePageTermsLink = Terms
+homePagePrivacyLink = Cookies
+
+// Leave Screenshots page
+
+leavePageConfirmDelete = Confirm account deletion
+leavePageErrorAddonRequired = You must have Firefox Screenshots installed to delete your account
+leavePageErrorGeneric = An error occurred
+leavePageWarning = This will permanently erase all of your Firefox Screenshots data.
+leavePageButtonProceed = Proceed
+leavePageButtonCancel = Cancel
+leavePageDeleted = All of your screenshots have been erased!
+
+// Not Found page
+
+notFoundPageTitle = Page Not Found
+notFoundPageIntro = Oops.
+notFoundPageDescription = Page not found.
+
+// Shot page
+
+shotPageAlertErrorUpdatingExpirationTime = Error saving expiration
+shotPageAlertErrorDeletingShot = Error deleting shot
+shotPageAlertErrorUpdatingTitle = Error saving title
+shotPageConfirmDelete = Are you sure you want to delete this shot permanently?
+shotPageShareButton
+  .title = Share
+shotPageCopy = Copy
+shotPageCopied = Copied
+shotPageShareFacebook
+  .title = Share on Facebook
+shotPageShareTwitter
+  .title = Share on Twitter
+shotPageSharePinterest
+  .title = Share on Pinterest
+shotPageShareEmail
+  .title = Share link via email
+shotPageShareLink = Get a shareable link to this shot:
+shotPagePrivacyMessage = Anyone with the link can view this shot.
+shotPageCopyImageText
+  .label = Copy image text
+shotPageConfirmDeletion = Are you sure you want to delete this shot permanently?
+// Note: {$timediff} is a placeholder for a future relative time clause like 'in 3 days' or 'tomorrow'
+shotPageExpirationMessage = If you do nothing, this shot will be permanently deleted {$timediff}.
+// Note: {$date} is a placeholder for a localized future date as returned by Date.toLocaleString.
+// For example, in en-US, {$date} could be "7/12/2017, 1:52:50 PM".
+shotPageRestoreButton = restore until {$date}
+shotPageExpiredMessage = This shot has expired.
+// Note: This phrase is followed by an empty line, then the URL of the source page
+shotPageExpiredMessageDetails = Here is the page it was originally created from:
+shotPageDeleteButton
+  .title = Delete this shot
+shotPageAbuseButton
+  .title = Report this shot for abuse, spam, or other problems
+shotPageDownloadShot
+  .title = Download
+shotPageDownload = Download
+// Note: do not translate 'Firefox Screenshots' when translating this string
+shotPageScreenshotsDescription = Screenshots made simple. Take, save, and share  screenshots without leaving Firefox.
+shotPageUpsellFirefox = Get Firefox now
+shotPageDoesNotExpire = Does not expire
+shotPageDMCAMessage = This shot is no longer available due to a third party intellectual property claim.
+// Note: {$dmca} is a placeholder for a link to send email (a 'mailto' link)
+shotPageDMCAContact = Please email {$dmca} to request further information.
+// Note: do not translate 'Firefox Screenshots' when translating this string
+shotPageDMCAWarning = If your Shots are subject to multiple claims, we may revoke your access to Firefox Screenshots.
+// Note: {$url} is a placeholder for a shot page URL
+shotPageDMCAIncludeLink = Please include the URL of this shot in your email: {$url}
+shotPageKeepFor = How long should this shot be retained?
+// Note: shotPageSelectTime is a placeholder label for the time selection dropdown.
+shotPageSelectTime = Select time
+shotPageKeepIndefinitely = Indefinitely
+shotPageKeepTenMinutes = 10 Minutes
+shotPageKeepOneHour = 1 Hour
+shotPageKeepOneDay = 1 Day
+shotPageKeepOneWeek = 1 Week
+shotPageKeepTwoWeeks = 2 Weeks
+shotPageKeepOneMonth = 1 Month
+shotPageSaveExpiration = save
+shotPageCancelExpiration = cancel
+shotPageDoesNotExpire = does not expire
+// Note: {$timediff} is a placeholder for a future relative time clause, like "in 1 week" or "tomorrow"
+shotPageExpiresIn = expires {$timediff}
+// Note: {$timediff} is a placeholder for a past relative time clause, like "1 week ago" or "yesterday"
+shotPageExpired = expired {$timediff}
+timeDiffJustNow = just now
+timeDiffMinutesAgo = { $number ->
+  [one] 1 minute ago
+  *[other] {$number} minutes ago
+}
+timeDiffHoursAgo = { $number ->
+  [one] 1 hour ago
+  *[other] {$number} hours ago
+}
+timeDiffDaysAgo = { $number ->
+  [one] yesterday
+  *[other] {$number} days ago
+}
+timeDiffFutureSeconds = in a few seconds
+timeDiffFutureMinutes = { $number ->
+  [one] in 1 minute
+  *[other] in {$number} minutes
+}
+timeDiffFutureHours = { $number ->
+  [one] in 1 hour
+  *[other] in {$number} hours
+}
+timeDiffFutureDays = { $number ->
+  [one] tomorrow
+  *[other] in {$number} days
+}
+
+// Shotindex page
+
+// {$status} is a placeholder for an HTTP status code, like '500'.
+// {$statusText} is a text description of the status code, like 'Internal server error'.
+shotIndexPageErrorDeletingShot = Error deleting shot: {$status} {$statusText}
+// {$searchTerm} is a placeholder for text the user typed into the search box
+shotIndexPageSearchResultsTitle = My Shots: search for {$searchTerm}
+// {$error} is a placeholder for a non-translated error message that could be shared
+// with developers when debugging an error.
+shotIndexPageErrorRendering = Error rendering page: {$error}
+shotIndexPageSearchPlaceholder
+  .placeholder = Search my shots
+shotIndexPageSearchButton
+  .title = Search
+shotIndexPageNoShotsMessage = No saved shots.
+shotIndexPageNoShotsInvitation = Go on, create some.
+shotIndexPageLookingForShots = Looking for your shots…
+shotIndexPageNoSearchResultsIntro = Hmm
+shotIndexPageNoSearchResults = We canʼt find any shots that match your search.
+shotIndexPageClearSearchButton
+  .title = Clear search
+shotIndexPageConfirmShotDelete = Delete this shot?
+
+// Metrics page
+// Note: all metrics strings are optional for translation
+
+// Note: 'Firefox Screenshots' should not be translated
+metricsPageTitle = Firefox Screenshots Metrics
+metricsPageTotalsQueryTitle = Totals
+metricsPageTotalsQueryDescription = An overview of Screenshots
+metricsPageTotalsQueryDevices = Total devices registered
+metricsPageTotalsQueryActiveShots = Active shots
+metricsPageTotalsQueryExpiredShots = Expired (but recoverable)
+metricsPageTotalsQueryExpiredDeletedShots = Expired (and deleted)
+metricsPageShotsQueryTitle = Shots By Day
+metricsPageShotsQueryDescription = Number of shots created each day (for the last 30 days)
+metricsPageShotsQueryCount = Number of shots
+metricsPageShotsQueryDay = Day
+metricsPageUsersQueryTitle = Users By Day
+metricsPageUsersQueryDescription = Number of users who created at least one shot, by day (last 30 days)
+metricsPageUsersQueryCount = Number of users
+metricsPageUsersQueryDay = Day
+metricsPageUserShotsQueryTitle = Number of Shots per User
+metricsPageUserShotsQueryDescription = The number of users who have about N total shots
+metricsPageUserShotsQueryCount = Number of users
+metricsPageUserShotsQueryShots = Approximate number of active (unexpired) shots
+metricsPageRetentionQueryTitle = Retention By Week
+metricsPageRetentionQueryDescription = Number of days from a userʼs first shot to most recent shot, grouped by starting week
+metricsPageRetentionQueryUsers = Number of users
+metricsPageRetentionQueryDays = Days from the userʼs first to most recent shot
+metricsPageRetentionQueryFirstWeek = Week the user first created a shot
+metricsPageTotalRetentionQueryTitle = Total Retention
+metricsPageTotalRetentionQueryDescription = Length of time users have been creating shots, grouped by week
+metricsPageTotalRetentionQueryUsers = Number of users
+metricsPageTotalRetentionQueryDays = Days the user has been creating shots
+metricsPageVersionQueryTitle = Add-on Version
+metricsPageVersionQueryDescription = The version of the add-on used during login, in the last 14 days
+metricsPageVersionQueryUsers = Number of users logging in
+metricsPageVersionQueryVersion = Add-on version
+metricsPageVersionQueryLastSeen = Day
+metricsPageHeader = Metrics
+// Note: {$created} is a placeholder for a localized date and time, like '4/21/2017, 3:40:04 AM'
+metricsPageGeneratedDateTime = Generated at: {$created}
+// Note {$time} is a placeholder for a number of milliseconds, like '100'
+metricsPageDatabaseQueryTime = (database time: {$time}ms)
+
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -24,24 +24,29 @@ homePageDownloadFirefoxSubTitle = Free Download
 homePageGetStarted = Get Started
 // Note: do not translate 'Firefox Screenshots' when translating this string
 homePageHowScreenshotsWorks = How Firefox Screenshots Works
-homePageGetStarted = Get Started
+homePageGetStartedTitle = Get Started
+// Note: Screenshots is an abbreviation for Firefox Screenshots, and should not be translated.
 homePageGetStartedDescription = Find the new Screenshots icon on your toolbar. Select it, and the Screenshots menu will appear on top of your browser window.
 homePageCaptureRegion = Capture a Region
+// Note: Screenshots is an abbreviation for Firefox Screenshots, and should not be translated.
 homePageCaptureRegionDescription = Click and drag to select the area you want to capture. Or just hover and click — Screenshots will select the area for you. Like what you see? Select Save to access your screenshot online or the down arrow button to download it to your computer.
 homePageCapturePage = Capture a Page
 homePageCapturePageDescription = Use the buttons in the upper right to capture full pages. The Save Visible button will capture the area you can view without scrolling, and the Save Full Page will capture everything on the page.
 homePageSaveShare = Save and Share
+// Note: Screenshots is an abbreviation for Firefox Screenshots, and should not be translated.
 homePageSaveShareDescription = When you take a shot, Firefox posts your screenshot to your online Screenshots library and copies the link to your clipboard. We automatically store your screenshot for two weeks, but you can delete shots at any time or change the expiration date to keep them in your library for longer.
 homePageLegalLink = Legal
 homePagePrivacyLink = Privacy
 homePageTermsLink = Terms
-homePagePrivacyLink = Cookies
+homePageCookiesLink = Cookies
 
 // Leave Screenshots page
 
 leavePageConfirmDelete = Confirm account deletion
+// Note: do not translate 'Firefox Screenshots' when translating this string
 leavePageErrorAddonRequired = You must have Firefox Screenshots installed to delete your account
 leavePageErrorGeneric = An error occurred
+// Note: do not translate 'Firefox Screenshots' when translating this string
 leavePageWarning = This will permanently erase all of your Firefox Screenshots data.
 leavePageButtonProceed = Proceed
 leavePageButtonCancel = Cancel
@@ -91,8 +96,7 @@ shotPageAbuseButton
 shotPageDownloadShot
   .title = Download
 shotPageDownload = Download
-// Note: do not translate 'Firefox Screenshots' when translating this string
-shotPageScreenshotsDescription = Screenshots made simple. Take, save, and share  screenshots without leaving Firefox.
+shotPageScreenshotsDescription = Screenshots made simple. Take, save, and share screenshots without leaving Firefox.
 shotPageUpsellFirefox = Get Firefox now
 shotPageDoesNotExpire = Does not expire
 shotPageDMCAMessage = This shot is no longer available due to a third party intellectual property claim.
@@ -175,16 +179,17 @@ shotIndexPageConfirmShotDelete = Delete this shot?
 // Note: 'Firefox Screenshots' should not be translated
 metricsPageTitle = Firefox Screenshots Metrics
 metricsPageTotalsQueryTitle = Totals
+// Note: Screenshots is an abbreviation for Firefox Screenshots, and should not be translated.
 metricsPageTotalsQueryDescription = An overview of Screenshots
 metricsPageTotalsQueryDevices = Total devices registered
 metricsPageTotalsQueryActiveShots = Active shots
 metricsPageTotalsQueryExpiredShots = Expired (but recoverable)
 metricsPageTotalsQueryExpiredDeletedShots = Expired (and deleted)
-metricsPageShotsQueryTitle = Shots By Day
+metricsPageShotsQueryTitle = Shots by Day
 metricsPageShotsQueryDescription = Number of shots created each day (for the last 30 days)
 metricsPageShotsQueryCount = Number of shots
 metricsPageShotsQueryDay = Day
-metricsPageUsersQueryTitle = Users By Day
+metricsPageUsersQueryTitle = Users by Day
 metricsPageUsersQueryDescription = Number of users who created at least one shot, by day (last 30 days)
 metricsPageUsersQueryCount = Number of users
 metricsPageUsersQueryDay = Day
@@ -192,7 +197,7 @@ metricsPageUserShotsQueryTitle = Number of Shots per User
 metricsPageUserShotsQueryDescription = The number of users who have about N total shots
 metricsPageUserShotsQueryCount = Number of users
 metricsPageUserShotsQueryShots = Approximate number of active (unexpired) shots
-metricsPageRetentionQueryTitle = Retention By Week
+metricsPageRetentionQueryTitle = Retention by Week
 metricsPageRetentionQueryDescription = Number of days from a userʼs first shot to most recent shot, grouped by starting week
 metricsPageRetentionQueryUsers = Number of users
 metricsPageRetentionQueryDays = Days from the userʼs first to most recent shot
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -98,7 +98,6 @@ shotPageDownloadShot
 shotPageDownload = Download
 shotPageScreenshotsDescription = Screenshots made simple. Take, save, and share screenshots without leaving Firefox.
 shotPageUpsellFirefox = Get Firefox now
-shotPageDoesNotExpire = Does not expire
 shotPageDMCAMessage = This shot is no longer available due to a third party intellectual property claim.
 // Note: {$dmca} is a placeholder for a link to send email (a 'mailto' link)
 shotPageDMCAContact = Please email {$dmca} to request further information.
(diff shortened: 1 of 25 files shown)


locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+// creatingPageTitleDefault.
+creatingPageTitle = Creating {title}
+creatingPageTitleDefault = page
+
+// home page
+
+homePageDescription
+  .content = Intuitive screenshots baked right into the browser. Capture, save and share screenshots as you browse the Web using Firefox.
+homePageButtonMyShots = Go To My Shots
+homePageTeaser = Coming Soon...
```
flodolo: Use …

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+
+// leave screenshots page
+
+leavePageConfirmDelete = Confirm account deletion
+leavePageErrorAddonRequired = You must have the addon installed to delete your account
```
flodolo: add-on

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+// TODO: do we need this separate error message?
+leavePageErrorAddonRequired2 = You must have the addon installed to leave
```
flodolo: add-on

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+leavePageErrorGeneric = An error occurred
+leavePageWarning = This will permanently erase all of your Firefox Screenshots data.
+leavePageButtonProceed = Proceed
+leavePageButtonCancel = cancel
```
flodolo: Is the case correct here, given "Proceed"?

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+shotPageAlertErrorUpdatingExpirationTime = Error saving expiration: {status} {statusText}
+shotPageAlertErrorDeletingShot = Error deleting shot: {status} {statusText}
+shotPageAlertErrorUpdatingTitle = Error saving title: {status} {statusText}
+
+shotPageShareButton
+  .title = Share
+shotPageCopy = Copy
+shotPageCopied = Copied
+shotPageShareFacebook
+  .title = Share to Facebook wall or message
```
flodolo: Is "Facebook wall" still a thing?

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+shotPageShareEmail
+  .title = Create email with link
+shotPageShareLink = Get a shareable link to this shot:
+shotPagePrivacyMessage = This shot is only visible to you until you share the link.
+shotPageCopyImageText
+  .label = Copy Image Text
+shotPageConfirmDeletion = Are you sure you want to delete this shot permanently?
+// Note: {timediff} is a placeholder for a future time like 'tomorrow' or 'next week'
+// TODO: "in {timediff}" probably won't work across languages. maybe an absolute date is best.
+shotPageExpirationMessage = If you do nothing, this shot will be permanently deleted in {timediff}.
```
flodolo: How do you plan to localize the time in both cases?

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+// Note: {timediff} is a placeholder for a time duration phrase like '30 days'
+// TODO: again, "for {timediff}" may not work well.
+shotPageRestoreButton = restore for {timediff}
```
flodolo: Is lowercase correct? Also not sure what this means in practice

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+shotPageDeleteButton
+  .title = Delete this shot permanently
+shotPageAbuseButton
+  .title = Report this shot for abuse, spam, or other problems
+shotPageMyShots = My Shots
+shotPageDownloadShot
+  .title = Download the shot image
+shotPageDownload = Download
+// Note: do not translate 'Firefox Screenshots' when translating this string
+shotPageScreenshotsDescription  = Firefox Screenshots made simple. Take, save and share screenshots without leaving Firefox.
```
flodolo: Should this actually be "Firefox screenshots made easy", as in "Screenshots are now easy in Firefox"?

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+// with developers when debugging an error.
+shotIndexPageErrorRendering = Error rendering page: {error}
+shotIndexPageSearchPlaceholder
+  .placeholder = search my shots
+shotIndexPageSearchButton
+  .title = search
+shotIndexPageNoShotsMessage = No saved shots.
+shotIndexPageNoShotsInvitation = Go on, create some.
+// TODO should we include the ellipsis as three periods, or leave a placeholder for an html entity?
+shotIndexPageLookingForShots = Looking for your shots...
```
flodolo: I'm confused by the comment: why not using the proper unicode character …?

locales/en-US/server.ftl
```diff
@@ -0,0 +1,161 @@
...
+metricsPageUsersQueryTitle = Users By Day
+metricsPageUsersQueryDescription = Number of users who created at least one shot, by day (last 30 days)
+metricsPageUsersQueryCount = Number of users
+metricsPageUsersQueryDay = Day
+metricsPageUserShotsQueryTitle = Number of Shots per User
+metricsPageUserShotsQueryDescription = The number of users who have about N total shots
+metricsPageUserShotsQueryCount = Number of users
+metricsPageUserShotsQueryShots = Approximate number of active (unexpired) shots
+metricsPageRetentionQueryTitle = Retention By Week
+metricsPageRetentionQueryDescription = Number of days from a user's first shot to most recent shot, grouped by starting week
```
flodolo: Use proper apostrophes instead of straight quotes: ' -> ’

locales/en-US/server.ftl
6a68: I think we had discussed switching to absolute times in the past, as there was a ton of language-specific work if we tried to retain relative times.

I'll work with UX / product this week to clarify any string changes and I'll ping you when the PR is updated.

locales/en-US/server.ftl
6a68: In English, this reads like "restore for 30 days" or "restore for one week".

locales/en-US/server.ftl
6a68: Great question. @johngruen, thoughts?

locales/en-US/server.ftl
6a68: The comment was just a reminder to myself to ask about the unicode ellipsis, which you already mentioned in another comment :+1:

6a68:
Thanks for the feedback, @flodolo! I'll namecheck you here when the PR is updated.

locales/en-US/server.ftl
```diff
@@ -78,14 +102,20 @@ shotPageDeleteButton
...
   .title = Report this shot for abuse, spam, or other problems
-shotPageMyShots = My Shots
 shotPageDownloadShot
   .title = Download the shot image
 shotPageDownload = Download
 // Note: do not translate 'Firefox Screenshots' when translating this string
 shotPageScreenshotsDescription  = Firefox Screenshots made simple. Take, save and share screenshots without leaving Firefox.
 shotPageUpsellFirefox = Get Firefox now
 shotPageDoesNotExpire = does not expire
+shotPageDMCAMessage = This shot is no longer available due to a third party intellectual property claim.
```
6a68: @johngruen Are we going to want the community localizers to handle the DMCA legalese, or do we need to handle this via a paid service?

flodolo: It seems safe-ish to let localizers translate it.

locales/en-US/server.ftl
6a68: @johngruen this and the next commented-one have relative times (the `{timediff}` placeholder) that need to be converted to absolutes

server/src/l10n.js
```diff
@@ -0,0 +1,85 @@
...
+    const locales = negotiateLanguages(
+      userLangs,
+      availableLocales,
+      { defaultLocale: 'en-US' }
+    );
+    return locales;
+  });
+}
+
+module.exports = { init, getText, userLangs, generateMessages };
```
ianb: Generally the code uses a pattern of setting `exports.init` etc directly – there's a downside of potentially having to refer to `exports.init()` locally in the module, but otherwise I like seeing that something is public right where it is declared.

server/src/pages/creating/model.js
```diff
@@ -1,11 +1,12 @@
 exports.createModel = function(req) {
   let finishedUrl = `/${encodeURIComponent(req.params.id)}/${encodeURIComponent(req.params.domain)}`;
-  let title = req.query.title || "page";
+  let title = req.query.title || req.getText("creatingPageTitleDefault");
```
ianb: Probably the better way to do this is:

```js
let title = req.query.title ? req.getText("creatingPageTitle", {title: req.query.title}) : req.getText("creatingPageTitleDefault");
```

server/src/pages/homepage/view.js
```diff
@@ -601,45 +617,66 @@ class Body extends React.Component {
...
             <div className="section-image"></div>
           </div>
         </section>
         <section id="section-2">
           <div className="container">
             <div className="section-content">
-              <h3><span className="bold">Capture</span> a Region</h3>
-              <p>Click and drag to select the area you want to capture. Or just hover and click — Screenshots will select the area for you. Like what you see? Select Save to access your screenshot online or the down arrow button to download it to your computer.</p>
+              <Localized id="homePageCaptureRegion">
+                <h3><Localized id="homePageCapture"><span className="bold">Capture</span></Localized> a Region</h3>
```
ianb: This nested Localized won't work, will it?  I'm not sure what would work, it gets so complicated.  I think we should remove the bold and try to put it in later.

server/src/pages/homepage/view.js
```diff
@@ -601,45 +617,66 @@ class Body extends React.Component {
...
             <div className="section-image"></div>
           </div>
         </section>
         <section id="section-3">
           <div className="container">
             <div className="section-content">
-              <h3><span className="bold">Capture</span> a Page</h3>
-              <p>Use the buttons in the upper right to capture full pages. The Save Visible button will capture the area you can view without scrolling, and the Save Full Page will capture everything on the page.</p>
+              <Localized id="homePageCapturePage">
+                <h3><Localized id="homePageCapture"><span className="bold">Capture</span></Localized> a Page</h3>
```
ianb: Ditto nested Localized

server/src/pages/homepage/view.js
```diff
@@ -601,45 +617,66 @@ class Body extends React.Component {
...
             <div className="section-image"></div>
           </div>
         </section>
         <section id="section-4">
           <div className="container">
             <div className="section-content">
-              <h3><span className="bold">Save</span> and <span className="bold">Share</span></h3>
-              <p>When you take a shot, Firefox posts your screenshot to your online Screenshots library and copies the link to your clipboard. We automatically store your screenshot for two weeks, but you can delete shots at any time or change the expiration date to keep them in your library for longer. </p>
+              <Localized id="homePageSaveShare">
+                <h3><Localized id="homePageSave"><span className="bold">Save</span></Localized> and
```
ianb: Another nested case

server/src/pages/leave-screenshots/server.js
```diff
@@ -11,7 +11,7 @@ exports.app = app;
 
 app.get("/", csrfProtection, function(req, res) {
   if (!req.deviceId) {
-    res.status(403).send("You must have the addon installed to delete your account");
+    res.status(403).send(req.getText("leavePageErrorAddonRequired"));
```
ianb: I'm somewhat indifferent to translating these responses. They are plain text and shouldn't actually show up to anyone unless they do something weird.  But whatever.

server/src/pages/shot/view.js
```diff
@@ -250,11 +268,16 @@ class Body extends React.Component {
...
     if (this.props.isOwner) {
+      // todo l10n - timediffs
       restoreWidget = (
         <p>
-          If you do nothing,
-          this shot will be permanently deleted in <TimeDiff date={deleteTime} />.
-          <button className="button primary" onClick={this.onRestore.bind(this)}>restore for {intervalDescription(this.props.defaultExpiration)}</button>
+          <Localized id="shotPageExpirationMessage">
+            If you do nothing,
+            this shot will be permanently deleted in <TimeDiff date={deleteTime} />.
```
ianb: Does Fluent handle the `<TimeDiff>` here?

server/src/pages/shot/view.js
```diff
@@ -250,11 +268,16 @@ class Body extends React.Component {
...
+          </Localized>
+          <Localized id="shotPageRestoreButton">
+            <button className="button primary" onClick={this.onRestore.bind(this)}>restore for {intervalDescription(this.props.defaultExpiration)}</button>
```
ianb: Does Fluent turn `{intervalDescription(this.props.defaultExpiration)}` into a variable substitution automatically?

locales/en-US/server.ftl
johngruen: good catch

server/src/pages/shotindex/view.js
```diff
@@ -248,6 +287,7 @@ class Card extends React.Component {
   }
 
   displayTitle(title) {
+    // todo l10n: this won't work for rtl languages. use CSS ellipsis instead?
```
ianb: I recall overflow being wonky here, but the CSS has been refactored a few times so maybe it will work now.  In which case we might not need this at all, *or* we might just be able to lose the ellipsis (but keep the truncation) and let people figure it out.

server/src/reactruntime.js
```diff
@@ -3,6 +3,8 @@ require("core-js");
 const React = require("react");
 const ReactDOM = require("react-dom");
 const linker = require("./linker");
+const { LocalizationProvider } = require("fluent-react/compat");
+const l10n = require("./l10n");
```
ianb: Ah, this is the bad import!  reactruntime is used in the browser, and so it should only import modules that are okay to have in the browser, and `./l10n` is not.  Instead we should be injecting the output of l10n into the module somehow.

ianb: `./linker` is a good example here – instead of that module actually finding the git revision, it has a `setGitRevision` function. There's two codepaths for injecting that content, one for the server and another for the client.

`reactrender.js` might be where the localization should get injected into the content. It could even be as simple as adding `window.localizedMessages = ${JSON.stringify(generateMessages(l10n.userLangs))` and then using that in `reactruntime`. Or maybe it would be better to add the messages to the model (which can happen in reactruntime), similar to gitRevision.

ianb:
I think `TimeDiff.makeDiffString` is actually pretty localizable, even if it's a little tedious. Except for pluralization I guess. I don't know if we can change the strings a little to avoid that, like instead of `{n} minutes` use `minutes: {n}` ?

flodolo:
> I think TimeDiff.makeDiffString is actually pretty localizable, even if it's a little tedious. Except for pluralization I guess. I don't know if we can change the strings a little to avoid that, like instead of {n} minutes use minutes: {n} ?

The plural issues seems small when you only have two, when you have 5 or 6 it becomes really easy to see a string with grammar errors. IMO, the only solution at the moment is to display an absolute datetime.

locales/en-US/server.ftl
johngruen: Absolute dates seem kind of gross here...particularly when we get down to small increments. @flod ther code for these time diffs can be found here: https://github.com/mozilla-services/screenshots/blob/master/server/src/pages/shot/time-diff.js#L27

I've set the time diffs so they're all increments of time: 1 minute, 1 day etc. Seems like these would be l10n-able

johngruen: @ianb seems like we could l10n-ify all of the timediff options pretty readily, but i'm not sure how to read this last one: https://github.com/mozilla-services/screenshots/blob/master/server/src/pages/shot/time-diff.js#L57

ianb: Well... this one might be fixable by making it more of a constant. It's based on configuration, but right now we configure it to always be "restore for 14 days".  The equation in that file just translates seconds to days.

johngruen:
@flod @ianb I'd rather not use absolute times here if we can avoid them. The idea of giving a date stamp when a shot expires in 5 minutes seems like a major usability regression. Not sure i understand the difficulty around pluralization. Can you provide an example?

locales/en-US/server.ftl
flodolo: (pinged wrong GH username) For a moment I forgot we were talking about the website… Since it's using Fluent, there's proper plural support, so we should be able to localize this.
http://projectfluent.io/fluent/guide/selectors.html

johngruen: awesome!

6a68:
@flodolo I believe @johngruen was pinging you in the previous comment here

server/src/pages/homepage/view.js
6a68: Nesting Localized components seems to work just fine:

screen shot 2017-07-06 at 9 12 45 am

flodolo:
Pretty sure I replied already… 
https://github.com/mozilla-services/screenshots/pull/3068#discussion_r125340338

server/src/pages/leave-screenshots/server.js
6a68: Yeah, definitely a toss-up. I localized this one because it was offering a possibly confused user instructions on how to complete a task  ¯\_(ツ)_/¯

6a68:
@flodolo Sorry, should've been more specific - I was hoping you could respond to [this question](https://github.com/mozilla-services/screenshots/pull/3068#issuecomment-312709024) related to [your comment](https://github.com/mozilla-services/screenshots/pull/3068#issuecomment-312706897) about using absolute datetimes in `TimeDiff.makeDiffString`.

flodolo:
Same answer, that can be solved by having proper plural support. It will become messy very fast though, e.g.

```
timediff-future-min = { $n ->
        [one]  If you do nothing, this shot will be permanently deleted in one minute
       *[other] If you do nothing, this shot will be permanently deleted in $n minutes
}

timediff-future-hours = { $n ->
        [one]  If you do nothing, this shot will be permanently deleted in one hour
       *[other] If you do nothing, this shot will be permanently deleted in $n hours
} 
```

Also cc @stasm, in case I'm missing something obvious. We don't have anything to manage relative times, do we?

server/src/pages/shot/view.js
6a68: Good catch. Because the timediff is a placeholder in the ftl file, I think it's best to explicitly make it a variable attribute.

server/src/pages/homepage/view.js
6a68: Hmm. This example seems to suggest passing in nested components as variables, so I'll switch to that https://github.com/projectfluent/fluent.js/wiki/Localized

6a68:
@stas @flodolo Sorry if I keep asking the same question, but TimeDiff is inserted into English strings in a way that I'm not sure is easy to localize, for example: "this shot expired $time ago", "this shot expires in $time", or "keep this shot for $time", where $time is a relative time, like "5 minutes".

What restrictions do we need to have around how these strings are constructed? Does the "ago", "in", or "for" need to be included in the relative time string, so that we'd build up the string like: "this shot expires " + "in 5 minutes", as opposed to "this shot expires in " + "5 minutes"?

I'm finishing up the TimeDiff localization and wondering where to put the "in" and "ago".

flodolo:
in, ago, etc. should be part of the string, i.e. this shot expires " + "in 5 minutes"
http://www.unicode.org/cldr/charts/31/summary/en.html#1543

We even have a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1270140) to expose these via Intl API, but that wouldn't solve the issue on a website :-\

6a68:
Adding my current TODO list to try to give some visibility into progress:

- [x] install errors (fixed by force-clearing the npm 5 cache and regenerating npm-shrinkwrap)
- [x] timediff busted
- [x] expiry picker busted on shot page
  - note, removed the bullet to get around this
- [ ] on shotindex page, deleting a shot, window.confirm says “Delete shotTitle?” instead of properly inserting the shot title

ianb:
:+1: on followup for home page, react keys. Probably window.confirm too, we can just change it to "Delete this shot?" for now (or heck, forever).

6a68:
> Probably window.confirm too, we can just change it to "Delete this shot?" for now (or heck, forever)

Hey, great idea! I did something similar to get rid of the intervalDescription, replacing the slightly awkward "restore for 1 day 1 hour 15 seconds" with "restore until 7/12/2017, 3:19pm".

Let me push latest and you can take a final spin or just land it

6a68:
doing a final run-through now

6a68:
did a final rebase and ran through all the flows/pages. everything seems to work :+1:

fixing up the linter errors...

6a68:
@ianb OK, I think this is fully baked. Mind taking a last look before landing?

flodolo:
> @ianb OK, I think this is fully baked. Mind taking a last look before landing?

Please do not land without coordination with l10n. We need to explain to localizers why all of a sudden all these strings appear in the project, make sure that Pontoon works as expected, etc.

It's already Thursday in Europe, I wonder if we should just wait and merge early on Monday next week, unless @mathjazz has cycles to take care of this.

6a68:
@flodolo Thanks for catching that comment while on PTO! I'll coordinate with @mathjazz before landing

locales/en-US/server.ftl
```diff
@@ -0,0 +1,205 @@
...
+shotPageKeepOneMonth = 1 Month
+shotPageSaveExpiration = save
+shotPageCancelExpiration = cancel
+shotPageDoesNotExpire = does not expire
+// Note: {$timediff} is a placeholder for a future relative time clause, like "in 1 week" or "tomorrow"
+shotPageExpiresIn = expires {$timediff}
+// Note: {$timediff} is a placeholder for a past relative time clause, like "1 week ago" or "yesterday"
+shotPageExpired = expired {$timediff}
+timeDiffJustNow = just now
+// TODO: should we just have one string with plural rules?
```
flodolo: We can't expose strings like this, we need proper plurals for these strings

locales/en-US/server.ftl
```diff
@@ -0,0 +1,205 @@
...
+timeDiffOneMinuteAgo = 1 minute ago
+timeDiffMinutesAgo = {$number} minutes ago
```
flodolo:
```
timeDiffMinutesAgo = { $number ->
        [one] 1 minute ago
       *[other] {$number} minutes ago
    }
```

locales/en-US/server.ftl
```diff
@@ -0,0 +1,205 @@
...
+homePageLegalLink = Legal
+homePagePrivacyLink = Privacy 
+homePageTermsLink = Terms
+homePagePrivacyLink = Cookies
+
+// Leave Screenshots page
+
+leavePageConfirmDelete = Confirm account deletion
+leavePageErrorAddonRequired = You must have Firefox Screenshots installed to delete your account
+leavePageErrorAddonRequired2 = You must have Firefox Screenshots installed to leave
```
flodolo: What "leave" means in this context? Log out? Might be worth a comment

locales/en-US/server.ftl
```diff
@@ -0,0 +1,205 @@
...
+shotPageExpiredMessageDetails = Here is the page it was originally created from:
+shotPageDeleteButton
+  .title = Delete this shot
+shotPageAbuseButton
+  .title = Report this shot for abuse, spam, or other problems
+shotPageDownloadShot
+  .title = Download
+shotPageDownload = Download
+// Note: do not translate 'Firefox Screenshots' when translating this string
+shotPageScreenshotsDescription  = Screenshots made simple. Take, save, and share  screenshots without leaving Firefox.
```
flodolo: Double whitespace

locales/en-US/server.ftl
6a68: Thanks, @flodolo!

@stasm Is this plural format supported by fluent-react right now? I'm using `Localized` as follows (from time-diff.js)
```html
<Localized id={l10nID} $number={elapsedTime}>
  <span title={this.dateString(this.props.date)}></span>
</Localized>
```
but I'm not seeing the output rendered at all. Any thoughts on why that might be?

6a68: @flodolo hmm, if stas is out of office, I'm inclined to land this with the existing pluralization and file a bug to fix later. 55 goes to release channel August 8th, so we have less than a month to localize the server strings. Are you OK with fixing the plurals later?

flodolo: It really sucks, because I think several languages won't be able to translate this in a meaningful way. It's not a stand-alone sentence that you can turn around, e.g. "5 minutes remaining" to "Minutes remaining: 5".

I'm tempted to ask you to add proper plural forms to the FTL file, so we can start translating, and then fix the code later as soon @stasm is around. And if that's not supported by fluent-react (I'd be surprised), we can find a different solution.

Pike: Too bad https://github.com/tc39/proposal-intl-relative-time isn't here yet :-/

I agree that we should do this with proper plurals, it's going to look yucky in quite a few languages without that.

Alternatively, how about we keep the two strings for this element hard-coded in english (by adding the verbatim after the addMessages(...server.js content)) ? Also yucky, but unblocks the localization of the other ton of strings.

I did look a tiny bit into the backend, the Localized impl seems like it tries to do the right thing. I'm not familiar with react at all, so I have a hard time saying whether the tests in https://github.com/projectfluent/fluent.js/blob/master/fluent-react/test/localized_render_test.js cover that.

6a68:
@Pike @flodolo After blowing away node_modules and reinstalling, the plurals seem to be working fine with fluent-react. I'm good to land this if you are.

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+shotPageDoesNotExpire = does not expire
+// Note: {$timediff} is a placeholder for a future relative time clause, like "in 1 week" or "tomorrow"
+shotPageExpiresIn = expires {$timediff}
+// Note: {$timediff} is a placeholder for a past relative time clause, like "1 week ago" or "yesterday"
+shotPageExpired = expired {$timediff}
+timeDiffJustNow = just now
+timeDiffMinutesAgo = { $number ->
+  [one] 1 minute ago
+  *[other] {$number} minutes ago
+}
```
6a68: I've updated the FTL file with proper pluralization

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+  .content = Intuitive screenshots baked right into the browser. Capture, save and share screenshots as you browse the Web using Firefox.
+homePageButtonMyShots = Go To My Shots
+homePageTeaser = Coming Soon…
+homePageDownloadFirefoxTitle = Firefox
+homePageDownloadFirefoxSubTitle = Free Download
+homePageGetStarted = Get Started
+// Note: do not translate 'Firefox Screenshots' when translating this string
+homePageHowScreenshotsWorks = How Firefox Screenshots Works
+homePageGetStarted = Get Started
+homePageGetStartedDescription = Find the new Screenshots icon on your toolbar. Select it, and the Screenshots menu will appear on top of your browser window.
```
flodolo: Add a comment explaining that Screenshots here is meant an abbreviation for "Firefox Screenshots", and should not be translated. BTW, the icon says Firefox Screenshots in my Nightly.

```
// Note: Screenshots is an abbreviation for Firefox Screenshots, and should not be translated.
```

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+homePageCaptureRegion = Capture a Region
+homePageCaptureRegionDescription = Click and drag to select the area you want to capture. Or just hover and click — Screenshots will select the area for you. Like what you see? Select Save to access your screenshot online or the down arrow button to download it to your computer.
```
flodolo: Same comment here about Screenshots

```
// Note: Screenshots is an abbreviation for Firefox Screenshots, and should not be translated.
```

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+homePageCapturePage = Capture a Page
+homePageCapturePageDescription = Use the buttons in the upper right to capture full pages. The Save Visible button will capture the area you can view without scrolling, and the Save Full Page will capture everything on the page.
+homePageSaveShare = Save and Share
+homePageSaveShareDescription = When you take a shot, Firefox posts your screenshot to your online Screenshots library and copies the link to your clipboard. We automatically store your screenshot for two weeks, but you can delete shots at any time or change the expiration date to keep them in your library for longer.
```
flodolo: Same here 

```
// Note: Screenshots is an abbreviation for Firefox Screenshots, and should not be translated.
```

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+// Note: This phrase is followed by an empty line, then the URL of the source page
+shotPageExpiredMessageDetails = Here is the page it was originally created from:
+shotPageDeleteButton
+  .title = Delete this shot
+shotPageAbuseButton
+  .title = Report this shot for abuse, spam, or other problems
+shotPageDownloadShot
+  .title = Download
+shotPageDownload = Download
+// Note: do not translate 'Firefox Screenshots' when translating this string
```
flodolo: There's no Firefox Screenshots in the string, and I think Screenshots in this sentence is a noun, not the brand. Drop the comment?

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+shotPageScreenshotsDescription = Screenshots made simple. Take, save, and share  screenshots without leaving Firefox.
```
flodolo: double space after share

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+homePageSaveShareDescription = When you take a shot, Firefox posts your screenshot to your online Screenshots library and copies the link to your clipboard. We automatically store your screenshot for two weeks, but you can delete shots at any time or change the expiration date to keep them in your library for longer.
+homePageLegalLink = Legal
+homePagePrivacyLink = Privacy
+homePageTermsLink = Terms
+homePagePrivacyLink = Cookies
+
+// Leave Screenshots page
+
+leavePageConfirmDelete = Confirm account deletion
+leavePageErrorAddonRequired = You must have Firefox Screenshots installed to delete your account
```
flodolo: Since you added it to all strings, add 

```
// Note: do not translate 'Firefox Screenshots' when translating this string
```

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+leavePageErrorGeneric = An error occurred
+leavePageWarning = This will permanently erase all of your Firefox Screenshots data.
```
flodolo: Same here

```
// Note: do not translate 'Firefox Screenshots' when translating this string
```

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+  .title = Clear search
+shotIndexPageConfirmShotDelete = Delete this shot?
+
+// Metrics page
+// Note: all metrics strings are optional for translation
+
+// Note: 'Firefox Screenshots' should not be translated
+metricsPageTitle = Firefox Screenshots Metrics
+metricsPageTotalsQueryTitle = Totals
+metricsPageTotalsQueryDescription = An overview of Screenshots
```
flodolo:
```
// Note: Screenshots is an abbreviation for Firefox Screenshots, and should not be translated.
```

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+metricsPageTotalsQueryDevices = Total devices registered
+metricsPageTotalsQueryActiveShots = Active shots
+metricsPageTotalsQueryExpiredShots = Expired (but recoverable)
+metricsPageTotalsQueryExpiredDeletedShots = Expired (and deleted)
+metricsPageShotsQueryTitle = Shots By Day
```
flodolo: Not a native speaker, but is this proper title case? I assume "by" should be lowercase.

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+metricsPageShotsQueryDescription = Number of shots created each day (for the last 30 days)
+metricsPageShotsQueryCount = Number of shots
+metricsPageShotsQueryDay = Day
+metricsPageUsersQueryTitle = Users By Day
```
flodolo: Same question here about "by"

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+metricsPageUsersQueryDescription = Number of users who created at least one shot, by day (last 30 days)
+metricsPageUsersQueryCount = Number of users
+metricsPageUsersQueryDay = Day
+metricsPageUserShotsQueryTitle = Number of Shots per User
+metricsPageUserShotsQueryDescription = The number of users who have about N total shots
+metricsPageUserShotsQueryCount = Number of users
+metricsPageUserShotsQueryShots = Approximate number of active (unexpired) shots
+metricsPageRetentionQueryTitle = Retention By Week
```
flodolo: Same here: by or By?

locales/en-US/server.ftl
stasm: This is likely related to indentation in the FTL file. We should have better error reporting for this. The entire body of the translation which follows the `=` must be indented in Fluent:

This is correct:

```
timeDiffMinutesAgo = { $number ->
        [one] 1 minute ago
       *[other] {$number} minutes ago
    }
```

This is correct as well (and my preferred style):

```
timeDiffMinutesAgo =
    { $number ->
        [one] 1 minute ago
       *[other] {$number} minutes ago
    }
```

This OTOH results in a Syntax error because the closing `}` is not indented.

```
timeDiffMinutesAgo = { $number ->
    [one] 1 minute ago
   *[other] {$number} minutes ago
}
```

There's an open syntax spec issue about this https://github.com/projectfluent/fluent/issues/42 although I'm not sure we should allow this. It creates an inconsistency and I prefer to only allow it once we've depleted other options for preventing this error (error reporting, linting, prittifying).

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+  .title = Delete this shot
+shotPageAbuseButton
+  .title = Report this shot for abuse, spam, or other problems
+shotPageDownloadShot
+  .title = Download
+shotPageDownload = Download
+// Note: do not translate 'Firefox Screenshots' when translating this string
+shotPageScreenshotsDescription = Screenshots made simple. Take, save, and share  screenshots without leaving Firefox.
+shotPageUpsellFirefox = Get Firefox now
+shotPageDoesNotExpire = Does not expire
```
flodolo: I don't see an instance with this case in the code?

Pike:
https://gist.github.com/Pike/a91c004ff2627a875807f94767312105 is my config file, add that to the root directory, and 

```shell
pip install compare-locales
compare-locales l10n.toml .
```

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+homePageCaptureRegion = Capture a Region
+homePageCaptureRegionDescription = Click and drag to select the area you want to capture. Or just hover and click — Screenshots will select the area for you. Like what you see? Select Save to access your screenshot online or the down arrow button to download it to your computer.
+homePageCapturePage = Capture a Page
+homePageCapturePageDescription = Use the buttons in the upper right to capture full pages. The Save Visible button will capture the area you can view without scrolling, and the Save Full Page will capture everything on the page.
+homePageSaveShare = Save and Share
+homePageSaveShareDescription = When you take a shot, Firefox posts your screenshot to your online Screenshots library and copies the link to your clipboard. We automatically store your screenshot for two weeks, but you can delete shots at any time or change the expiration date to keep them in your library for longer.
+homePageLegalLink = Legal
+homePagePrivacyLink = Privacy
+homePageTermsLink = Terms
+homePagePrivacyLink = Cookies
```
flodolo: I guess homePageCookiesLink

server/src/pages/homepage/view.js
```diff
@@ -648,10 +683,18 @@ class Body extends React.Component {
...
+              <Localized id="homePageLegalLink">
+                <a href="https://www.mozilla.org/about/legal/" className="boilerplate">Legal</a>
+              </Localized>
+              <Localized id="homePagePrivacyLink">
+                <a href="https://www.mozilla.org/privacy/firefox-cloud/" className="boilerplate">Privacy</a>
+              </Localized>
+              <Localized id="homePageTermsLink">
+                <a href="https://www.mozilla.org/about/legal/terms/services/" className="boilerplate">Terms</a>
+              </Localized>
+              <Localized id="homePagePrivacyLink">
```
flodolo: homePageCookiesLink

locales/en-US/server.ftl
```diff
@@ -0,0 +1,214 @@
...
+homePageDescription
+  .content = Intuitive screenshots baked right into the browser. Capture, save and share screenshots as you browse the Web using Firefox.
+homePageButtonMyShots = Go To My Shots
+homePageTeaser = Coming Soon…
+homePageDownloadFirefoxTitle = Firefox
+homePageDownloadFirefoxSubTitle = Free Download
+homePageGetStarted = Get Started
+// Note: do not translate 'Firefox Screenshots' when translating this string
+homePageHowScreenshotsWorks = How Firefox Screenshots Works
+homePageGetStarted = Get Started
```
flodolo: Probably homePageGetStartedTitle to differentiate from the other one, which is a link?

server/src/pages/homepage/view.js
```diff
@@ -601,45 +616,65 @@ class Body extends React.Component {
...
-        <h2>How Screenshots Works</h2>
+        <Localized id="homePageHowScreenshotsWorks">
+          <h2>How Screenshots Works</h2>
+        </Localized>
         <section id="section-1">
           <div className="container">
             <div className="section-content">
-              <h3>Get <span className="bold">Started</span></h3>
-              <p>Find the new Screenshots icon on your toolbar. Select it, and the Screenshots menu will appear on top of your browser window.</p>
+              <Localized id="homePageGetStarted">
```
flodolo: homePageGetStartedTitle

Pike:
PPS: 
    compare-locales l10n.toml . en-US
treats en-US as a localization of itself, and reports errors and parser problems with location etc.

locales/en-US/server.ftl
6a68: It's used [here](https://github.com/mozilla-services/screenshots/pull/3068/files#diff-c089cb65f9a6ab0a43538517ddbcc5b6R525):

flodolo: To clarify: this string appear twice with the same ID, different case in the FTL file. In the code there's only the instance you pointed out, and it's all lowercase.

locales/en-US/server.ftl
6a68: Thanks, @stasm! I'm in favor of keeping the indentation requirement for these plural variants. It's nice to have consistency with other more complex cases, like keys whose attributes contain localizable strings. Also makes it easy to visually scan the file for anything that might need a little extra attention. And, I'm sure, it's easier to maintain a parser where the special cases are always indented off the start of the following line :-)

Issue #3080, opened 2017-07-03
FxA support


ianb:
These are requirements for FxA:

I propose we land this as two chunks, the first:

- [ ] Can access a /settings page from My Shots
- [ ] When you are not logged into FxA, a button is presented to connect / add an account
- [ ] Clicking that button puts you through the OAuth FxA flow
- [ ] If you cancel the login flow (on the FxA site), you should return to /settings
- [ ] If you complete the flow, you should end up back at /settings
- [ ] If you are in /settings and are logged in, you should see your avatar image, and your name or email
- [ ] We should store the avatar image, name, and email in the database (note: we initially left out the email, but we should include it as well)
- [ ] There should be a `x` button on the settings page, that redirects the user back to `/shots`
- [ ] If you are not logged in and try to access `/settings` you should get permission denied
- [ ] If you are in /settings and logged in, then there should be a Disconnect button
- [ ] The Disconnect button should erase `devices.accountid` for the current login (`req.deviceId`)
- [ ] There should be CSRF protection on Disconnect (see `/api/set-expiration` for an example); there does not have to be protection on login/connect
- [ ] The settings page and the link to the settings page should be controlled by a configuration variable (defined in `config.js`). We should default it to off. We can ask about turning it on just for dev.

Then as a follow-on:

- [ ] When you are logged in, you should be able to access shots made on associated devices as though you own them
- [ ] You should be able to delete shots, set their expiration, and set their title

youwenliang:
@niharikak101 
Here's the UX measurement:
https://mozilla.github.io/testpilot-assets/Firefox_Screenshots/FxA_Integration/

youwenliang:
@ianb 
How are we implementing the disconnect behavior now? Are we keeping the shots that are not connected to an account, or are we removing all the shots once a user disconnected?

niharikak101:
@youwenliang We are removing the shots associated with the account that don't belong to the device which the user disconnects from the account, but keeping the shots that were taken from that specefic device/profile. Essentially, when a user disconnects they can't access the shots that were taken from other devices associated with the account.

I've also added some screenshots of what the settings page looks like right now. @ianb and I talked about saying 'connect' instead of sign in, and I've removed the Manage Accounts button because we don't have any functionality for that, yet.

Connected Account:
![screenshot-2017-7-13 settings](https://user-images.githubusercontent.com/16074636/28150134-499cb162-67b0-11e7-9625-d581ced1749d.png)

Guest Account:
[...]

PR #3082, opened 2017-07-03
FxA

niharikak101:
This takes care of most (all?) of the functionality for the first part of #3080


.env.dev
```diff
@@ -1,5 +1,11 @@
-CLIENT_ID=209ee70cb3cdc851
-CLIENT_SECRET=efde5874713abf2a8ec9d8af8488fae7c7e0a933dba08208f644fa93b5622dab
+# Note, this is for a development/testing OAuth server:
+FXA_CLIENT_ID=f7de760c7c55ec9b
+FXA_CLIENT_SECRET=14c2a3c7577fd3b28ee7141ad726dcd4160f24b2619e2f908960844430bdee39
+# ?? Do we need FXA_REDIRECT_URI?
```
ianb: So far it appears we don't, so we can remove this reference

server/src/config.js
```diff
@@ -281,6 +274,13 @@ var conf = convict({
     default: "",
     env: "CDN",
     arg: "cdn"
+  },
+  userSettings: {
```
ianb: Let's name this something with the name "enable" in it, like "enableUserSettings" – then it's clearer that it's a feature flag.

server/src/pages/settings/controller.js
```diff
@@ -0,0 +1,33 @@
...
+  req.setRequestHeader("content-type", "application/x-www-form-urlencoded");
+  req.onload = function() {
+    if (req.status >= 300) {
+      // FIXME: a lame way to do an error message
+      window.alert("Error disconnecting: " + req.status + " " + req.statusText);
+    } else {
+      location.href = model.backend + "/settings";
+    }
+  };
+  console.log("csrf: " + model.csrfToken);
```
ianb: You should do an eslint check, I'm guessing it will fail, as we have a console.log rule.

server/src/pages/settings/model.js
```diff
@@ -0,0 +1,26 @@
+const db = require("../../db");
+
+exports.createModel = function(req) {
+  let title = "Settings";
+  let model = {
+    noAnalytics: true,
```
ianb: Oh, analytics!  I think we should have analytics, and we should add events for connect and disconnect.  I think the default pageview will handle the basic count of how many people visit the settings page.

server/src/pages/settings/view.js
```diff
@@ -0,0 +1,72 @@
...
+              <a className="account-buttons disconnect" href="" onClick={ this.onClickDisconnect.bind(this) }>Disconnect</a>
+            </div>
+        </div>
+      );
+    } else {
+      info = (
+        <div className="account-info">
+          <img src={ defaultAvatar } height="100" width="100" />
+            <div className="info-container">
+              <p>Guest Account</p>
```
ianb: Is the Guest Account notion in the wireframes? I don't have an opinion, but @youwenliang might.

server/src/pages/settings/view.js
```diff
@@ -0,0 +1,72 @@
...
+              <a className="account-buttons connect" href="/api/fxa-oauth/login">Connect</a>
+            </div>
+        </div>
+      );
+    }
+    return <div className="preferences">
+      <h1 className="header">Firefox Screenshots Settings:</h1>
+      <hr></hr>
```
ianb: Should be `<hr />`

server/src/users.js
```diff
@@ -174,12 +174,47 @@ exports.tradeCode = function(code) {
     if (res.statusCode >= 200 && res.statusCode < 300) {
       return body;
     }
+    mozlog.warn('fxa-tradecode-failed', {status: res.statusCode});
     throw errors.badToken();
   });
 };
 
+exports.disconnectDevice = function(deviceId) {
```
ianb: I'm pretty sure there's old functions in here that we should delete, from the old oauth flow.

static/css/settings.scss
```diff
@@ -0,0 +1,47 @@
...
+	background-size: $grid-unit $grid-unit;
+  background-repeat: no-repeat;
+  background-position: center center;
+  @include flex-container(row, flex-start, stretch);
+  position: absolute;
+  right: 0;
+  margin: 40px 80px auto;
+}
+
+.preferences {
```
ianb: There seems to be inconsistent indentation

server/src/pages/settings/view.js
niharikak101: @ianb it is - https://mozilla.invisionapp.com/share/6DCDAQPES#/screens/236667833

niharikak101:
The second commit takes care of part two of #3080 (but this still needs some adjustment)

docs/METRICS.md
```diff
@@ -225,6 +225,14 @@ These are events that an add-on user can encounter on a shot they own
 7. [x] Click download from tile: `web/download/myshots-tile`
 8. [x] Clear search with button: `web/clear-search/button`
 
+#### Settings (My Shots)
+
+1. [x] Click connect button `web/start-connect/settings`
```
ianb: Is there an event when the login completes?  Seems like there should be.  It might have to be sent from the server instead of the web.  (There's other examples in the server of sending GA events.)

server/db-patches/patch-16-17.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE devices DROP COLUMN avatarurl;
+ALTER TABLE devices DROP COLUMN nickname;
+ALTER TABLE accounts ADD COLUMN avatarurl text;
+ALTER TABLE accounts ADD COLUMN nickname text;
```
ianb: We should be saving the email too

server/src/pages/settings/controller.js
```diff
@@ -0,0 +1,32 @@
...
+  let url = model.backend + "/api/disconnect-device";
+  let req = new XMLHttpRequest();
+  req.open("POST", url);
+  req.setRequestHeader("content-type", "application/x-www-form-urlencoded");
+  req.onload = function() {
+    if (req.status >= 300) {
+      // FIXME: a lame way to do an error message
+      window.alert("Error disconnecting: " + req.status + " " + req.statusText);
+    } else {
+      location.href = model.backend + "/settings";
```
ianb: Maybe we should send an event here?  I don't feel that strongly though.

server/src/pages/settings/model.js
```diff
@@ -0,0 +1,25 @@
+const db = require("../../db");
+
+exports.createModel = function(req) {
+  let title = "Settings";
+  let model = {
+    title,
+    docTitle: title,
+  };
+  return db.select(`
+    SELECT accounts.avatarurl, accounts.nickname, accounts.email FROM accounts, devices
```
ianb: This refers to `accounts.email` but that's not in the migration.  I'm guessing your database got out of sync with the migrations.

server/src/pages/shot/model.js
```diff
@@ -23,7 +23,7 @@ exports.createModel = function(req) {
     id: req.shot.id,
     productName: req.config.productName,
     isExtInstalled: !!req.deviceId,
-    isOwner: req.deviceId == req.shot.ownerId,
+    isOwner: req.deviceId == req.shot.ownerId || (req.accountId && req.accountId == req.shot.accountId),
```
ianb: This is a much better solution to ownership than I suggested!

ianb: Which is good, because what I had in my mind was crazy complicated.

server/src/pages/shotindex/view.js
```diff
@@ -105,6 +106,12 @@ class Body extends React.Component {
     );
   }
 
+  renderSettingsPage() {
+    return (
+      <a className="button preferences" href="/settings"></a>
```
ianb: This should have an `aria-label` on it. (Maybe other things too? It's something I only barely remember to do, but we should be aware of.)

server/src/server.js
```diff
@@ -570,10 +582,13 @@ app.post("/api/login", function(req, res) {
         deviceId: vars.deviceId,
         userAbTests
       };
+      let accountId = retrieveAccount(vars.deviceId).then((accountId) => {
+        sendParams.accountId = accountId;
+      });
```
ianb: This looks really racy – if this doesn't complete before `Show.checkOwnership` then none of this will work.

This should be chained with the sendParamsPromise

server/src/server.js
```diff
@@ -497,6 +505,7 @@ window.location = ${redirectUrlJs};
 
 app.post("/api/register", function(req, res) {
   let vars = req.body;
+  let accountId = linkAccount(vars.deviceId);
```
ianb: Where does this function come from? What does it do? Something seems odd.

server/src/server.js
```diff
@@ -682,6 +697,21 @@ app.post("/api/delete-shot", csrfProtection, function(req, res) {
   });
 });
 
+app.post("/api/disconnect-device", csrfProtection, function(req, res) {
+  if (!req.deviceId) {
+    sendRavenMessage(req, "Attempt to disconnect without login");
+    simpleResponse(res, "Not logged in", 401);
+    return;
+  }
+  disconnectDevice(req.deviceId).then((result) => {
```
ianb: How does this change the accountId cookie?  It should delete that cookie, otherwise it looks like you'll still have the associated account for the current session.

server/src/server.js
```diff
@@ -694,11 +724,15 @@ app.post("/api/set-title/:id/:domain", csrfProtection, function(req, res) {
     simpleResponse(res, "Not logged in", 401);
     return;
   }
-  Shot.get(req.backend, shotId, req.deviceId).then((shot) => {
+  Shot.get(req.backend, shotId).then((shot) => {
     if (!shot) {
       simpleResponse(res, "No such shot", 404);
       return;
     }
+    if (req.accountId && shot.accountId != req.accountId) {
```
ianb: It looks like this replaces the deviceId permission check with an accountId permission check – but as a result drops the deviceId check so that a guest user can't edit their title.

server/src/servershot.js
```diff
@@ -433,17 +435,37 @@ Shot.getRawValue = function(id, deviceId) {
...
       deleted: row.deleted,
-      blockType: row.block_type
+      blockType: row.block_type,
+      accountId: row.accountid
     };
   });
 };
 
-Shot.checkOwnership = function(shotId, deviceId) {
+Shot.checkOwnership = function(shotId, deviceId, accountId) {
```
ianb: I think this can be much simpler, like:

```sql
SELECT data.id
FROM data, devices
WHERE data.id = $1
    AND data.deviceid = device.id
    AND (device.id = $2 OR device.accountid = $3)
```

server/src/servershot.js
```diff
@@ -526,34 +548,69 @@ Shot.getShotsForDevice = function(backend, deviceId, searchQuery) {
 };
 
 Shot.setExpiration = function(backend, shotId, deviceId, expiration) {
-  if (expiration === 0) {
+  return db.select(
```
ianb: ditto here, I think this can be a much simpler SQL query

server/src/servershot.js
```diff
@@ -526,34 +548,69 @@ Shot.getShotsForDevice = function(backend, deviceId, searchQuery) {
...
-     WHERE id = $2
-           AND deviceid = $3
-    `,
-    [expiration, shotId, deviceId]
-  );
+  })
 };
 
 Shot.deleteShot = function(backend, shotId, deviceId) {
-  return Shot.get(backend, shotId, deviceId)
```
ianb: and here too (adding accountId to the method signature)

static/css/settings.scss
ianb: indentation is still inconsistent

static/css/shot-index.scss
```diff
@@ -261,6 +261,16 @@ h1 {
   }
 }
 
+.preferences {
```
ianb: 4 spaces instead of 2

server/src/pages/settings/controller.js
niharikak101: We have a couple when the user clicks on disconnect, cancels or confirms, so we could skip this I think. @ianb

server/src/pages/settings/model.js
niharikak101: Done!

ianb:
Looks great! This was a big task, you did really well.

youwenliang:
@ian @niharikak101 How can I test out the Fxa integration here? I've pulled the master branch but didn't see the fxa option in my shot page.

PR #3088, opened 2017-07-04
Fix #3029, about:home is not treated like about:newtab

### LINKED ISSUE #3029: about:home is not treated like about:newtab
When you try to take a shot on about:newtab it takes you to My Shots, but it doesn't work the same for about:home

gvanwaelvelde:


6a68:
@gvanwaelvelde Hi, thanks for the contribution! Works for me, merging :+1:

PR #3109, opened 2017-07-10
add context fill icons

johngruen:
Please note, this is dependent on https://bugzilla.mozilla.org/show_bug.cgi?id=1379464


6a68:
added a lil comment so this doesn't get merged in a flurry of PR queue clearing

ianb:
Landing details have been figured out

msujaws:
Why did the filenames have to change to add "-v2"? We should have been able to keep the old filenames, right?

johngruen:
@msujaws yes most likely, but i was encoutering some caching issues with the old icons while testing so i changed the names in order to pick up the diff.

msujaws:
probably needed to run `./mach build` instead of `./mach build faster` to fix the caching issue. i've seen that before with changes to the pocket system addon

ianb:
Our normal development process sideloads the add-on via `web-ext`, so @johngruen doesn't typically use mach.

johngruen:
@msujaws jaws, maybe so. would you prefer i revert the names or leave as is?

msujaws:
well, it's your project so do as you wish, but building version numbers in to filenames is doing a job that our version control system already does :)

johngruen:
got it. let's keep it as since it's make-work to revert it now. The -v2 will stand as a memorial of what's happened here :)

PR #3126, opened 2017-07-14
Shot annotations

niharikak101:
I made a really basic highlighter tool - it's a work in progress.


niharikak101:
I added a pen tool and removed the `mix-blend-mode` because I figured the best move would be to alter opacity, but it doesn't look as crisp as the `mix-blend-mode`.

niharikak101:
The last update adds some other functionality like an eraser, a better looking highlighter and rectangles!
Next Steps: Making the shapes movable and resizable.

server/src/editor.js
```diff
@@ -0,0 +1,202 @@
...
+  onClickRectangle() {
+    this.setState({tool: 'rectangle'});
+  }
+
+  componentDidMount() {
+    editor = document.getElementById("editor");
+    context = this.refs.editor.getContext('2d');
+    highlighter = document.getElementById("highlighter");
+    highlightContext = this.refs.highlighter.getContext('2d');
+    let imageCanvas = document.getElementById("image-holder");
```
ianb: There's a React way to get this node, maybe `this.ref.imageHolder` (and then `ref="imageHolder"`) on the element

niharikak101:
Ready for review @ianb

server/src/editor.js
```diff
@@ -0,0 +1,140 @@
...
+      highlightContext.strokeStyle = '#ff0';
+      drawContext = highlightContext;
+    } else if (this.state.tool == 'pen') {
+      context.strokeStyle = this.state.color;
+      highlightContext.strokeStyle = this.state.color;
+      context.globalCompositeOperation = 'source-over';
+      drawContext = context;
+    }
+    context.lineWidth = this.state.size;
+    document.querySelector("#canvas-container").addEventListener("mousemove", this.draw);
```
ianb: There needs to be a `this.draw.bind(this)` here, a function without a bind will lose its association with `this`. (Sometimes React fixes this, but not always.)  *Actually*, in `constructor()` there should probably be something like `this.draw = this.draw.bind(this)`, because function identity is going to be important for removing these functions.

Generally there needs to be some more careful lifecycle tracking here. I don't think multiple handlers will be too easy, because the `#canvas-container` element will change when these need to change. But still, things get weird when handlers don't get cleaned up.

Also these should use the refs system like other elements.

server/src/editor.js
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  setPosition(e) {
+    var rect = editor.getBoundingClientRect();
+    pos.x = e.clientX - rect.left,
+    pos.y = e.clientY - rect.top
+  }
+
+  draw(e) {
+    if (e.buttons !== 1) return;
```
ianb: Probably this should also check for modifier keys (ctrl/etc), and do nothing in that case.

server/src/editor.js
```diff
@@ -0,0 +1,140 @@
...
+    drawContext.beginPath();
+
+    drawContext.lineCap = 'round';
```
ianb: I wonder if a vertical bar would be more appropriate than a circle? Eh, probably not what people expect.

server/src/pages/shot/controller.js
```diff
@@ -116,6 +116,27 @@ exports.deleteShot = function(shot) {
   req.send(`id=${encodeURIComponent(shot.id)}&_csrf=${encodeURIComponent(model.csrfToken)}`);
 };
 
+exports.saveEdit = function(shot, shotUrl) {
+  let url = model.backend + "/api/save-edit";
+  let req = new XMLHttpRequest();
```
ianb: We should use the fetch API for any new code: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

server/src/pages/shot/view.js
```diff
@@ -388,6 +394,7 @@ class Body extends React.Component {
 
     return (
       <reactruntime.BodyTemplate {...this.props}>
+        { this.state.imageEditing ? this.renderEditor(clip) : null }
```
ianb: Seems odd this would render the entire page, plus the editor...?  I'd expect it to render the editor *or* the normal page.

server/src/pages/shot/view.js
```diff
@@ -436,6 +444,22 @@ class Body extends React.Component {
     </div>;
   }
 
+  onClickEdit() {
```
ianb: There should be an event on this

server/src/server.js
```diff
@@ -743,6 +743,30 @@ app.post("/api/set-title/:id/:domain", csrfProtection, function(req, res) {
...
+app.post("/api/save-edit", csrfProtection, function(req, res) {
+  let vars = req.body;
+  if (!req.deviceId) {
+    sendRavenMessage(req, "Attempt to set expiration without login");
+    simpleResponse(res, "Not logged in", 401);
+    return;
+  }
+  let id = vars.shotId;
+  let url = vars.url;
+  Shot.get(req.backend, id).then((shot) => {
```
ianb: There's no permission check here. There should be a deviceId/accountId permission check as part of `Shot.get()`

server/src/pages/shot/view.js
niharikak101: The highlighter and pen both call the edit function as soon as they are clicked. It's just a general set up for both the tools, and not a tool in itself. Does it still need an event? @ianb

server/src/editor.js
```diff
@@ -0,0 +1,140 @@
...
+
+let pos, context, editor, highlightContext, highlighter, drawContext;
+
+exports.Editor = class Editor extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      tool: 'none',
+      highlighter: 'inactive',
+      pen: 'inactive',
```
niharikak101: The highlighter and the pen states are more for the state of their respective icons than anything else, but I'm sure I can figure out a way to have only one tool state. @ianb

niharikak101:
Ready for review! @ianb

server/src/pages/shot/view.js
niharikak101: My bad, added an event here!

server/src/pages/shot/editor.js
```diff
@@ -0,0 +1,150 @@
...
+      highlighterState = 'inactive';
+      penState = 'active';
+    } else {
+      penState = 'inactive';
+      highlighterState = 'inactive';
+    }
+    return <div>
+      <div className="editor-header default-color-scheme">
+        <div className="shot-main-actions">
+          <a className={`button pen-button ${penState}`} id="pen" onClick={this.onClickPen.bind(this)} title="pen"></a>
```
ianb: I'd generally prefer these to be buttons, since that's semantically more correct.

server/src/pages/shot/editor.js
```diff
@@ -0,0 +1,150 @@
...
+    let dataUrl = this.imageCanvas.toDataURL();
+    this.props.onClickSave(dataUrl);
+  }
+
+  onClickHighlight() {
+    if (this.state.tool != 'highlighter') {
+      this.setState({tool: 'highlighter'});
+      sendEvent("highlighter-select", "annotation-toolbar");
+    } else {
+      this.setState({tool: 'none'});
```
ianb: Does it make sense for there ever to be no tool? Seems like we're creating a state that is going to be confusing: the user wants to edit, they clicked edit, and now they are effectively just looking at an image and clicks do nothing.

If you can't have "no" tool active, then I'd imagine that clicking a button always activates that tool.

ianb: This article talks about a somewhat different area, but it's a concept I go back to often in my mind: http://limi.net/checkboxes-that-kill/ – or generally, don't create flows that can lead someone to a broken (or in this case just pointless) destination

server/src/pages/shot/editor.js
```diff
@@ -0,0 +1,150 @@
...
+      this.canvasContainer.addEventListener("mousemove", this.draw);
+      this.canvasContainer.addEventListener("mousedown", this.setPosition);
+      this.canvasContainer.addEventListener("mouseenter", this.setPosition);
+    }
+  }
+
+  setPosition(e) {
+    var rect = this.editor.getBoundingClientRect();
+    pos.x = e.clientX - rect.left,
+    pos.y = e.clientY - rect.top
```
ianb: Using a global for position seems... funny. I think it could be safely added as `this.pos`?  I don't think the component will get recreated anytime when we want to keep the position.

server/src/pages/shot/editor.js
```diff
@@ -0,0 +1,150 @@
...
+  edit() {
+    pos = { x: 0, y: 0 };
+    if (this.state.tool == 'highlighter') {
+      this.highlightContext.lineWidth = 20;
+      this.highlightContext.strokeStyle = '#ff0';
+      this.drawContext = this.highlightContext;
+    } else if (this.state.tool == 'pen') {
+      this.context.strokeStyle = this.state.color;
+      this.context.globalCompositeOperation = 'source-over';
+      this.drawContext = this.context;
```
ianb: I think the tool should be handled in `draw()` not here – so instead of swapping `this.drawContext` here, instead do it later.

server/src/pages/shot/editor.js
```diff
@@ -0,0 +1,150 @@
...
+    var rect = this.editor.getBoundingClientRect();
+    pos.x = e.clientX - rect.left,
+    pos.y = e.clientY - rect.top
+  }
+
+  draw(e) {
+    if (e.buttons !== 1) {
+      return null;
+    }
+    this.drawContext.beginPath();
```
ianb: Per previous comment, I think this would be better like: `let context = this.state.tool == "pen" ? this.context : this.highlightContext` (or as a switch or something)

server/src/server.js
```diff
@@ -743,6 +743,30 @@ app.post("/api/set-title/:id/:domain", csrfProtection, function(req, res) {
...
+  if (!req.deviceId) {
+    sendRavenMessage(req, "Attempt to set expiration without login");
+    simpleResponse(res, "Not logged in", 401);
+    return;
+  }
+  let id = vars.shotId;
+  let url = vars.url;
+  Shot.get(req.backend, id, req.deviceId, req.accountId).then((shot) => {
+    if (!shot) {
+      simpleResponse(res, "No such shot", 404);
```
ianb: We should send a raven message here too, as it should be mostly impossible to get to this state (except maybe to delete a shot in a different tab while you are editing).

server/src/pages/shot/editor.js
niharikak101: I figure we could start with a tool active, but the user still has the ability to deselect all tools by clicking on an activated tool button.

Issue #3137, opened 2017-07-17
Update full page save button to match selection save button


johngruen:
@niharikak101 

<img width="234" alt="screen shot 2017-07-17 at 12 29 34 pm" src="https://user-images.githubusercontent.com/3323249/28264519-a74c1cc4-6aeb-11e7-87ec-6b17cd9e56ac.png">

This needs to be updated to look like the one on top:
<img width="308" alt="screen shot 2017-07-17 at 12 29 41 pm" src="https://user-images.githubusercontent.com/3323249/28264518-a74aff7e-6aeb-11e7-943e-37a9c06a9300.png">

johngruen:
This is done

PR #3138, opened 2017-07-17
update shot preview save icon

niharikak101:
Fix for #3137


johngruen:
Hey @niharikak101, it looks like that by recycling that class you are failing a test. I might keep the original class instead and just breakdown the CSS by shared styles and individual styles.

I'm a bit concerned that changing class names might mess up event pings as well

PR #3144, opened 2017-07-18
Update original URL link handling

niharikak101:
Update the original link format (Bugzilla #1381189)
I added a ga event to track outbound clicks, is this the correct way to do it?


6a68:
Tests passed, merging

PR #3148, opened 2017-07-19
Bug 1366026 - add an item to the library panel that opens shots page

gijsk:
This will also require some changes to Firefox which I will put on the bug.

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -17,6 +17,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "Services",
                                   "resource://gre/modules/Services.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "LegacyExtensionsUtils",
                                   "resource://gre/modules/LegacyExtensionsUtils.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "CustomizableUI",
+                                  "resource:///modules/CustomizableUI.jsm");
 
 let addonResourceURI;
 let appStartupDone;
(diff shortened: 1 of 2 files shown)


gijsk:
@6a68  is there anything else I need to do to get this to be reviewed / merged?

6a68:
@gijsk Nope, ran into some unplanned server debugging today. Looking now

addon/bootstrap.js
```diff
@@ -59,7 +61,59 @@ const appStartupObserver = {
   }
 }
 
+const LibraryButton = {
+  ITEM_ID: "appMenu-library-screenshots",
+
+  init(webExtension) {
+    this._initialized = true;
+    let permissionPages = [...webExtension.extension.permissions].filter(p => (/^https?:\/\//i).test(p));
```
6a68: It would be cool if this worked for local development as well, where the server URL [is `http://localhost:10080`](https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/manifest.json.template#L66). Could you change the regex?

addon/bootstrap.js
```diff
@@ -59,7 +61,59 @@ const appStartupObserver = {
...
+    this.PAGE_TO_OPEN = permissionPages.length ? permissionPages[0].replace(/\*$/, "") : "https://screenshots.firefox.com/";
+    this.PAGE_TO_OPEN += "shots";
+    this.ICON_URL = webExtension.extension.getURL("icons/icon-16-v2.svg");
+    this.ICON_URL_2X = webExtension.extension.getURL("icons/icon-32-v2.svg");
```
6a68: I'm not sure how the photon library UI works (I haven't seen mocks or specs), but right now, in the toolbar button, we [show a special icon](https://github.com/mozilla-services/screenshots/blob/9959edeff2b9c8a91916ad63037cecc92438860e/addon/webextension/background/startBackground.js#L55-L60) to users who haven't tried Screenshots before, and we [swap in the regular icon](https://github.com/mozilla-services/screenshots/blob/9959edeff2b9c8a91916ad63037cecc92438860e/addon/webextension/background/main.js#L53-L57) after they complete onboarding. (Note that startBackground runs first, lazy-loading the rest of the webextension, including main.js.)

If this UI is replacing the toolbar button UI, I think we'd want to keep that special icon behavior for new users.

If this item is in addition to some photon-specific "take a shot" UI, I'm not sure what the flow should be. If a user hasn't tried Screenshots before, I think we'd want to show them the onboarding flow, and maybe not open a link to the shots page. @johngruen, thoughts?

I'm also not sure we'd want to reuse the same icon for multiple commands related to the same feature (take a shot vs visit shots page).

addon/bootstrap.js
```diff
@@ -125,6 +180,10 @@ function start(webExtension) {
 
 function stop(webExtension, reason) {
   webExtension.shutdown(reason);
+
+  if (reason != APP_SHUTDOWN) {
```
6a68: Is there any reason to not call `uninit` if Firefox is shutting down?

locales/en-US/webextension.properties
```diff
@@ -53,3 +53,5 @@ termsAndPrivacyNotice2 = By using Firefox Screenshots, you agree to our {termsAn
 termsAndPrivacyNoticeTermsLink = Terms
 # This string is used as the text for a link in termsAndPrivacyNoticeCloudServices
 termsAndPrivacyNoticyPrivacyLink = Privacy Notice
+# This string is used to label the item in the Library panel that opens the screenshots page
+libraryLabel = Screenshots
```
6a68: Everywhere else in the product, we use the phrase 'My Shots' to refer to the user's shots page. @johngruen thoughts?

addon/bootstrap.js
gijsk: Why wouldn't the regex work? `(/^https?:\/\//i).test("http://localhost:10080")` in my JS console returns `true`. :-)

addon/bootstrap.js
gijsk: > I'm not sure how the photon library UI works (I haven't seen mocks or specs), but right now, in the toolbar button, we show a special icon to users who haven't tried Screenshots before, and we swap in the regular icon after they complete onboarding. (Note that startBackground runs first, lazy-loading the rest of the webextension, including main.js.)

https://mozilla.invisionapp.com/share/ENBBK0F9U#/screens/229252053

Screenshot of mock

> If this UI is replacing the toolbar button UI, I think we'd want to keep that special icon behavior for new users.

It's not replacing the toolbar button. We will replace the toolbar button with the page action ( https://bugzilla.mozilla.org/show_bug.cgi?id=1366041 ) which is a separate bug.

> If this item is in addition to some photon-specific "take a shot" UI, I'm not sure what the flow should be. If a user hasn't tried Screenshots before, I think we'd want to show them the onboarding flow, and maybe not open a link to the shots page. @johngruen, thoughts?
> 
> I'm also not sure we'd want to reuse the same icon for multiple commands related to the same feature (take a shot vs visit shots page).

That also works for me, I think it's a discussion to have with Aaron Benson - I'm just doing what the mock says to do right now. We might even just want to not show the screenshots item until we know the user has taken at least 1 screenshot (but I don't know how to determine that, or to determine we need to show the different icon / different page. Presumably the website could just show the onboarding flow if it's opened and the user has no shots? That would be the simplest from my perspective...

addon/bootstrap.js
gijsk: It's just not necessary - all `uninit` does is remove a bunch of DOM nodes, but if we're shutting down all the windows go away so there's no point. Part of the quantum/photon work is making sure startup and shutdown are as quick as possible, so I avoid doing the busy-work here.

It might be that some of the webextension shutdown stuff also doesn't need to happen if we're shutting down, but I haven't examined that in detail and it's kind of orthogonal to the rest of this PR.

locales/en-US/webextension.properties
gijsk: Again, I just went with what the mock said to do. Happy to make changes. Aaron Benson probably has more context here. However, I will say this:

- All the other items in this menu ("Bookmarks", "History", "Downloads", "Synced Tabs", "Pocket List") have no possessive. I think it might be odd if this one stood out, and given we are consistent about not using possessives elsewhere in Firefox menus, I don't think we should change that here.
- Just 'Shots' with no context I think would be less clear (ie as a user, I'd wonder what that was...) than "Screenshots"

johngruen: No... it should say Screenshots

PR #3150, opened 2017-07-19
Add locales to Docker app

6a68:
@relud Maybe this is why the locales are missing from dev right now?

The server code in /server/src/l10n.js looks for the locales at "../../locales", which I think corresponds to the one-liner added to the Dockerfile.

Not sure how to test this, as I haven't gotten the Docker image working locally.


relud:
lgtm

PR #3151, opened 2017-07-20
Fix homePageCapturePageDescription for en-US

flodolo:
Either is missing a "button", or it should not have a "the" before. Since there are already two "button" in the string, I opted for removing "the".


flodolo:
Thanks @TonnesM for noticing

PR #3154, opened 2017-07-20
L20n syntax subset

Pike:
The fluent serializers sadly don't deal gracefully with the full fluent syntax spec.

While @stasm is on vacation, let's make the server.ftl conform with the support subset.

I haven't found a better way than running the file through tools/fluentfmt.py in python-fluent, and even that only works with a local edit to serialize to utf-8.

@mathjazz, can we test this on stage in some sort?

### DIFF
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -8,23 +8,23 @@
 gMyShots = My Shots
 gHomeLink = Home
 gNoShots
-  .alt = No shots found
+    .alt = No shots found
 gScreenshotsDescription = Screenshots made simple. Take, save, and share screenshots without leaving Firefox.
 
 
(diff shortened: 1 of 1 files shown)


Pike:
Also, review request on @6a68 once we had some more eyes on this from the pontoon side.

flodolo:
I tested this on Pontoon stage, forking the repository in mozilla-l10n/screenshots and adding these changes, then trying to commit translations. Everything is working as expected.

flodolo:
@6a68 I'll go ahead and merge this one, so we can re-enable sync on Pontoon production and have a few more European hours to look out for issues.

6a68:
@flodolo @Pike Thanks for writing & merging this! Just saw it in my bugmail backlog

PR #3176, opened 2017-07-21
Adds an event for non-owner click on original url

### LINKED ISSUE #3146: Add event for a non-owner click on original url
There's an event defined in metrics `goto-original-url` for non-owner clicks on the original url link, that isn't used anywhere. Do we want to track owner/non-owner clicks seperately? Right now there's only one event `view-original` that is used to track both.

niharikak101:
Fixes #3146

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -465,7 +465,11 @@ class Body extends React.Component {
   }
 
   onClickOrigUrl(label) {
-    sendEvent("view-original", label, {useBeacon: true});
+    if (this.props.isOwner) {
+      sendEvent("view-original", `${label}-owner`, {useBeacon: true});
+    } else {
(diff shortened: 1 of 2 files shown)


server/src/pages/shot/view.js
```diff
@@ -465,7 +465,11 @@ class Body extends React.Component {
   }
 
   onClickOrigUrl(label) {
-    sendEvent("view-original", label, {useBeacon: true});
+    if (this.props.isOwner) {
+      sendEvent("view-original", label, {useBeacon: true});
+    } else {
+      sendEvent("goto-original-url", label, {useBeacon: true});
+    }
```
ianb: Usually we use the label to specify owner or non-owner. Since we're passing in a label, let's just append it there, like `label + "-owner"`, and leave the action ("view-original") alone.

Also this should have a METRICS.md update

PR #3186, opened 2017-07-21
Fix #3178, hide site metadata when editing the expiration

### LINKED ISSUE #3178: "Select time", "Save" and "Cancel" buttons are not fully visible if the browser is resized
[Notes]:
- The new string "How long should this shot be retained?" is longer then the previous one, which makes the issue easy to encounter.
- Also, the saved shot title is not entirely visible.

[Affected versions]:
- Nightly 56.0a1
- Beta 55.0b11
- Screenshots 11.0.0

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have the latest Screenshots stage build installed on a new clean profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to any website.
2. Perform a shot on the website and save it.
3. Resize the browser window to be half of the screen.
4. Click the "expires in 14 days" button.
5. Observe the "Select", "Save" and "Cancel" buttons.
 
[Expected result]:
- The buttons are fully visible.

[Actual result]:
- The "Select", "Save" and "Cancel" buttons are not entirely visible.

[Additional Notes]:
- The issue is reproducible on dev and stage servers.
- Also, if the site domain is very long, the issue is much easier to reproduce.
- I have logged a similar issue regarding the mobile view in #2127. Probably if that issue will be fixed this one will no longer be reproducible.   
expires buttons

ianb:
With site metadata hidden there is much more room for the editing widget and explanation string

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -182,7 +182,8 @@ class Body extends React.Component {
     super(props);
     this.state = {
       hidden: false,
-      closeBanner: false
+      closeBanner: false,
+      isChangingExpire: false
     };
   }
 
@@ -333,6 +334,7 @@ class Body extends React.Component {
       expiresDiff = <span className="expire-widget">
       <ExpireWidget
         expireTime={this.props.expireTime}
+        onChanging={this.onChangingExpire.bind(this)}
         onSaveExpire={this.onSaveExpire.bind(this)} />
       </span>;
     }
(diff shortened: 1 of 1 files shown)


server/src/pages/shot/view.js
```diff
@@ -393,9 +395,11 @@ class Body extends React.Component {
           <div className="shot-main-actions">
             <div className="shot-info">
               <EditableTitle title={shot.title} isOwner={this.props.isOwner} />
-              <div className="shot-subtitle"> { favicon }
-                { linkTextShort ? <a className="subtitle-link" rel="noopener noreferrer" href={ shot.url } target="_blank" onClick={ this.onClickOrigUrl.bind(this, "navbar") }>{ linkTextShort }</a> : null }
-                <span className="time-diff">{ timeDiff }</span> { expiresDiff }
+              <div className="shot-subtitle">
+                { this.state.isChangingExpire ? null : favicon }
+                { linkTextShort && !this.state.isChangingExpire ? <a className="subtitle-link" rel="noopener noreferrer" href={ shot.url } target="_blank" onClick={ this.onClickOrigUrl.bind(this, "navbar") }>{ linkTextShort }</a> : null }
+                { this.state.isChangingExpire ? null : <span className="time-diff">{ timeDiff }</span> }
```
6a68: What if, instead of repeating three `this.state.isChangingExpire` checks, we wrap these items in a div that can have the `hidden` attribute toggled if `this.state.isChangingExpire`?

Fine to land as-is, though.

ianb: Yeah, I played with that, but some part of the CSS doesn't like another intermediate element here, so I decided to take the easy way out.

PR #3191, opened 2017-07-21
Merge v10

### LINKED ISSUE #3142: Merge back divergent changes in addon release branch
Currently, the `v10.4.0-release` branch has some changes that haven't made their way to master.

This bug is a reminder to merge back to master.

ianb:
Fixes #3142

The ui.js merge was a little hairy, extra attention to that file would be welcome.


addon/webextension/selector/ui.js
```diff
@@ -275,14 +277,16 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
     },
 
     hide() {
-      window.removeEventListener("scroll", this.onScroll);
+      window.removeEventListener("scroll", watchFunction(assertIsTrusted(this.onScroll)));
```
6a68: I didn't catch this before, but just as with `this.onResize`, if we wrap `this.onScroll` inline, removeEventListener will fail to detach the listener. It doesn't lead to any thrown errors, but is a memory leak. I'm fine to address this here or in a followup bug.

ianb: I'll make a followup

Issue #2208, opened 2017-02-22
Server shouldn't listen for HTTP until after initialization


ianb:
Right now we're doing too much at the top level of server.js.  We should do database connection, Keygrip initialization, and any other initialization before listening for HTTP.

Really all code at the top level of server.js should be in functions, and then we should have some launch() function that puts them all together.

ianb:
This issue is being closed (without being fixed) because [we are discontinuing the Screenshots server](https://blog.mozilla.org/futurereleases/2019/01/24/clarifying-the-future-of-firefox-screenshots/).

PR #3199, opened 2017-07-24
When initing l10n, cache contents of FTL files, not user locale (#3170)

### LINKED ISSUE #3170: [l10n] seeing the wrong language on dev
My Firefox languages are en-US and en, but I"m seeing this on dev:

6a68:
Fixes #3170 for me locally

### DIFF
--- a/server/src/l10n.js
+++ b/server/src/l10n.js
@@ -6,18 +6,16 @@ const negotiateLanguages = require("fluent-langneg/compat");
 const { MessageContext } = require("fluent/compat");
 const mozlog = require("./logging").mozlog("l10n");
 
-let inited = false;
-
 const rawStrings = {};
 
+let initPromise;
 exports.init = function() {
-  if (inited) {
-    return Promise.resolve();
+  if (initPromise) {
+    return initPromise;
   }
-  inited = true;
   const localesGlob = path.join(__dirname, "..", "..", "locales") + path.normalize("/*/server.ftl");
 
-  return globby(localesGlob).then(paths => {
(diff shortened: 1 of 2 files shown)


server/src/server.js
```diff
@@ -291,21 +291,18 @@ app.use(function(req, res, next) {
...
-  l10n.init(languages).then(() => {
-    req.getText = l10n.getText;
-    req.userLocales = l10n.userLangs;
-    return l10n.getStrings().then(strings => {
-      req.messages = strings;
-      next();
-    });
-  }).catch(err => {
-    mozlog.info("l10n-error", {msg: "Error loading FTL files", description: err});
+  l10n.init().then(() => {
```
ianb: This (`l10n.init()`) is weird in a handler, but without #2208 I guess it's the best we got.

server/src/l10n.js
```diff
@@ -6,98 +6,84 @@ const negotiateLanguages = require("fluent-langneg/compat");
...
 
-exports.init = function(userLocales = []) {
+const rawStrings = {};
+
+exports.init = function() {
   if (inited) {
     return Promise.resolve();
   }
   inited = true;
-  exports.userLangs = userLocales;
```
ianb: there's a possibility of a dogpile here, if multiple requests come in early. It should sort itself out, but I think it would be safer to do:

```
let initPromise;
exports.init = function() {
  if (initPromise) {
    return initPromise;
  }
  initPromise = globby(...);
  return initPromise;
};
```

PR #3208, opened 2017-07-25
Remove full-page & visible buttons and remove from onboarding

6a68:


msujaws:
I assume this was because of issue 3182?

PR #3222, opened 2017-07-26
Gracefully handle toolbar button mashing

### LINKED ISSUE #3097: Multiple errors are thrown while opening and closing Firefox Screenshots
[Affected versions]:
Firefox 55.0b7 (Build Id:20170706085221)
Firefox 56.0a1 (Build Id:20170709030204)
Firefox Screenshots version: 10.3.0

[Affected platforms]:
Windows 10 64 bit
macOS 10.11.6
Ubuntu 16.04 64bit

[Steps to reproduce]:

1. Launch Firefox with a clean profile.
2. Enable Firefox Screenshots.
3. Open Firefox Screenshots.
4. Skip the Firefox Screenshots tutorial.
5. Close Firefox Screenshots
6. Access the http://www.bbc.com/ webpage.
7. Click on any available article.
8. Immediately click the Firefox Screenshots button several times and fast (while the article loads and after the article has been successfully loaded).

[Expected result]:
Firefox Screenshots opens and closes without any errors.

[Actual result]:
The following errors are constantly displayed:

Windows 10 and Ubuntu 16.04
"No matching message handler"
"catcher is not defined"
"ui is not defined"

macOS 10.11.6:
"We can't screenshot this page" is constantly thrown even if the page was successfully loaded. Also the "catcher is not defined" error is thrown in the console.

[Additional notes]
Please note that this issue is not reproducible at all times (you may need to repeat steps 7 and 8).
Also please note that sometimes the Firefox Screenshots layer remains permanently on the screen and if you try to close/open Firefox Screenshots the "ui is not defined error is constantly thrown".

For further information regarding this issue please observe the following video: https://goo.gl/VLtRgY

6a68:
This patch consistently fixes #3097 for me. Two parts:

1. When loading selector code into a page, manually concatenate all scripts, then kick off 1 `browser.tabs.executeScript` request. The old approach kicked off N `executeScript` promises sequentially, one for each script.

2. Double-check the list of already-loading tabs just before starting to load code into a tab. Although this check is done elsewhere already, adding a check at the moment of truth seems to fix the remaining few weird timing bugs that lead to 'haywire' notifications.

### DIFF
--- a/addon/webextension/background/selectorLoader.js
+++ b/addon/webextension/background/selectorLoader.js
@@ -64,6 +64,11 @@ this.selectorLoader = (function() {
   let loadingTabs = new Set();
 
   exports.loadModules = function(tabId, hasSeenOnboarding) {
+    // Sometimes loadModules is called when the tab is already loading.
+    // Adding a check here seems to help avoid catastrophe.
+    if (loadingTabs.has(tabId)) {
+      return Promise.resolve(null);
+    }
     let promise;
     loadingTabs.add(tabId);
     if (hasSeenOnboarding) {
--- a/addon/webextension/background/selectorLoader.js
+++ b/addon/webextension/background/selectorLoader.js
@@ -80,21 +80,43 @@ this.selectorLoader = (function() {
     });
   };
 
-  function executeModules(tabId, scripts) {
-    let lastPromise = Promise.resolve(null);
-    scripts.forEach((file) => {
-      lastPromise = lastPromise.then(() => {
-        return browser.tabs.executeScript(tabId, {
-          file,
-          runAt: "document_end"
-        }).catch((error) => {
-          log.error("error in script:", file, error);
-          error.scriptName = file;
-          throw error;
-        });
+  function readFile(path) {
+    return fetch(path, { mode: 'same-origin' }).then((response) => {
+      return response.blob();
+    }).then(blob => {
+      let reader = new FileReader();
+      return new Promise((resolve, reject) => {
+        reader.readAsText(blob);
+        reader.onload = (evt) => {
+          let contents = evt.target.result;
+          resolve(contents);
+        };
+        reader.onerror = (evt) => {
+          reject(evt.target.error);
+        };
       });
     });
-    return lastPromise.then(() => {
+  }
+
+  function executeModules(tabId, scripts) {
+    let filePromises = [];
+    scripts.forEach((file) => {
+      let fileURL = browser.extension.getURL(file);
+      let fileContents = readFile(fileURL);
+      filePromises.push(fileContents);
+    });
+    return Promise.all(filePromises).then(fileContents => {
+      let allScripts = fileContents.reduce((concatted, script) => { return concatted + ' ; ' + script });
+      return Promise.resolve(allScripts);
+    }).then(scripts => {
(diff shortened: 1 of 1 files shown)


addon/webextension/background/selectorLoader.js
```diff
@@ -80,21 +85,43 @@ this.selectorLoader = (function() {
...
-          throw error;
-        });
-      });
+      let fileURL = browser.extension.getURL(file);
+      let fileContents = readFile(fileURL);
+      filePromises.push(fileContents);
     });
-    return lastPromise.then(() => {
+    return Promise.all(filePromises).then(fileContents => {
+      let allScripts = fileContents.reduce((concatted, script) => { return concatted + ' ; ' + script });
```
ianb: Couldn't this be `fileContents.join("; ")` ?

addon/webextension/background/selectorLoader.js
```diff
@@ -64,6 +64,11 @@ this.selectorLoader = (function() {
   let loadingTabs = new Set();
 
   exports.loadModules = function(tabId, hasSeenOnboarding) {
+    // Sometimes loadModules is called when the tab is already loading.
+    // Adding a check here seems to help avoid catastrophe.
+    if (loadingTabs.has(tabId)) {
```
ianb: Won't this get in the way of loading modules on the same tab, when it's a new page? E.g., just reload the page and try to re-invoke screenshots, and then this check will stop it from loading...?

addon/webextension/background/selectorLoader.js
```diff
@@ -80,21 +85,43 @@ this.selectorLoader = (function() {
     });
   };
 
+  function readFile(path) {
+    return fetch(path, { mode: 'same-origin' }).then((response) => {
+      return response.blob();
```
ianb: Can this be `response.text()` instead of `response.blob()`, skipping the FileReader stuff?

addon/webextension/background/selectorLoader.js
ianb: One bummer here is that we no longer get filenames or line positions. This is going to make debugging and Sentry much harder to work with.

I guess we could do a sourcemap? Though if we do, then using something like browserify (and doing this all as a make step) might be advantageous. I did a quick lookup of the sourcemap format, since maybe for pure concatenation it isn't that hard, but it looks pretty complicated. Though *maybe* it's not that complicated, since it looks like something like `{offset: {line: 10}, map: {file: "aScript.js"}}}` might work?

ianb:
After discussion, some options to explore:

1. Generate the sourcemap as part of the concatenation, generating them manually (appending the sourcemap as a data URL)
1. Create a build step that concatenates files, maybe using [concat-with-sourcemaps](https://www.npmjs.com/package/concat-with-sourcemaps)
1. Create a build step that uses browserify, hacking our primitive module system to work with browserify
1. Create a build step that uses browserify, and update our module system to use CommonJS modules (i.e., require, exports)

If we want to make minimal logic changes, then we should build two files, one that includes slides.js/etc, and one that does not. Right now slides.js (onboarding) runs based on whether the files are included. For the last option (where we use browserify properly) we probably have to change how onboarding is invoked. Maybe we can just use a query string on the script name?

6a68:
I'm now thinking this is the wrong approach. Closing

Issue #2935, opened 2017-05-31
"Share" and "Flag" buttons from the saved shot page preview are not working on Edge browser


SoftVision-CarmenFat:
[Affected versions]:
- Microsoft Edge 38.14393.1066.0
- Screenshots 8.2.1970

[Affected Platforms]:
- All Windows

[Steps to reproduce]:
1. Open Edge browser.
2. Navigate to a someone else's shot (https://screenshots.dev.mozaws.net/Uehl0P90TGlR08jh/ro.pinterest.com).
3. Observe the way the image is loaded and displayed.
4. Click on the "Flag" button and observe what happens next.
5. Click on the "Share" button and observe what happens next.

[Expected result]:
3. The saved shot image is loaded and correctly displayed in the center of the page. 
4. The mail client is opened in order to report the shot.
5. The "Share" pop-up is opened.

[Actual result]:
3. Only part of a stuck spinner is displayed.
4. The mail client is not opened.
5. The "Share" pop-up is not opened.

[Notes]:
- Please see the following screencast:
![edge](https://cloud.githubusercontent.com/assets/26000308/26631513/003291bc-4614-11e7-9949-2a79fe972468.gif)

SoftVision-CarmenFat:
Retested this issue today and here are the results:
- the image is now loaded and correctly displayed, the stuck spinner is no longer shown when opening someone's else's shot;
- still, the issues mentioned at steps 4&5 are still reproducible, nothing happens when clicking the "Flag" and "Share" buttons.

Reopening this issue due to the fact that "Flag" and "Share" buttons are clickable, but not actionable.

6a68:
@SoftVision-CarmenFat I'm seeing expected behavior with the 'flag' and 'share' buttons on dev for the link in the description(https://screenshots.dev.mozaws.net/Uehl0P90TGlR08jh/ro.pinterest.com).

Would you mind taking another look? If you're still seeing problems, could you give details on the exact versions of windows and edge?

SoftVision-CarmenFat:
@6a68
Tested this again on multiple versions of Edge and here are the results:
- on older versions (25.10586.672.0 and 38.14393.1066.0), this issue is still reproducible;
- on the latest version (40.15063.0.0), this is **fixed**: the buttons work as expected.

Tested on Windows 10 Pro x64 bit.

johngruen:
Note that metrics suggests that these buttons are not widely used.

6a68:
[...]

PR #3238, opened 2017-07-28
Load shot image directly

6a68:
TL;DR: moves shot image to a plain old image tag, making the shot page functional in IE, Edge, and any browser with JS disabled.

The shot image wasn't loading in Edge or IE (#2935) because those browsers don't support the CSP nonce directive. (The very latest version of Edge, released this week with the Windows 10 Creators Update, does support CSP level 2.) I'm not sure why the regular image loading detection code failed, but the CSP error explains why the inline fallback code failed.

This seemed like a good opportunity to get rid of the code that hides the image until it's loaded. Simplifies the code and makes the pages accessible without JS.

I set the spinner as a centered background image of the image tag itself; it's weird-looking now, but once jgruen replaces the svg with a simple png, should look good.

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -36,7 +36,7 @@ class Clip extends React.Component {
       return null;
     }
     let backgroundStyle = "transparent url('" + this.props.staticLink("/static/img/spinner.svg") + "') center no-repeat";
-    let node = <img id="clipImage" style={{height: "auto", width: clip.image.dimensions.x + "px", maxWidth: "100%", display: this.state.imageDisplay, background: backgroundStyle }} ref={clipImage => this.clipImage = clipImage} src={ clip.image.url } alt={ clip.image.text } />;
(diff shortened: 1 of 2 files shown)


ianb:
Looks like there's a conflict.

Do you want to land this as-is and let John fix it up, or have him add to this branch?

ianb:
That was a through-the-web conflict resolution, so I haven't actually tested it.

6a68:
@ianb Rebased against current master, and it seems to be working in Edge. Mind taking another look?

server/src/pages/shot/view.js
```diff
@@ -41,34 +35,16 @@ class Clip extends React.Component {
       console.warn("Somehow there's a shot without an image");
       return null;
     }
-    let node = <img id="clipImage" style={{height: "auto", width: clip.image.dimensions.x + "px", maxWidth: "100%", display: this.state.imageDisplay}} ref={clipImage => this.clipImage = clipImage} src={ clip.image.url } alt={ clip.image.text } onLoad = { this.onImageLoaded.bind(this) } />;
-    // Note that in server/src/pages/shot/page.js there is also JavaScript defined
-    // that displays the image onload, as a backup to make sure the image always
-    // gets displayed even if the bundle doesn't load
+    let backgroundStyle = "transparent url('" + this.props.staticLink("/static/img/spinner.svg") + "') center no-repeat";
```
ianb: This could go in a stylesheet, but eh, whatever

PR #3242, opened 2017-07-31
Cleanup  empty l10n files, and add l10n.toml

Pike:
Two commits here, one cleaning up empty l10n files, and one adding compare-locales configuration file l10n.toml, which allows to check for errors etc.


Pike:
@ianb has a couple of related tickets, I think. r?

Pike:
CC @flodolo and @mathjazz so that they know

ianb:
I've opened #3247 about linting locales. Though... I'm not sure exactly what to do with lint errors. It feels like Pontoon basically manages our locale files (for better or worse).

Anyway, these changes seem correct, even if linting would make them more useful.

PR #3258, opened 2017-08-01
Fix #2410, use 'Screenshot: title' for the <title> of shot pages

### LINKED ISSUE #2410: Design page title
Right now we adopt the title of the page the shot was taken from.  Seems like it should be "Screenshot: {title}" or something like that.

ianb:


locales/en-US/server.ftl
```diff
@@ -81,6 +81,8 @@ notFoundPageDescription = Page not found.
 
 [[ Shot page ]]
 
+# This is the page title, put into the page <title></title>
```
flodolo: Wrong format for the comment. Make sure to run compare-locales locally ;-)

Also, I would simplify it: "This is used for the HTML title tag of the page

flodolo: For example 

```
$ compare-locales l10n.toml . en-US
en-US/server.ftl
    ERROR: Unparsed content "# This is the page title, put into the page <title></title>" from line 84 column 1 to line 84 column 60
    WARNING: Parser error in en-US
```

That's because comments start with //

ianb: Oops, I'm not keeping up with the new formats... fixed now

PR #3269, opened 2017-08-03
Fix S3

ianb:
The bugs weren't big, the tests were more complicated...


6a68:
@ianb Looks like a flake8 error and some addon test flakiness. Good otherwise

Issue #3268, opened 2017-08-03
Header / footer links are blue on shots page


6a68:
Is this intentional? 

header on shots page:

<img src="https://screenshots.firefoxusercontent.com/images/6dd331eb-7050-4177-9d44-2f3f744e683f.png" height="100">

footer on shots page:

<img src="https://screenshots.firefoxusercontent.com/images/328ddfa2-75f3-419b-a0d6-6e1d97b88f20.png" height="80">

wresuolc:
Let's do whatever is on the spec

PR #3270, opened 2017-08-03
Update fonts and onboarding svgs

6a68:
@johngruen This one's probably for you :-)

- Remove the full page / save visible step from the onboarding page
- Tweak the onboarding SVGs to make the Save button look more realistic
  - Note, the second image is scaled, so the button looks funny nonetheless :-\
- Use the Photon standard font-family incantation everywhere
  - Aside from tweaking font-weights in the onboarding headers, I haven't done much twiddling here. Could use some love; feel free to land tweaks into this branch.


6a68:
This might also be a good opportunity to fix the shots page headers / footer links being blue instead of black (#3268).

Also, the SVG logo on the homepage apparently uses Helvetica shapes instead of text, and I wasn't up for creating a replacement. We can file a followup bug, maybe?

6a68:
Old SVG / new SVG, onboarding image 2:




old / new, image 4:

6a68:
I think the onboarding image removal is #3270, fixed by this PR.

niharikak101:
Added a fix for #3268

static/css/shot-index.scss
```diff
@@ -105,6 +105,10 @@ h1 {
     color: $dark-default;
   }
 
+  a {
+    color: $dark-default;
```
6a68: @niharikak101 Hmm, it seems funny that the style above this, `.default-color-scheme & a`, isn't getting applied.

Looking at the git history for the shot index page view, it looks like the `default-color-scheme` class may have been [accidentally removed](https://github.com/mozilla-services/screenshots/commit/1efbd956d6397be8c44887d8e47c81a57d981e6f#diff-d45944c551eead471a128c9a716e2a79L34).

And, after trying it out locally, it looks like adding the `default-color-scheme` classname to the landing page fixes the footer color issues on that page, too.

Would you mind removing your changes and adding that classname to the outermost divs in the views for the shotindex and homepage? You can just force-push your changes; ask if you have any questions.

niharikak101:
@6a68 Does this look okay?

johngruen:
@niharikak101 @6a68 Talked to Taipei team, they are firmly against switching to Fira here for the time being, but we should extend our font stack to include system fonts.

6a68:
@johngruen Updated, mind taking another look?

@niharikak101 looks ok to me, @johngruen may have some feedback

johngruen:
Yeah, let me pull this down and check it out

johngruen:
WFM

Issue #3147, opened 2017-07-19
The loading spinner is not animated when the image is creating (dev)


SoftVision-CosminMuntean:
[Affected versions]:
- Beta 55.0b9
- Nightly 56.0a1

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" boolean pre is set to "false".
- Have a custom Screenshots dev build installed on a new Firefox profile.

[Steps to reproduce]:
1. Start the browser with the profile from prerequisites.
2. Navigate to any website.
3. Click on the Screenshots icon from browser toolbar.
4. Perform a selection and click the "Save" button.
5. Observe the loading spinner when the image is creating.

[Expected result]:
- The loading spinner is correctly animated.

[Actual results]:
- The loading spinner is not animated and is stuck until the image is displayed.

[Notes]:
- The issue is only reproducible on dev server.
- Also, there is no spinner on "Creating..." page.
- Here is a screen shot with the issue:
![loading spinner1](https://user-images.githubusercontent.com/19226534/28374118-7b3d8768-6cac-11e7-985e-8abbe0a9a5b2.gif)

SoftVision-CosminMuntean:
The behavior is changed a little and the loading spinner is still not displayed on "Creating..." page.
Here is a screen recording with the current behavior:
![spinner 2](https://user-images.githubusercontent.com/19226534/29027535-c018bb96-7b89-11e7-9c70-2861cb552ebb.gif)

johngruen:
Fixed on dev

PR #3285, opened 2017-08-07
replace svg loader with pure css

johngruen:


johngruen:
Fixes #3147

server/src/config.js
```diff
@@ -133,7 +133,7 @@ var conf = convict({
   // This is mostly configurable for debugging purposes:
   checkDeletedInterval: {
     doc: "Frequency in seconds to check for items that should be purged",
-    format: Number,
+    format: "int",
```
6a68: Looks like you are reverting a change that's landed in master. Probably need to fetch master and rebase.

PR #3289, opened 2017-08-07
fix colors in headers and footers

johngruen:
Quick fix to tidy up header and footer colors everywhere.


6a68:
@johngruen Niha already has a fix for this in the svgs-and-fonts branch, want to provide some guidance in that PR?

PR #3293, opened 2017-08-09
Update locales v10.10.0

flodolo:
This takes localization from master and update the v10.10.0 branch

Discussion happening over email.


johngruen:
@flodolo commented in that email, but the failures caused by commit 6863352 make this branch significantly more risky. I'd recommend resubmitting w/o that commit, and then handling that separately.

flodolo:
> @flodolo commented in that email, but the failures caused by commit 6863352 make this branch significantly more risky. I'd recommend resubmitting w/o that commit, and then handling that separately.

I've checked, there are no commits passing tests on the v10.10.0 branch since Jul 20
https://github.com/mozilla-services/screenshots/commits/v10.10.0

While the diff on this patch is big, it only touches localized FTL files (and doesn't touch en-US), so I don't see how it can make tests fail.

ianb:
Tests have been failing because we removed the Full Page functionality, which was used by the tests. So yes, the failure is expected/OK

PR #3302, opened 2017-08-09
Detach app startup observer on shutdown

### LINKED ISSUE #3301: Detach sessionstore observer on shutdown (bug 1387598)
The addon waits for the `sessionstore-windows-restored` signal to be fired before fully initializing. However, if that signal never fires, as happens in [a specific Talos test](https://dxr.mozilla.org/mozilla-central/source/testing/talos/talos/test.py#182), then we never detach the observer on shutdown. This apparently can sometimes cause a crash on Windows!

Refs https://bugzilla.mozilla.org/show_bug.cgi?id=1387598

6a68:
One of the Talos tests prevents this event from firing, leading to
intermittent crashes on Windows (https://bugzil.la/1387598).

Fixes #3301.

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -135,6 +135,7 @@ function startup(data, reason) { // eslint-disable-line no-unused-vars
 
 function shutdown(data, reason) { // eslint-disable-line no-unused-vars
   prefObserver.unregister();
+  appStartupObserver.unregister();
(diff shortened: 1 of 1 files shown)


6a68:
Closing: the actual fix may be in the JS interpreter code, see discussion in #3301

Issue #3217, opened 2017-07-26
Send image to server in binary


ianb:
Right now we send the image as a base64 data URL. It would save some space to send it in binary.

Because we send a JSON chunk, we don't have a good technique to handle this currently. The body won't be able to be JSON. I guess we could use [netstrings](https://en.wikipedia.org/wiki/Netstring) or something – one string for the JSON, another for the binary.  Or [bencode](https://en.wikipedia.org/wiki/Bencode)? ([Library](https://github.com/benjreinhart/bencode-js))

To make it "easy" to reconstruct the JSON from the request body, maybe we could use a pattern where we generate UUIDs for each binary attachment, put the UUID in the JSON body, and substitute the binary in for the UUID. A little MIMEish.

We'd still want to support the current PUT bodies indefinitely.

Issue #3218, opened 2017-07-26
Send image from content to background as binary


ianb:
It might be worth the savings to decode the image from base64 and send it as a binary blob.

To implement this alone (before #3217), we'd still have to encode the image back into base64 before uploading.  As a result it might be best to do #3217 first or at the same time.

PR #3308, opened 2017-08-10
Send image to server in binary

niharikak101:
Fix for #3217 
This converts the binary blob back to base64 before saving it to the database, which is what I'm a little unsure about. @ianb

### DIFF
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -52,6 +52,9 @@ const { errorResponse, simpleResponse, jsResponse } = require("./responses");
 const selfPackage = require("./package.json");
 const { b64EncodeJson, b64DecodeJson } = require("./b64");
 const { l10n } = require("./middleware/l10n");
+const multer = require("multer");
+const storage = multer.memoryStorage();
+const upload = multer({storage});
 
 const PROXY_HEADER_WHITELIST = {
   "content-type": true,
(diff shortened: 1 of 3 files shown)


niharikak101:
The second commits adds a fix for #3218.

server/src/server.js
```diff
@@ -576,7 +579,7 @@ app.post("/api/login", function(req, res) {
   });
 });
 
-app.put("/data/:id/:domain", function(req, res) {
+app.put("/data/:id/:domain", upload.single('blob'), function(req, res) {
```
ianb: This doesn't appear to work with our test script, but I'm guessing also doesn't work with an old version of the add-on (which we'll need to support).

It's possible this could use a different endpoint, or maybe there's a way to use the upload functionality not as middleware.

ianb: After thinking about it, just having the `upload.single` middleware probably doesn't break the tests, but something else.

server/src/server.js
```diff
@@ -585,7 +588,11 @@ app.put("/data/:id/:domain", function(req, res) {
     res.end();
     return;
   }
-  let bodyObj = req.body;
+  let bodyObj = JSON.parse(req.body.shot);
+  let clipId = Object.getOwnPropertyNames(bodyObj.clips)[0];
+  let b64 = req.file.buffer.toString("base64");
+  b64 = "data:image/png;base64," + b64;
+  bodyObj.clips[clipId].image.url = b64;
```
ianb: This is where it breaks. Probably `req.file` won't be defined if there's no multipart upload? In which case the image should have a clip image already.

ianb:
The tests are still failing, though the failure is... peculiar. I rebased this against master to see if it was a regression from the docTitle change, but that didn't change things.

The error seems to be a failure to actually set the title properly, which implies something went wrong with the JSON part of the upload. The title `Screenshot: Screenshot` is, I think, the default title when there's no other title available.

PR #3309, opened 2017-08-10
clean up home

johngruen:
This issue does a few things: 
- Replaces all screenshots icons on the home page with the current
- removes `no-warn` error from `sass-lint` (this rule triggers a warning on our responsive mixin)
- changes the globbing pattern for sass linting so that partials are linted, and corrects all linter errors and warnings we previously did not see
- moves homepage CSS to it's own file in the css directory with all other files
- lightly modifies home page styles to better handle variable widths
- corrects broken button states and images for the Firefox download button


johngruen:
Note, filed #3310 as an eventual follow up

johngruen:
Looks like this is doing that nightly fail thing :)

static/css/home.scss
```diff
@@ -0,0 +1,507 @@
...
+    bottom: 0;
+    right: -10%;
+  }
+
+  section {
+    min-height: 500px;
+  }
+}
+
+@media (max-width: 768px) {
```
pdehaan: Not sure if you want to use the `@include respond-to("small") {` style helpers for consistency.
🤷‍♀️

static/css/home.scss
```diff
@@ -0,0 +1,507 @@
...
+#section-2 .section-image,
+#section-4 .section-content {
+  order: 1;
+}
+
+#section-1 .section-image,
+#section-2 .section-content,
+#section-4 .section-image {
+  order: 2;
+}
```
pdehaan: Not sure if I'm missing something, but wouldn't this just overrule what you did on lines 316-320 above?

PR #3317, opened 2017-08-10
Fix #3306, make tests resilient to a browserAction or pageAction

### LINKED ISSUE #3306: Tests broken due to Photon-related regression in Nightly
The tests are broken today, also today when you start the add-on you get the toolbar button and no longer get the pageAction. 

Something changed with Nightly and should be fixed, so we get the pageAction. But also our tests should look for the button in either location and work either way.

ianb:
Also the label changed for unknown (but presumably unimportant) reasons

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -35,7 +35,8 @@ test:
     - ./bin/load_test_exercise.py http://localhost:10080
     # run zap baseline against the server
     - docker pull owasp/zap2docker-weekly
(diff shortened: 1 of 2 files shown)


ianb:
We got some failed test collision here, the danger of having failing tests and then not noticing failed-harder tests. Anyway, this DOES fix the browser test that was broken.

Issue #3323, opened 2017-08-11
On Firefox shutdown, immediately shut down webextension


6a68:
When Firefox is exiting, the webextension should be immediately shutdown.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1386333#c17

PR #3330, opened 2017-08-12
Export shutdown fix

6a68:
Bringing the fix for #3323 over to the export branch for uplift


6a68:
Already got an R+ for the original patch; merging

PR #3332, opened 2017-08-14
validate original url

### LINKED ISSUE #3297: Fix Bug 1388781
See [Bug 1388781](https://bugzilla.mozilla.org/show_bug.cgi?id=1388781)

niharikak101:
Fixes #3297 (Bugzilla: 1388781)

### DIFF
--- a/server/src/servershot.js
+++ b/server/src/servershot.js
@@ -6,6 +6,7 @@ const linker = require("./linker");
 const config = require("./config").getProperties();
 const fs = require("fs");
 const mozlog = require("./logging").mozlog("servershot");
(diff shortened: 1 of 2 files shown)


ianb:
Ideally this would actually go in `shot.js` – then if `shot.url` gets used in My Shots or any other context, we'll still be safe.

Annoyingly `shot.js` runs in multiple contexts, and makes it harder to use `valid-url`. There's a pattern for base64 encoding at the bottom of `servershot.js`, and we could use something similar there – maybe using a simple validation approach, and then in the servershot.js subclass putting in the `require("valid-url").isWebUri()` check.

PR #3340, opened 2017-08-14
Fix #3333, make run-addon work with new legacy pref

### LINKED ISSUE #3333: ./bin/run-addon --bootstrap has stopped working
I think this is the same legacy add-on issue as the failed tests, probably the same pref has to be added to the web-ext invocation.

ianb:
Unfortunately while Screenshots loads up with this patch (instead of being disabled), it doesn't work in Nightly due to regressions.

### DIFF
--- a/bin/run-addon
+++ b/bin/run-addon
@@ -165,8 +165,10 @@ run_jpm() {
   echo "Running Firefox.  Reloading with jpm *will not* happen"
   echo "If the Screenshots icon doesn't show up go to:"
   echo "  Tools > Add-ons > Extensions and look for an error loading the add-on"
+  settings_file="./build/.jpm-settings.json"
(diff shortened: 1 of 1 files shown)


bin/run-addon
```diff
@@ -165,8 +165,10 @@ run_jpm() {
   echo "Running Firefox.  Reloading with jpm *will not* happen"
   echo "If the Screenshots icon doesn't show up go to:"
   echo "  Tools > Add-ons > Extensions and look for an error loading the add-on"
+  settings_file="./build/.jpm-settings.json"
+  echo '{"extensions.legacy.enabled": true}' > "$settings_file"
```
6a68: Since we're in here, do you want to also set `xpinstall.signatures.required` to false?

6a68:
@ianb One optional question, feel free to self-merge. Works for me.

PR #3343, opened 2017-08-14
Deal with lack of MOZ_PHOTON_THEME on 57+

### LINKED ISSUE #3342: Screenshots is not using pageAction in Nightly
Sometime recently, I think due to changes in Firefox, Screenshots has reverted back to using a toolbar button and is not showing the pageAction.

The Library still contains the Screenshots item.

/cc @0c0w3

gijsk:
@ianb does this fix #3342


ianb:
Gijs confirmed the pref is defined back to 55, so this should safely keep us as a toolbar button in our older supported versions: https://dxr.mozilla.org/mozilla-release/source/toolkit/modules/AppConstants.jsm#365

PR #3348, opened 2017-08-15
empty docTitle value in JSON data

niharikak101:
Fix for [Bug 1389706](https://bugzilla.mozilla.org/show_bug.cgi?id=1389706)


ianb:
I thought this was a funny way to do it, but after thinking about it I like it – this way we can bring back the title, and effectively we're doing the private/public data approach we eventually want to get to, except for now treating everything like public.

ianb:
Now I don't know if I misread the commit or somehow it got applied in an odd way, but it's just removing the title entirely

PR #3353, opened 2017-08-15
Tweak header management middleware

### LINKED ISSUE #3299: Fix Bug 1388780
See [Bug 1388780](https://bugzilla.mozilla.org/show_bug.cgi?id=1388780)

6a68:
Fixes #3299

### DIFF
--- a/server/src/pages/leave-screenshots/server.js
+++ b/server/src/pages/leave-screenshots/server.js
@@ -4,7 +4,7 @@ const reactrender = require("../../reactrender");
 const { Shot } = require("../../servershot");
 const mozlog = require("../../logging").mozlog("leave-screenshots");
 
-const csrfProtection = csrf({cookie: true});
+const csrfProtection = csrf({cookie: {httpOnly: true, secure: true, sameSite: 'lax', key: '__Host-csrf'}});
 let app = express();
 
 exports.app = app;
--- a/server/src/pages/leave-screenshots/server.js
+++ b/server/src/pages/leave-screenshots/server.js
@@ -1,10 +1,9 @@
 const express = require("express");
-const csrf = require('csurf');
 const reactrender = require("../../reactrender");
 const { Shot } = require("../../servershot");
 const mozlog = require("../../logging").mozlog("leave-screenshots");
+const { csrfProtection } = require("../../middleware/csrf");
 
-const csrfProtection = csrf({cookie: {httpOnly: true, secure: true, sameSite: 'lax', key: '__Host-csrf'}});
 let app = express();
 
 exports.app = app;
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -23,7 +23,7 @@ const dbschema = require("./dbschema");
 const express = require("express");
 const bodyParser = require('body-parser');
 const contentDisposition = require("content-disposition");
-const csrf = require("csurf");
+const { csrf, csrfProtection, csrfErrorHandler } = require("./middleware/csrf");
 const morgan = require("morgan");
 const linker = require("./linker");
 const { randomBytes } = require("./helpers");
@@ -99,8 +99,6 @@ function initDatabase() {
 }
 
 initDatabase();
-
-const csrfProtection = csrf({cookie: {httpOnly: true, secure: true, sameSite: 'lax', key: '__Host-csrf'}});
 
 const app = express();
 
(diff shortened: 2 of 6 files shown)


6a68:
@g-k Any comments?

server/src/server.js
```diff
@@ -219,7 +219,17 @@ app.use(function(req, res, next) {
     req.accountId = authInfo.accountId;
   }
   req.cookies = cookies;
-  req.cookies._csrf = cookies.get("_csrf"); // csurf expects a property
+  // The cookies library doesn't detect duplicates; check manually
+  let rawCookies = req.get("cookie") || "";
+  let pairs = rawCookies.split(";");
+  let csrfTokens = pairs.filter(item => { return item.match(/__Host-csrf=/); });
```
pdehaan: Since you're using an arrow function here anyways, I think you can lose the `{`, `return`, and `; }` bits:

```js
const csrfTokens = pairs.filter(item => item.match(/__Host-csrf=/));
```

server/src/server.js
```diff
@@ -219,7 +219,17 @@ app.use(function(req, res, next) {
...
+  if (csrfTokens.length > 1) {
+    let exc = new Error("Duplicate CSRF cookies");
+    exc.headerValue = rawCookies;
+    captureRavenException(exc);
+    simpleResponse(res, "Bad request", 400);
+  }
+  req.cookies._csrf = cookies.get("__Host-csrf"); // csurf expects a property
```
pdehaan: nit: Should that be `// csurf` or `// csrf`?

server/src/server.js
g-k: This also matches cookies like "foo__Host-csrf=", but that shouldn't matter since duplicates are handled below.

server/src/server.js
6a68: csurf, it's a node package. good eye though :+1:

g-k: I think csurf is right.

Confusingly enough we're using the csurf module which uses the csrf module.

server/src/server.js
```diff
@@ -101,7 +101,7 @@ function initDatabase() {
 
 initDatabase();
 
-const csrfProtection = csrf({cookie: true});
+const csrfProtection = csrf({cookie: {httpOnly: true, secure: true, sameSite: 'lax', key: '__Host-csrf'}});
```
g-k: We're initializing multiple csrf middleware in multiple places, so I think we need to update the configs there too or have them all share one config or middleware layer.

```
» git log -1 --format=oneline
12197ff596cf9514c24813e2aef2314481b59016 (HEAD -> master, origin/master) Merge pull request #3358 from j605/fix-shebang
» git grep -n csrf\(
server/src/pages/leave-screenshots/server.js:7:const csrfProtection = csrf({cookie: true});
server/src/pages/settings/server.js:9:app.get("/", csrf({cookie: true}), function(req, res) {
server/src/pages/shot/server.js:12:app.get("/:id/:domain", csrf({cookie: true}), function(req, res) {
server/src/pages/shotindex/server.js:11:app.get("/", csrf({cookie: true}), function(req, res) {
server/src/server.js:104:const csrfProtection = csrf({cookie: true});   # the one fixed here
```

6a68: ooh, nice catch. I'll pull that config out to a module.

6a68:
Updated. @ianb r?

server/src/server.js
6a68: module seemed too heavyweight, just search/replaced inline

server/src/pages/leave-screenshots/server.js
```diff
@@ -4,7 +4,7 @@ const reactrender = require("../../reactrender");
 const { Shot } = require("../../servershot");
 const mozlog = require("../../logging").mozlog("leave-screenshots");
 
-const csrfProtection = csrf({cookie: true});
+const csrfProtection = csrf({cookie: {httpOnly: true, secure: true, sameSite: 'lax', key: '__Host-csrf'}});
```
ianb: Seems like we shouldn't repeat this. Can we make a `csrf` module (maybe confusing to call it that) that includes the `csrfProtection` object, and maybe a portion of that cookie handling?

6a68: Sure, I can make that change. It seemed like overkill to move a one-liner into a module, but it's also not satisfying to spread this long incantation all over.

ianb: Well... I thought the code that splits out the csrf cookie would fit nicely into that module too. Unfortunately the other modules can't just import it from server due to circular dependencies, and none of the other modules seem right.

g-k:
I rebased this and have been working on getting it passing.

Appears to be failing due to the `secure: true`, so I'm planning to set the secure attr based on config expectProtocol being https like: https://github.com/mozilla-services/screenshots/blob/f300d64cbf0b251c3a47610fdb13cee26f1ceede/server/src/server.js#L127

6a68:
@g-k Hey, I'm having a bit of trouble getting this to work in the case of changing shot expiration. It seems like the secret, not the whole token, is getting set in the query string, so that when `csurf` passes just the secret to `csrf`, and `csrf` checks the secret against secret + salt, the CSRF token always comes back as invalid. I've tried moving the token check inside a `value` function, but that didn't help. Would you mind taking a look?

6a68:
closing in favor of #3428

PR #3363, opened 2017-08-16
Server tests et al

ianb:


ianb:
This turned into a bucket of things, but I'll cut it off here for review

server/src/db.js
```diff
@@ -133,3 +144,49 @@ exports.markersForArgs = function(starting, numberOfArgs) {
...
+function initTiming() {
+  if (!logQueryLimit) {
+    return doNothing;
+  }
+  let caller = getCallerPosition(2);
+  if (caller == "skip") {
+    // This happens when getCallerPosition detects we shouldn't time this function call
+    return doNothing;
+  }
+  let start = Date.now();
```
6a68: Maybe [`process.hrtime`](https://nodejs.org/api/process.html#process_process_hrtime_time) instead, to get a high resolution timestamp?

6a68:
just curious, what is the purpose of tracking image sizes in the database?

ianb:
For future reporting: during some performance discussions we realized we didn't know how big (in bytes) images were, and certainly couldn't see if some users had a particularly large set of images.

server/src/db.js
```diff
@@ -115,9 +124,11 @@ exports.transaction = function(func) {
 exports.del = exports.update;
 
 exports.exec = function(sql, args) {
+  let timer = initTiming();
   return getConnection().then(function([client, done]) {
```
6a68: Would it be useful to track slow connection times as well? Maybe shove a second timer in here?

server/src/db.js
```diff
@@ -133,3 +144,49 @@ exports.markersForArgs = function(starting, numberOfArgs) {
...
+
+function doNothing() {
+}
+
+function initTiming() {
+  if (!logQueryLimit) {
+    return doNothing;
+  }
+  let caller = getCallerPosition(2);
+  if (caller == "skip") {
```
6a68: I'm a little bit confused by this getCallerPosition-related stuff, it seems a little dangerously clever.

Could we instead issue the timing calls from the files that _call into_ `db.js`, like `servershot.js`? That might be an opportunity to pass a label into `initTiming` to identify the query, which would make it directly greppable. Would also get us moving in the direction of adding timing for arbitrary other events.

server/src/servershot.js
```diff
@@ -856,6 +856,10 @@ Shot.upgradeSearch = function() {
             return resolve();
           }
           Shot.get("upgrade_search_only", rows[index].id).then((shot) => {
+            // This shouldn't really happen, but apparently can...
```
6a68: Is this related to another bug, or just something unfiled that you ran into?

ianb:
It also has occurred to me that I added `npm run test:server`, but it doesn't run in CI

server/src/db.js
ianb: As I think about it, I somehow remember that we removed connection pooling. But I'm not sure. It might be good for a followup, but I'd be surprised if we saw anything interesting.

server/src/db.js
ianb: I guess I was trying to make this minimal, without adding a bunch of stuff throughout the codebase. Given that, it required some cleverness to identify queries (there might have been an option to identify queries based on the SQL, but then there's a question of how to uniquely identify SQL without putting every query into the log).

But I don't think this should be forever code... like, this code isn't Mr. Right, just Mr. Right Now.

server/src/servershot.js
ianb: Something weird in my local database, I don't know if it was due to testing or some latent bug I didn't notice before.

server/src/db.js
ianb: I looked at process.hrtime, but it splits the time into two numbers, and seemed complicated to work with. We don't really care about anything that's less than (or even close to) 1ms anyway.

server/src/db.js
6a68: lol ok

PR #3364, opened 2017-08-17
seperate script blocks

niharikak101:
Fix for [1389704](https://bugzilla.mozilla.org/show_bug.cgi?id=1389704)

### DIFF
--- a/server/src/reactrender.js
+++ b/server/src/reactrender.js
@@ -46,17 +46,20 @@ exports.render = function(req, res, page) {
     } else {
       body = ReactDOMServer.renderToString(viewModule.BodyFactory(serverModel));
     }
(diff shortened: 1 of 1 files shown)


server/src/reactrender.js
```diff
@@ -46,17 +46,20 @@ exports.render = function(req, res, page) {
     } else {
       body = ReactDOMServer.renderToString(viewModule.BodyFactory(serverModel));
     }
+    let jsonString = JSON.stringify(jsonModel).replace(/\u2028/g, '\\u2028').replace(/\u2029/g, '\\u2029');
```
ianb: There was some code that would have applied to this before, but with the change won't anymore; from reactutils:

    addScript = addScript.replace(/<script/ig, "\\x3cscript").replace(/<\/script/ig, "\\x3c/script");

This is essentially another issue similar to `\u2028`, except worse, as code like this can happen:

    <script type="data">{"title": "</script>Now I get to write anything and it goes in the page!"}</script>

So we'll have to move this logic back in:

    addScript = addScript.replace(/<script/ig, "\\x3cscript").replace(/<\/script/ig, "\\x3c/script");

PR #3366, opened 2017-08-17
Set language default correctly, add server unit tests

6a68:
Deliberately making the unit test additions as small as possible here, while still providing diagnostic value.

Open question: should we change `npm test` to `npm-run-all test:*`?


6a68:
oops, testing the header incorrectly

PR #3367, opened 2017-08-17
Set language default correctly, add server unit tests

6a68:
(Thought I had goofed up the header mocking in the previous PR, but it seems correct.)

Deliberately making the unit test additions as small as possible here, while still providing diagnostic value.

Open question: should we change npm test to npm-run-all test:*?

### DIFF
--- a/server/src/middleware/l10n.js
+++ b/server/src/middleware/l10n.js
@@ -1,0 +1,28 @@
+require("../logging").installConsoleHandler();
+const mozlog = require("../logging").mozlog("l10n-middleware");
+const accepts = require("accepts");
+const l10n = require("../l10n");
+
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -87,6 +87,7 @@
     "nsp": "nsp check",
     "posttest": "npm run lint",
     "test": "make bootstrap_zip && mocha test/",
+    "test:server": "mocha --recursive test/server/unit/",
```
6a68: I'm happy to just back this line out and figure out the details later. I"ll file a followup bug.

ianb: It could just be renamed to test:server_unit or something. Or does `mocha test/ test/server/unit/` work?

PR #3370, opened 2017-08-17
Update fluent-langneg to 0.1.0

stasm:
See https://github.com/projectfluent/fluent.js/issues/61#issuecomment-323153344


ianb:
note: test failure is a byproduct of a server test problem that is fixed on master, it's not related to this PR

PR #3375, opened 2017-08-18
Send exceptions to raven when alert is called

### LINKED ISSUE #3303: Send exceptions to Raven whenever alert() is called
We have some places where we call alert() in the codebase, all generally unexpected fatal request errors. In each location we should also be sending information to Raven/Sentry.

niharikak101:
Fixes #3303

### DIFF
--- a/server/src/pages/settings/controller.js
+++ b/server/src/pages/settings/controller.js
@@ -18,6 +18,7 @@ exports.disconnectDevice = function() {
     if (req.status >= 300) {
       // FIXME: a lame way to do an error message
       window.alert("Error disconnecting: " + req.status + " " + req.statusText);
+      window.Raven.captureException(new Error(`Error disconnecting: ${req.status} ${req.statusText}`));
     } else {
       location.href = model.backend + "/settings";
     }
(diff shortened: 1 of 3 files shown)


server/src/pages/settings/controller.js
```diff
@@ -18,6 +18,7 @@ exports.disconnectDevice = function() {
     if (req.status >= 300) {
       // FIXME: a lame way to do an error message
       window.alert("Error disconnecting: " + req.status + " " + req.statusText);
+      window.Raven.captureException(new Error("Error disconnecting"));
```
ianb: This should have some more information. Probably it would be enough to have

    new Error(`Error disconnecting: ${req.status} ${req.statusText}`)

server/src/pages/shot/controller.js
```diff
@@ -209,6 +214,7 @@ exports.setTitle = function(title) {
     if (req.status >= 300) {
       let errorMessage = document.getElementById("shotPageAlertErrorUpdatingTitle").textContent;
       window.alert(errorMessage);
+      window.Raven.captureException(new Error(`${errorMessage}: ${req.status} ${req.statusText}`));
```
ianb: These should use something simpler than `${errorMessage}` – that's going to be localized, which will confuse things in Sentry.  Something like `Error calling /api/set-title: ${req.status} ${req.statusText}`

PR #3396, opened 2017-08-24
Update FxA UI

youwenliang:
Update FxA UI to match the visual spec:

image


ianb:
Merging, but with a followup in #3402

PR #3404, opened 2017-08-25
Fix review feedback for screenshots 16.0.0

gijsk:
@ianb I believe this works and addresses all the non-CSS non-how-to-hide-a-browser-action stuff.

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -71,7 +71,10 @@ const LibraryButton = {
   init(webExtension) {
     this._initialized = true;
     let permissionPages = [...webExtension.extension.permissions].filter(p => (/^https?:\/\//i).test(p));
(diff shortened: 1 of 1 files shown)


addon/bootstrap.js
```diff
@@ -256,7 +260,7 @@ function initPhotonPageAction(api) {
           type: "click",
           tab: {
             url: browserWin.gBrowser.selectedBrowser.currentURI.spec,
-            id: tabTracker.getId(browserWin.gBrowser.selectedTab),
+            id: tabManager.getWrapper(browserWin.gBrowser.selectedTab).id,
```
kmaglione: Ideally, we should probably do something like:

```js
  let tab = tabManager.getWrapper(browserWin.gBrowser.selectedTab);
  ...
    tab: { id: tab.id, url: tab.url },
```

ianb:
Updated in #3408

PR #3408, opened 2017-08-25
Fix review feedback for screenshots 16.0.0

ianb:
This is @gijsk's changes from #3404, plus the change @kmaglione suggested


ianb:
The test failure is because of some installation problem with eslint. I'm fairly confident it's unrelated to this change, so I'm going to move ahead on it

PR #3412, opened 2017-08-27
Fix #3241 : Moved buttons from right to left 

### LINKED ISSUE #3241: Improve the Save, Download & Cancel buttons for RTL
Current state:
current state

Desired state:
desired state

Changes made in the latter screenshot:
1. The buttons group was moved to the left side as is, underneath the frame (the order of the buttons is okay also in the current state)
2. The cloud icon within the "Save" button is located to the right of the button's text

ankita240796:
Changes:-
1. Shifted all three buttons (cancel,download,save) from right to left
2. Swapped the position of text and image on save button

### DIFF
--- a/static/css/inline-selection.scss
+++ b/static/css/inline-selection.scss
@@ -238,13 +238,13 @@ $very-light-grey: #ededed;
   @extend .button;
   @extend .primary;
   background-image: url("MOZ_EXTENSION/icons/cloud.svg");
-  background-position: 8px center;
+  background-position: 55px center;
   background-repeat: no-repeat;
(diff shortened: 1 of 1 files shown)


ankita240796:
Requested changes applied

ianb:
Thanks! The approach looks good, but the button doesn't appear to render quite right. I used an [Arabic build](https://www.mozilla.org/en-US/firefox/nightly/all/?q=arabic), and got this:

image

Note the cloud icon is slightly cut off on the right.

ankita240796:
Made the fix

PR #3415, opened 2017-08-28
Generate l10n Messages from ftl

chenba:
- Added a make target to generate the l10n message js files from the ftl files.
- The server now uses the js files instead of ftl.
- Script tags are used to include the messages in the page.
- The messages are no longer in the body of the page.


chenba:
@ianb and/or @6a68 Would appreciate review/feedback on this.  Thanks.

bin/build-scripts/ftl-to-js.js
```diff
@@ -0,0 +1,28 @@
...
+const fs = require('fs');
+const destDir = process.argv[2];
+const ftlPaths = process.argv.slice(3);
+
+for (let ftl of ftlPaths) {
+  let locale = ftl.split('/')[1];
+  let destFilePath = `${destDir}/${locale}.js`;
+  let ftlFileContent = fs.readFileSync(ftl, 'utf8');
+  let jsFileContent = `
+    (context => {
```
ianb: I'm not 100% sure where this file is built into, and then whether it gets translated with babel – but if it doesn't go through babel then the ES6 syntax here will cause problems on some browsers.

bin/build-scripts/ftl-to-js.js
```diff
@@ -0,0 +1,28 @@
...
+      let messages=${JSON.stringify(ftlFileContent)};
+      if (typeof exports === 'object' && context === exports) {
+        exports.messages = messages;
+      } else {
+        context.l10nMessages = context.l10nMessages || {};
+        context.l10nMessages['${locale}'] = messages;
```
ianb: It might be better to do `context.l10nMessages[${JSON.stringify(locale)}]`

bin/build-scripts/ftl-to-js.js
```diff
@@ -0,0 +1,28 @@
...
+      }
```
ianb: This would be a good place to put in a call to the effect of "your l10n messages have arrived".  E.g., `if (typeof window !== "undefined" && window.notifyL10nLoaded) {window.notifyL10nLoaded(locale);}`

server/src/locale-messages.js
```diff
@@ -0,0 +1,29 @@
...
+        }
+        if (wait > 0) {
+          wait -= 100;
+          setTimeout(checkForMessages, 100);
+          return;
+        }
+        reject(locale);
+      };
+      checkForMessages();
+    });
```
ianb: You can use the callback referenced earlier (`notifyL10nLoaded`) to avoid the polling here. Though it's still good to account for either ordering (i.e., the locales load first, then this file; or this file loads first then the locales).

bin/build-scripts/ftl-to-js.js
chenba: I don't understand this.  Since `locale` is an element from an array returned from a `string.split()` call, it already should be a string.  What would calling `JSON.stringify` get us?

chenba: Oh, nevermind. Just realized it. 👍

chenba:
With these changes, I believe we can stop copying the `locales` directory (https://github.com/mozilla-services/screenshots/blob/master/Dockerfile#L4).

chenba:
This is ready for re-review.  Adding a param to `l10n.init` for testing didn't feel too bad.  But I can be talked out of it.

Issue #3401, opened 2017-08-24
l10n error when accessing My Shots with scratch profile


ianb:
To reproduce:

* `./bin/run-addon --scratch`
* Go to http://localhost:10080/shots
* View:

```
General error:
TypeError: Cannot read property 'map' of undefined

TypeError: Cannot read property 'map' of undefined
    at Object.exports.createModel (/Users/ianbicking/src/pageshot/build/server/pages/shotindex/model.js:32:18)
    at Object.exports.render (/Users/ianbicking/src/pageshot/build/server/reactrender.js:9:31)
    at _render (/Users/ianbicking/src/pageshot/build/server/pages/shotindex/server.js:30:17)
    at /Users/ianbicking/src/pageshot/build/server/pages/shotindex/server.js:13:5
    at Layer.handle [as handle_request] (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/route.js:137:13)
    at csrf (/Users/ianbicking/src/pageshot/node_modules/csurf/index.js:117:5)
    at Layer.handle [as handle_request] (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/layer.js:95:5)
    at /Users/ianbicking/src/pageshot/node_modules/express/lib/router/index.js:281:22
    at Function.process_params (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/index.js:335:12)
    at next (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/index.js:275:10)
    at expressInit (/Users/ianbicking/src/pageshot/node_modules/express/lib/middleware/init.js:40:5)
    at Layer.handle [as handle_request] (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/Users/ianbicking/src/pageshot/node_modules/express/lib/router/index.js:317:13)
    at /Users/ianbicking/src/pageshot/node_modules/express/lib/router/index.js:284:7
[...]

PR #3426, opened 2017-08-29
Ensure that `shots` is an array before calling `shots.map()` (#3401)

6a68:


server/src/pages/shotindex/model.js
```diff
@@ -28,7 +28,7 @@ exports.createModel = function(req) {
       image.url = createProxyUrl(req, image.url);
     }
   }
-  if (shots !== null && shots !== undefined) {
+  if (shots && shots.length) {
```
pdehaan: Would we want to use [`Array.isArray()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)? Although, I can't imagine that `shots` would ever be a string and we'd get odd behavior w/ it checking a string length.

6a68: ah, got merged before I had a chance to fix it up ^_^

PR #3428, opened 2017-08-29
3299 header tweaks take 2

g-k:
Rebased and slightly updated version of https://github.com/mozilla-services/screenshots/pull/3353

Rolling back the cookie key to the default of `_csrf` might have fixed https://github.com/mozilla-services/screenshots/pull/3353#issuecomment-325711537 for now.

Quickly ran some simple functional tests: reload the webextension, take a shot, edit it, change expiration, change title, delete it, delete account.

Didn't verify that the secure attr was set but hopefully that's OK to test in dev or stage.

### DIFF
--- a/server/src/middleware/csrf.js
+++ b/server/src/middleware/csrf.js
@@ -19,7 +19,7 @@ exports.csrf = function(req, res, next) {
     let exc = new Error("Duplicate CSRF cookies");
     exc.headerValue = rawCookies;
     captureRavenException(exc);
-    simpleResponse(res, "Bad request", 400);
+    return simpleResponse(res, "Bad request", 400);
   }
   req.cookies._csrf = req.cookies.get("_csrf"); // csurf expects a property
   next();
--- a/server/src/middleware/csrf.js
+++ b/server/src/middleware/csrf.js
@@ -3,20 +3,24 @@ const mozlog = require("../logging").mozlog("csrf-middleware");
 const { captureRavenException } = require("../ravenclient");
 const { simpleResponse } = require("../responses");
 
-exports.csrfProtection = csrf({cookie: {httpOnly: true, secure: true, sameSite: 'lax', key: '__Host-csrf'}});
+const config = require("../config").getProperties();
+
+const useSecureCsrfCookie = (config.expectProtocol && /^https$/.test(config.expectProtocol));
+
+exports.csrfProtection = csrf({cookie: {httpOnly: true, secure: useSecureCsrfCookie}});
 
 exports.csrf = function(req, res, next) {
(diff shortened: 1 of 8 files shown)


server/src/middleware/csrf.js
```diff
@@ -0,0 +1,34 @@
...
+  let rawCookies = req.get("cookie") || "";
+  let pairs = rawCookies.split(";");
+  let csrfTokens = pairs.filter(item => item.match(/_csrf=/));
+  if (csrfTokens.length > 1) {
+    let exc = new Error("Duplicate CSRF cookies");
+    exc.headerValue = rawCookies;
+    captureRavenException(exc);
+    return simpleResponse(res, "Bad request", 400);
+  }
+  req.cookies._csrf = req.cookies.get("_csrf"); // csurf expects a property
```
6a68: I thought we wanted to change the cookie name from '_csrf' to '__Host-csrf'?

g-k: We do.  I think changing prefixes in a separate PR will be less likely to break things and didn't realize the `__Host-` requires the secure attribute (so we'd need to something like change prefixes depending on whether we're setting secure or not).

https://tools.ietf.org/html/draft-ietf-httpbis-cookie-prefixes-00#section-3.2

6a68: I'm guessing local development is the only reason to allow non-https server URLs. Maybe this issue is a good excuse to force us to get https working locally

ianb: For a while we did require https locally, it was pretty complicated to setup. I felt good when I was able to remove it.

6a68: Yeah, I just noticed #2265. Webpack has a pretty nice https out of the box experience, but it can be pretty painful to setup manually, accept the certs in the temporary profile, etc.

server/src/middleware/csrf.js
```diff
@@ -0,0 +1,34 @@
...
+  next();
+};
+
+exports.csrfErrorResponse = function(err, req, res) {
+  assert(err.code === "EBADCSRFTOKEN", "Returning csrf response for non-csrf error code.");
+  mozlog.info("bad-csrf", {id: req.ip, url: req.url});
```
ianb: Whoops, this should be `ip` instead of `id` (I think I made the same typo in some other code, maybe it got copied?)

server/src/middleware/csrf.js
ianb: Followup in #3455

server/src/middleware/csrf.js
g-k: Oh yeah, good catch!

PR #3432, opened 2017-08-29
Fix #2254, build production versions of bundles

### LINKED ISSUE #2254: Warning: It looks like you're using a minified copy of the development build of React
> _**Warning:** It looks like you're using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See https://fb.me/react-minification for more details._

view_source_in_debugger_ _https___pageshot_net_static_js_shot-bundle_js_1_12106_and_

ianb:
This always builds bundles as a production build. envify and uglifyify should also help improve overall bundle sizes

### DIFF
--- a/bin/build-scripts/bundle_dependencies
+++ b/bin/build-scripts/bundle_dependencies
@@ -34,7 +34,9 @@ build() {
   echo -n "Building $(basename $dest): deps... "
   browserify --list -e "$@" | sed "s!$(pwd)/!!g" | grep -v build-time > "$(depfile)"
   echo -n "bundle... "
-  browserify -o "$dest" \
+  NODE_ENV=production browserify -o "$dest" \
+    -g [ envify --NODE_ENV production ] \
+    -g uglifyify \
     -e "$@"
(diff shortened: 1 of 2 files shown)


bin/build-scripts/bundle_dependencies
```diff
@@ -34,7 +34,9 @@ build() {
   echo -n "Building $(basename $dest): deps... "
   browserify --list -e "$@" | sed "s!$(pwd)/!!g" | grep -v build-time > "$(depfile)"
   echo -n "bundle... "
-  browserify -o "$dest" \
+  NODE_ENV=production browserify -o "$dest" \
+    -g [ envify --NODE_ENV production ] \
+    -g uglifyify \
```
pdehaan: What does this mean in relation to the `$NO_UGLIFY` block below? Even if we specify NO_UGLIFY, the bundle would still be minified, no?

6a68: Yup, see #3433 for the rest of the story

PR #3434, opened 2017-08-29
Fix #3197, swapped next and prev buttons for RTL 

### LINKED ISSUE #3197: The intro slides prev and next buttons are swapped in RTL
The introduction slides (I hope I’m using the correct term) move from right to left correctly in RTL locales, but the prev and next buttons are swapped i.e. not RTL'd (prev should be on the right and next should be on the left).

ankita240796:


ianb:
A nice RTL followup, thank you ;)

PR #3440, opened 2017-08-30
Bug 1389705

g-k:
validate clip image urls and don't render shotindex for invalid urls

In addition to the tests added here ran the functional test:

* select and save a clip
* in DB edit clip to invalid image url e.g. `UPDATE data SET value = REPLACE(value, '.png"', '.png); background-color: red;"') WHERE id = 'FxBLw0JuOJIsvtpz/localhost';`
* loading the shot index for that page returns a 500 (assertion error) and loading myshots skips the image


g-k:
I don't think this error from CI is related since running run-server again fixes the issue for me locally:

```
Building shot-bundle.js: deps... bundle... minimizing... Parse error at build/server/static/js/shot-bundle.js:33522,27
function assert(condition, ...args) {
                           ^
```

r? @ianb

g-k:
Ah, so browserify is outputting an es6 spread/splat `function assert(t,...e)` that uglify chokes on since it's es5 only.

ianb:
Ah, I see – you were importing the un-built shot.js, instead of the version that had already gone through babel

Issue #3407, opened 2017-08-25
Unstyled content flashes on homepage


johngruen:
Will try to grab a recording, but basically css gets passed in after content :(

6a68:
Here's a screencast of the issue:

![fouc-3407](https://user-images.githubusercontent.com/96396/29933417-6034add8-8e2c-11e7-9917-01f75ca32046.gif)

I bet moving the CSS above the ~700kb of JS would help

6a68:
Might need to move the homepage CSS above the GA and Raven snippets for this approach to work. It looks like those scripts (small yellow blocks) get loaded before the first paint (green block), but the big homepage JS (long yellow block) doesn't:

<img width="1095" alt="screen shot 2017-08-31 at 9 29 58 am" src="https://user-images.githubusercontent.com/96396/29934340-981dba34-8e2f-11e7-8156-a2c392860164.png">

The first attempt PR moves the homepage CSS just above the homepage JS, so it might (hopefully will) get loaded before the first paint. I'll try on a low-spec machine once this hits stage, too. 

Other approaches might include delaying rendering until the CSS loads, or inlining some/all of the homepage CSS. Hopefully this tweak will be good enough.

6a68:
Looks like this is actually a Nightly bug. See the discussion on dev-platform: https://groups.google.com/d/topic/mozilla.dev.platform/uN4idW92CCQ/discussion

Indeed, I cannot reproduce this bug on Chrome.

Probably still good to load CSS earlier, if only to improve perception among Nightly users.

PR #3447, opened 2017-08-31
Move homepage CSS above JS to avoid FOUC

6a68:
See discussion in #3407. Hopefully the simple fix will be good enough


ianb:
I think in discussion @6a68 said that this didn't really help... but it also can't hurt.

Issue #3411, opened 2017-08-27
Make the screenshots web page highlight the Screenshots page action button


Fischer-L:
The corresponding bug on Bugzilla: Bug 1393668

Photon Onboarding tour have one Screenshots tour. That tour will lead user to the screenshots web page[1] so we would like to have the web page can highlight the Screenshots page action button to introduce users where they could find this screenshots feature.

Please see Bug 1393668 comment 9 [2] for how to highlight the Screenshots page action button in the web page.

One thing we have to consider is what is the appropriate timing and flow to do the highlight so that users would know what we are delivering and not being confused, as Bug 1393668 comment 1, 3, 4 and 6 [3][4][5][6].

[1] https://screenshots.firefox.com 
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1393668#c9
[3] https://bugzilla.mozilla.org/show_bug.cgi?id=1393668#c1
[4] https://bugzilla.mozilla.org/show_bug.cgi?id=1393668#c3
[5] https://bugzilla.mozilla.org/show_bug.cgi?id=1393668#c4
[6] https://bugzilla.mozilla.org/show_bug.cgi?id=1393668#c6

Fischer-L:
The bug 1371538[1] would make the Photon Onboarding Screenshots tour open https://screenshots.firefox.com/#hello.
Why open to the screenshots#hello page is because
- before the Screenshots page action highlight gets implemented, at least we could provide more useful tutorial to users in stead of nothing but one web page.
- per the comment[2], it seems an appropriate place to do the highlight job. 

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1371538
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1393668#c4

johngruen:
@Fischer-L The work here is as follows:

We need to implement a UI tour trigger from https://sceenshots.firefox.com/#tour, and the onboarding team needs to send users to that URL from the primary onboarding tour. Our onboarding will then occur independently based on whether a user has previously been through our onboarding.

Is this something you can implement for us?

ianb:
[...]

PR #3449, opened 2017-08-31
Add Mozilla.UITour call to homepage, for #3411

ianb:
We should confirm with @Fischer-L that this is the right way to do this


Fischer-L:
@ianb this looks good. If wanna try UITour on localhost or screenshots.firefox.com, we could do [1]. However, since showHighlight on "screenshots" is not yet ready, we could try `showHighlight("addons")` first.

[1] http://bedrock.readthedocs.io/en/latest/uitour.html#local-development

PR #3451, opened 2017-08-31
Fix icon appearance for Photon page action.

0c0w3:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1395284

Right now, the icon is too dark, so it doesn't match the
appearance of the other Photon page actions.

The problem is that the URI passed as the action's iconURL is a
file:// URI.  The Photon theme uses context-fill and
context-fill-opacity in SVG in order to style SVG icons
correctly, and SVG context painting is not supported for file
URIs: https://dxr.mozilla.org/mozilla-central/rev/db7f19e26e571ae1dd309f5d2f387b06ba670c30/layout/svg/SVGContextPaint.cpp#39

bootstrap.js should pass a moz-extension:// URI instead, which
context painting does support, and which is what the WebExtension
browser action toolbar button uses.

Additionally, the icon SVG used by the Photon page action needs
to be updated with fill-opacity="context-fill-opacity".


addon/webextension/icons/icon-16-v2.svg
```diff
@@ -1 +1 @@
...
+<svg width="32" height="32" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg" fill="context-fill" fill-opacity="context-fill-opacity"><title>icon-16</title><path d="M8 2a4 4 0 0 0-4 4h4V2zm12 0h-4v4h4V2zm8 0v4h4a4 4 0 0 0-4-4zM14 2h-4v4h4V2zm12 0h-4v4h4V2zm2 10h4V8h-4v4zm0 12a4 4 0 0 0 4-4h-4v4zm0-6h4v-4h-4v4zm-.882-4.334a4 4 0 0 0-5.57-.984l-7.67 5.662-3.936-2.76c.031-.193.05-.388.058-.584a4.976 4.976 0 0 0-2-3.978V8H4v2.1a5 5 0 1 0 3.916 8.948l2.484 1.738-2.8 1.964a4.988 4.988 0 1 0 2.3 3.266l17.218-12.35zM5 17.5a2.5 2.5 0 1 1 0-5 2.5 2.5 0 0 1 0 5zm0 12a2.5 2.5 0 1 1 0-5 2.5 2.5 0 0 1 0 5zm10.8-4.858l6.034 4.6a4 4 0 0 0 5.57-.984L19.28 22.2l-3.48 2.442z"/></svg>
```
msujaws: Might as well remove the `title` tag from both this, icon-32-v2.svg, and icon-starred-32-v2.svg while we're at it.

addon/webextension/icons/icon-starred-32-v2.svg
```diff
@@ -1 +1 @@
...
+<svg data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32" width="32" height="32"><title>icon-starred-32-v2</title><path fill="context-fill" fill-opacity="context-fill-opacity" d="M8 2a4 4 0 0 0-4 4h4zm6 0h-4v4h4zm14 22a4 4 0 0 0 4-4h-4zm-12.2.64l6 4.6a4 4 0 0 0 5.57-1l-8.09-6.04zM21.86 17.44L9.9 26a5 5 0 1 1-2.3-3.27l2.8-2L7.92 19A5 5 0 1 1 4 10.1V8h4v3a5 5 0 0 1 2 4 5 5 0 0 1-.06.58l3.94 2.76 4.46-3.29a9 9 0 0 0 3.52 2.39zM5 17.5A2.5 2.5 0 1 0 2.5 15 2.5 2.5 0 0 0 5 17.5zm0 12A2.5 2.5 0 1 0 2.5 27 2.5 2.5 0 0 0 5 29.5zM16.51 6H16V2h3.34a9 9 0 0 0-2.83 4zM28 17.49V18h4v-3.34a9 9 0 0 1-4 2.83z"/><circle fill="#00feff" fill-opacity="context-fill-opacity" cx="25" cy="9" r="7"/><path fill="#005a71" fill-opacity="context-fill-opacity" d="M25 4a.89.89 0 0 1 .89.69l.4 1.65.07.31.31-.09 1.64-.48a.91.91 0 0 1 1.11.54.87.87 0 0 1-.22 1L28 8.78l-.27.22.23.22 1.24 1.17a.87.87 0 0 1 .22 1 .91.91 0 0 1-1.11.54l-1.64-.48-.31-.09-.07.31-.4 1.65a.92.92 0 0 1-1.78 0l-.4-1.65-.07-.31-.31.09-1.64.48a.91.91 0 0 1-1.11-.54.87.87 0 0 1 .22-1L22 9.22l.27-.22-.27-.22-1.21-1.17a.87.87 0 0 1-.22-1 .91.91 0 0 1 1.11-.54l1.64.48.31.09.07-.31.4-1.65A.89.89 0 0 1 25 4"/></svg>
```
msujaws: The data-name attribute on `svg` can be removed too.

0c0w3:
Thanks Jared, but maybe there's a reason they need those things? idk

msujaws:
No reason for those, a lot of editors and tools insert those. We have removed them with no issues in other SVGs provided by UX.

ianb:
./node_modules/.bin/svgo should clean those up – we run all our server svgs through that, but it's not part of our add-on release process.

msujaws:
Cool, is there an issue open to run svgo through the add-on release process?

ianb:
No, we rewrite all of the files in the server into `build/`, which is where svgo fits in, but the ergonomics on the add-on side are a bit better if we run it mostly in-place. I guess a lint-like process would work for that case, but svgo isn't fast and isn't a linter so I think it would be annoying in practice.

msujaws:
Okay, so it seems we might as well clean this up while we're here then.

ianb:
Let me just land this, and I'll run svgo on that whole directory as a separate PR

0c0w3:
Thanks Ian!

PR #3457, opened 2017-09-01
Run all add-on svg files through svgo

ianb:
Followup to #3451 (and also optimizes a bunch of other svgs)


ianb:
r=me, since I don't think any of us are going to review the validity of the XML changes. Visually it looks good to me.

PR #3458, opened 2017-09-01
fix wrong path reference in server.js and proxy-url.js

relud:


6a68:
Oh, I see, related to 428f127

relud:
@6a68 the paths were not already correct, because the docker container would not even start without this change

ianb:
There must have been some build problem? This appears to change the imports to use the unbuilt imports

Issue #3443, opened 2017-08-31
Update slide image to match photon


johngruen:
For 57 we need to update the asset seen in this slide so that the browser chrome looks like firefox 57

<img width="843" alt="screen shot 2017-08-31 at 9 58 17 am" src="https://user-images.githubusercontent.com/3323249/29927063-3526b5b2-8e33-11e7-88fa-bc2748133195.png">

johngruen:
@fangshih

fangshih:
Updated the browser chrome to firefox 57 look.
![img-onboarding-4 2x](https://user-images.githubusercontent.com/4862573/29953826-dd098368-8f05-11e7-86cb-7b2fb4f38cd9.png)

https://github.com/mozilla/testpilot-assets/blob/master/Firefox_Screenshots/00_Assets/Onboarding/%402x/img-onboarding-4%402x.png

PR #3471, opened 2017-09-06
Slide changes

### LINKED ISSUE #3442: New 57-friendly slide to tour
Since Screenshots is now a page action, we should remind people where it lives in the tour.
Let's add a slide between the first and third slides in the tour:

**Image:** https://github.com/mozilla/testpilot-assets/blob/master/Firefox_Screenshots/00_Assets/Onboarding/%402x/img-onboarding-5%402x.png

**Title:** A new way to save
**Sentence:** Expand the page actions menu in the address bar any time you want to take a screenshot.

NB this image needs to be updated slightly to reflect the photonized icon, but we might want to land this quickly to get strings in l10n.

ianb:
Fixes #3442 and #3443 

The new slide:

image

Probably we should stop numbering the slides, so renames like this aren't necessary. But... instead I renamed everything as I inserted a slide between slide 1 and 2. Anyhow, 2550449 does only the renaming.

/cc @flodolo 

FYI, renames done with:

```
cd locales
for file in $(find . -name 'webextension.properties') ; do echo "Change file $file" ; sed -i '' -e 's/tourHeaderTwo =/tourHeaderThree =/' $file ; done
```

And so on for each string.


flodolo:
> Probably we should stop numbering the slides, so renames like this aren't necessary.
+100 on this.

Can we start doing it now? This is going to mess up Translation Memory badly, e.g.

tourHeaderTwo = Capture Just What You Want 

becomes

tourHeaderTwo = A new way to save

Pontoon will import the translation, but also keep the old one around in its history, and so on for six strings.

Personally I would suggest to add the new string with an identifier that doesn't include the number of the slide (e.g. `tourHeaderNewSave`), and keep the existing strings as they are. Who cares if there is a slide between 1 and 2 when it comes to referencing strings?

We should really stop editing localization files like this.

ianb:
@flodolo : does this look better? Lots of string id changes still, but at least these names should stick

PR #3478, opened 2017-09-08
Fix multipart upload

ianb:
This was just the worst. For reasons I don't understand the browser just wasn't sending FormData correctly. It truncated it at 1210 bytes for me every time. I switched to constructing the form submission manually.


ianb:
Note a followup in #3484

PR #3486, opened 2017-09-12
Fix #3468, remove browserAction

### LINKED ISSUE #3468: Remove browserAction
We've switched to a pageAction for Firefox 57 / Photon.  But we kept browserAction support so we could support older Firefox version. After discussion we realized we won't be deploying the current master to any versions of Firefox older than 57, so we can just remove the browserAction entirely.

This should remove the browserAction from manifest.json, and related code from bootstrap.js. We should probably get a review on this change from Mossop or kmag (if #3463 is done we could experiment with it for this).

ianb:
This changes the add-on to require the Photon page action, with no fallback to a browserAction
- Removes test for Photon (assumes it is present)
- Removes bootstrap.js code that deletes the browserAction button
- Removes webextension code references to browserAction
- Removes Photon conditionals (i.e., assume it's always Photon)
- Make 57a1 the minimum version for the webextension/install.rdf

### DIFF
--- a/addon/webextension/background/startBackground.js
+++ b/addon/webextension/background/startBackground.js
@@ -1,6 +1,5 @@
 /* globals browser, main, communication */
 /* This file handles:
-     browser.browserAction.onClicked
      clicks on the Photon page action
      browser.contextMenus.onClicked
      browser.runtime.onMessage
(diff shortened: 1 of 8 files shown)


ianb:
@gijsk or @0c0w3 : could you review this? (Especially bootstrap.js changes, but also @6a68 is away this week)

addon/webextension/background/startBackground.js
```diff
@@ -60,16 +51,12 @@ this.startBackground = (function() {
...
-        browser.browserAction.setIcon({path});
-      } else {
-        iconPath = path;
-        if (photonPageActionPort) {
-          photonPageActionPort.postMessage({
-            type: "setProperties",
-            iconPath
-          });
-        }
+      iconPath = path;
```
0c0w3: Nit: I think you can remove `path` now and just set `iconPath` to the path string directly.

0c0w3:
Yeah, I think these can be removed/changed too?

https://github.com/mozilla-services/screenshots/blob/master/test/addon/browser_screenshots_ui_check.js#L19

https://github.com/mozilla-services/screenshots/blob/master/test/addon/head.js#L18

https://github.com/mozilla-services/screenshots/blob/master/test/addon/head.js#L50

PR #3513, opened 2017-09-19
Upload jpeg

ianb:
- Allows JPEGs on the server, both to pass content checks, and to make use of stored content
-types (instead of assuming image/png). 
- Puts an clip.image.type into shot objects 
- Uses .jpg for filenames when appropriate 
- Adds a new buildSetting for controlling the cutoff when we use JPEG 
- If a PNG image is too large, tries to make a JPEG and substitutes if the JPEG is actually smaller 
- Refactor some data:-URL and blob conversion functions into their own module


6a68:
@ianb looks like a couple minor issues with the linter. I'll file a followup to fix em

PR #3521, opened 2017-09-20
Bump max image dimensions to canvas limit

6a68:
It turns out that canvas.toDataURL() has a limit of 32767 pixels in any direction (see [bug 1282074](https://bugzil.la/1282074) for details).

Bumping the max height/width to those dimensions, I created some test images from facebook, imgur, and giphy. Imgur came in biggest, around 12 MB. Seems like we can safely bump up the pixel limit without hitting the server max of 25 MB per image.

I tried bumping quality up to 1.0 but couldn't see any significant quality improvement. Smallish text on Facebook is reasonably legible in either case. I'll attach a couple example images to this PR.

### DIFF
--- a/addon/webextension/selector/uicontrol.js
+++ b/addon/webextension/selector/uicontrol.js
@@ -45,8 +45,9 @@ this.uicontrol = (function() {
 
   const { watchFunction, watchPromise } = catcher;
 
-  const MAX_PAGE_HEIGHT = 5000;
-  const MAX_PAGE_WIDTH = 5000;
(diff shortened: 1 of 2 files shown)


6a68:
Here's a biggish Facebook image. Text seems readable.

screenshot-2017-9-20 puerto rico - facebook search

ianb:
I took a large shot of imgur.com, and used the dev server. It took 8 seconds to get from hitting Full Page to actually seeing the full page preview. Then 24 seconds to upload. The upload wasn't actually too bad, though I am concerned with the delay simply presenting the full page preview. There's nothing there indicating it is doing work.

PR #3550, opened 2017-09-26
Fix #2129, notify user when full page is cut off

### LINKED ISSUE #2129: Notify user when Full Page is cut off
We have a 5000px limit to images, even Full Page.  People hit that and then are confused why the images are limited.  We should tell them.

ianb:
Adds a new captureType, fullPageTruncated

Looks like this:

image

### DIFF
--- a/addon/webextension/selector/uicontrol.js
+++ b/addon/webextension/selector/uicontrol.js
@@ -1,5 +1,5 @@
 /* globals log, catcher, util, ui, slides */
-/* globals shooter, callBackground, selectorLoader, assertIsTrusted */
+/* globals shooter, callBackground, selectorLoader, assertIsTrusted, buildSettings */
 
 "use strict";
 
@@ -45,8 +45,8 @@ this.uicontrol = (function() {
 
   const { watchFunction, watchPromise } = catcher;
 
-  const MAX_PAGE_HEIGHT = 5000;
-  const MAX_PAGE_WIDTH = 5000;
+  const MAX_PAGE_HEIGHT = buildSettings.maxImageHeight;
+  const MAX_PAGE_WIDTH = buildSettings.maxImageWidth;
   // An autoselection smaller than these will be ignored entirely:
   const MIN_DETECT_ABSOLUTE_HEIGHT = 10;
   const MIN_DETECT_ABSOLUTE_WIDTH = 30;
@@ -160,22 +160,28 @@ this.uicontrol = (function() {
     },
     onClickFullPage: () => {
       sendEvent("capture-full-page", "selection-button");
+      captureType = "fullPage";
       let width = Math.max(
         document.body.clientWidth,
         document.documentElement.clientWidth,
         document.body.scrollWidth,
         document.documentElement.scrollWidth);
+      if (width > MAX_PAGE_WIDTH) {
+        captureType = "fullPageTruncated";
+      }
       width = Math.min(width, MAX_PAGE_WIDTH);
       let height = Math.max(
         document.body.clientHeight,
         document.documentElement.clientHeight,
         document.body.scrollHeight,
         document.documentElement.scrollHeight);
+      if (height > MAX_PAGE_HEIGHT) {
+        captureType = "fullPageTruncated";
+      }
       height = Math.min(height, MAX_PAGE_HEIGHT);
       selectedPos = new Selection(
         0, 0,
         width, height);
-      captureType = 'fullPage';
       setState("previewing");
     },
     onSavePreview: () => {
@@ -368,7 +374,7 @@ this.uicontrol = (function() {
     start() {
       dataUrl = shooter.screenshotPage(selectedPos, captureType);
       ui.iframe.usePreview();
-      ui.Preview.display(dataUrl);
+      ui.Preview.display(dataUrl, captureType == "fullPageTruncated");
     }
   };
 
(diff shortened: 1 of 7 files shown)


ianb:
New style:

image

ianb:
Or with silly scales:

image

ianb:
new font size:

image

static/css/inline-selection.scss
```diff
@@ -402,6 +402,20 @@ $very-light-grey: #ededed;
   width: 400px;
 }
 
+#imageCroppedWarning {
+  background: rgba(0,0,0,.8);
```
6a68: Could you specify the opacity separately? Seems to make Windows High Contrast mode behave in a more reasonable way

shared/shot.js
```diff
@@ -707,7 +707,7 @@ class _Clip {
     }
     assert(checkObject(image, ["url"], ["dimensions", "text", "location", "captureType", "type"]), "Bad attrs for Clip Image:", Object.keys(image));
     assert(isValidClipImageUrl(image.url), "Bad Clip image URL:", image.url);
-    assert(image.captureType == "madeSelection" || image.captureType == "selection" || image.captureType == "visible" || image.captureType == "auto" || image.captureType == "fullPage" || !image.captureType, "Bad image.captureType:", image.captureType);
+    assert(image.captureType == "madeSelection" || image.captureType == "selection" || image.captureType == "visible" || image.captureType == "auto" || image.captureType == "fullPage" || image.captureType == "fullPageTruncated" || !image.captureType, "Bad image.captureType:", image.captureType);
```
6a68: this line is looking awfully long

locales/en-US/webextension.properties
```diff
@@ -16,6 +16,8 @@ notificationLinkCopiedTitle = Link Copied
 # modifier key used in the shortcut (do not translate it): for example, Ctrl-V
 # on Windows systems.
 notificationLinkCopiedDetails = The link to your shot has been copied to the clipboard. Press {meta_key}-V to paste.
+# This is displayed as a warning when a Full Page save will be cropped
```
6a68: Could also mention that {pixels} is a placeholder for a plural whole number, like 5000.

addon/webextension/selector/uicontrol.js
```diff
@@ -160,22 +160,28 @@ this.uicontrol = (function() {
     },
     onClickFullPage: () => {
       sendEvent("capture-full-page", "selection-button");
+      captureType = 'fullPage';
```
6a68: y u no double quotes in this file?

static/css/inline-selection.scss
```diff
@@ -402,6 +402,20 @@ $very-light-grey: #ededed;
...
+  bottom: 0;
+  color: #fff;
+  font-family: -apple-system, BlinkMacSystemFont, "segoe ui", "helvetica neue", helvetica, ubuntu, roboto, noto, arial, sans-serif;
+  font-size: 12px;
+  padding: 10px;
+  position: absolute;
+  text-align: center;
+  width: 100%;
+  z-index: 2;
+  position: absolute;
```
6a68: repeated CSS property

locales/en-US/webextension.properties
```diff
@@ -16,6 +16,8 @@ notificationLinkCopiedTitle = Link Copied
...
+imageCroppedWarning = This image has been cropped to {pixels} pixels.
```
6a68: Should the string be, "This image _will be_ cropped to {pixels} pixels" ?

flodolo: If you're using "pixel", you should use a proper plural form. The easy way out is to use px.
http://projectfluent.io/fluent/guide/selectors.html

Also CC @stasm to make sure we don't break anything if you go for proper plural.

6a68: @flodolo In this case the singular will never be needed, so I'm not sure I see the need for the extra selector syntax

static/css/inline-selection.scss
ianb: I don't think you can make the background opacity separate, only the entire element opacity...?

6a68:
@ianb Also, this might be a good time to bump the limit to 10k pixels (#3538)

locales/en-US/webextension.properties
ianb: @johngruen said "px", but I thought that encouraged a term that a normal human wouldn't know so I switched to pixels.

ianb:
This is for uplift to 57, so I think in a meeting we decided not to mess with anything unnecessarily (like the pixel limit)

locales/en-US/webextension.properties
flodolo: Once again, stop using your English brain to think ;-)

one vs many is not a plural form, it's something that works in a bunch of languages
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Use_proper_plural_forms

justfortherec:
> This is for uplift to 57

This sounds like this change is supposed to be in Firefox 57 but I cannot see the message.

Before filing a bug because of truncated full page screenshots I found this change. I cannot see the message in Firefox 57 on Ubuntu. Should I file a bug or will this be available in a future release?

chenba:
@justfortherec It will be in a future release.

PR #3554, opened 2017-09-27
Remove system pref

6a68:


6a68:
I'm nervous about ignoring this test error. Although I can't find any problems when manually testing this branch, reverting the pref change commit causes the "find the button" test to pass. Might be a race condition / timing issue.

6a68:
Finally green. Merging

PR #3563, opened 2017-09-28
Update packages for nsp/regex reasons

ianb:


ianb:
Let's get 'er done (and tests running)

PR #3564, opened 2017-09-28
[Do not merge] Fix #2129, notify user when full page is cut off

### LINKED ISSUE #2129: Notify user when Full Page is cut off

ianb:
Tracking for export. See [Bug 1404078](https://bugzilla.mozilla.org/show_bug.cgi?id=1404078)

---

Adds a new captureType, fullPageTruncated


flodolo:
Wasn't this string exposed already?

ianb:
This was exported to Firefox 58, but not uplifted to 57.

Issue #3083, opened 2017-07-03
Sanitize download filenames using Unicode categories


ianb:
I'm struggling to figure out how to sanitize download filenames properly. Right now we filter *out* certain characters, which seems unsafe. But if we only allow `[a-zA-Z0-9]` then we end up with nothing for, say, Cyrillic languages.

Unicode has a notion of [categories](https://en.wikipedia.org/wiki/Unicode_character_property#General_Category) and we would want the letter and number categories.  [This library](https://github.com/paulmillr/unicode-categories) has a list of appropriate categories, and maybe we'd just want to copy [those regexes](https://github.com/paulmillr/unicode-categories/blob/master/index.js#L5) into our code).  We'd want the union of Ll, Lt, Lm, Lo, and Nd.

PR #3569, opened 2017-09-29
Fix: remove control characters from filenames

rmisev:
This patch removes all control characters (0x00...0x1F) from filenames, not only "\n\r\t".

They are not allowed in Microsoft Windows file names, see:
https://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx#naming_conventions

And possibly in other systems:
https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words


ianb:
Thanks! (If you are interested in this, there's a ticket to turn this into a character whitelist: #3083)

PR #3584, opened 2017-10-02
[Do not merge] Fix #3549, remove code that checks for system-disabled pref

### LINKED ISSUE #3549: Drop the code that uses `extensions.screenshots.system-disabled` part 2
This was originally #3283 but that was closed with just removing the default pref.  We should remove the code in the add-on that looks at system-disabled altogether.

6a68:
Tracking for export. See [bug 1405166](https://bugzilla.mozilla.org/show_bug.cgi?id=1405166)


6a68:
This patch has now been uplifted to 57, see https://bugzil.la/1405166#c17. Merging

PR #3601, opened 2017-10-06
Try to fix #3581, make Screenshots work with third party cookies disabled

### LINKED ISSUE #3581: Make Screenshots work with third party cookies disabled
I believe what's causing Screenshots not to work with third-party cookies disabled is that we make all our authentication requests from the background page, and rely on those requests setting the authentication cookie.

The design constraint that led to this is that I didn't want to send the authentication information to a content page. That seemed like it would make it too easy to extract that information with any compromise to the site.

I think by sending the value of the `X-Screenshots-Auth` header to `sitehelper.js` we could do this Well Enough. That header value is basically the value of the cookie (though the cookie is httpOnly, so typically content never gets to see it). It's a value that can't be used quite as easily to authenticate other clients. Then `sitehelper.js` could create an XMLHttpRequest/fetch with that header, and get the cookie set – but it wouldn't appear to be a third-party cookie.

So the basic flow would be:

1. user goes to a page with no authentication (because they've blocked the normal authentication)
2. `wantsauth.js` is loaded
3. wantsauth asks `sitehelper.js` to authentication
4. sitehelper sends the value of X-Screenshots-Auth back to wantsauth
5. wantsauth sends a login request to a new server endpoint with X-Screenshots-Auth
6. the server responds with a response that sets the authentication cookie
7. wantsauth refreshes the page

@g-k @jvehent : feedback from a security perspective?

ianb:
This adds a second attempt to login to wantsauth logins, one that runs in sitehelper.js, and tries to get the cookie set on a request that appears to come from the content page itself.

Note: this now works, but doesn't provide full protection from a screenshots.firefox.com page (that is under attack) from getting the authentication header.

### DIFF
--- a/server/src/middleware/csrf.js
+++ b/server/src/middleware/csrf.js
@@ -26,7 +26,8 @@ function isCsrfExemptPath(path) {
   return isAuthPath(path)
     || path.startsWith("/data")
     || path === "/event"
-    || path === "/error";
+    || path === "/error"
+    || path === "/api/set-login-cookie";
 }
 
 function csrfHeadersValid(req) {
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -587,6 +587,19 @@ app.post("/api/login", function(req, res) {
   });
 });
 
+app.post("/api/set-login-cookie", function(req, res) {
+  if (!req.deviceId) {
+    sendRavenMessage(req, "Attempt to set login cookie without authentication");
+    simpleResponse(res, "Not logged in", 401);
+    return;
+  }
+  sendAuthInfo(req, res, {
+    deviceId: req.deviceId,
+    accountId: req.accountId,
+    userAbTests: req.abTests
+  });
+});
+
 app.put("/data/:id/:domain", upload.single('blob'), function(req, res) {
   let slowResponse = config.testing.slowResponse;
   let failSometimes = config.testing.failSometimes;
--- a/addon/webextension/background/auth.js
+++ b/addon/webextension/background/auth.js
@@ -200,12 +200,19 @@ this.auth = (function() {
 
   communication.register("getAuthInfo", (sender, ownershipCheck) => {
     return registrationInfoFetched.then(() => {
+      return exports.authHeaders();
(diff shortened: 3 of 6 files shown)


ianb:
@mixedpuppy pointed me to [Bug 1295660](https://bugzilla.mozilla.org/show_bug.cgi?id=1295660), which has a patch to expose the XMLHttpRequest, in the way that this patch attempts but fails to fetch it

server/src/middleware/csrf.js
```diff
@@ -26,7 +26,8 @@ function isCsrfExemptPath(path) {
   return isAuthPath(path)
     || path.startsWith("/data")
     || path === "/event"
-    || path === "/error";
+    || path === "/error"
+    || path === "/api/set-login-cookie";
```
g-k: If the POST `/api/set-login-cookie` sends an origin header, we should validate its protocol/scheme is `moz-extension://` in the csrf middleware.

server/src/server.js
```diff
@@ -587,6 +587,19 @@ app.post("/api/login", function(req, res) {
...
+    sendRavenMessage(req, "Attempt to set login cookie without authentication");
+    simpleResponse(res, "Not logged in", 401);
+    return;
+  }
+  sendAuthInfo(req, res, {
+    deviceId: req.deviceId,
+    accountId: req.accountId,
+    userAbTests: req.abTests
+  });
+});
```
g-k: nit: unit tests for this route or an issue to add them

server/src/middleware/csrf.js
ianb: It appears not to send any Origin header at all. I'm a bit surprised, I would have expected it to be the site origin. Or maybe it would only be set for CORS requests?

server/src/server.js
ianb: Tests added

ianb:
I created a followup to make use of [Bug 1295660](https://bugzilla.mozilla.org/show_bug.cgi?id=1295660) in #3626

Issue #3589, opened 2017-10-03
Remove Share panel from mobile shot pages


6a68:
Reported in slack by @ryanfeeley for ipad. I confirmed on iphone.

Looks like mobile safari doesn't just support `execCommand(copy)`. [This SO answer](https://stackoverflow.com/questions/32851485/make-clipboard-copy-paste-work-on-iphone-devices/43001673#43001673) might be worth investigating.

ianb:
In some previous designs we just dropped the share button on mobile, since the browser has its own share interface.

wresuolc:
Let's hide the share button on mobile views on both My Shots and the individual shot pages

ianb:
We should just display:none the share button on shot pages, using responsive media queries. Note we don't have to worry about My Shots currently, as no one on mobile has access to My Shots. (That will change with FxA support.)

chenba:
I'm going to use an existing media query for what's considered mobile.  But that won't work for iPads in landscape or iPad Pros (or similarly high-res tablets).  iPad was mentioned in the first comment (https://github.com/mozilla-services/screenshots/issues/3589#issue-262485471).

PR #3619, opened 2017-10-11
Hide share button on mobile with media query. (#3589)

chenba:


ianb:
I think if you literally say "Fix #3589" then it'll be closed when the commit is merged.

PR #3623, opened 2017-10-11
Set cache expiration for shot images. (#3591)

### LINKED ISSUE #3591: No cache expiration on shot images
This is a followup from #3202

The shot images themselves have no cache headers. Other resources have cache set properly. We should probably just use `require("./caching").setCache()` on the shot image response, though maybe with a shorter cache time since a long cache time might keep the image from being purged when expired or deleted.

chenba:
Without data, I'm taking a guess that shot access/share has a long tail, so the cache expiration is set to a day.

Fix #3591

### DIFF
--- a/server/src/caching.js
+++ b/server/src/caching.js
@@ -1,17 +1,23 @@
 const config = require("./config").getProperties();
 
+const aDayInSeconds = 24 * 60 * 60;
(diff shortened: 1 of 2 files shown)


chenba:
`s-maxage` added for shot images.

Issue #3234, opened 2017-07-28
Upgrade node-pg


ianb:
Version 7 of the `pg` library has some breaking changes, see [this upgrade guide](https://node-postgres.com/guides/upgrading)

ianb:
This issue is being closed (without being fixed) because [we are discontinuing the Screenshots server](https://blog.mozilla.org/futurereleases/2019/01/24/clarifying-the-future-of-firefox-screenshots/).

PR #3628, opened 2017-10-11
NPM Packages Update

### LINKED ISSUE #3588: Update raven-js to 3.18.1
This is a (semi-)automated bug making you aware that there is an available upgrade for an embedded third-party library. You can leave this bug open, and it will be updated if a newer version of the library becomes available. If you close it as WONTFIX, please indicate if you do not wish to receive any future bugs upon new releases of the library.

raven-js is currently at version 3.17.0 in mozilla-central, and the latest version of the library released is 3.18.1. 

I fetched the latest version of the library from https://github.com/getsentry/raven-js/.

chenba:
This patch upgrades a good number of npm packages, including some major version upgrades.  The tests passed and I didn't notice any errors when I tried it out.

`pg` (major version) and `convict` (patch version) were not upgraded because both caused `run-server` to fail.  The `react` (major) packages were not upgraded because of a peer dependency.

Fix #3588


ianb:
Yeah, we've held back pg before (and have #3234 to update it sometime)

PR #3630, opened 2017-10-12
Floor a potentially fractional width. (#3541)

### LINKED ISSUE #3541: Screenshots are blurry due to strange fractional width
First, props for a really cool tool.

I gave it a go and took [this](https://screenshots.firefox.com/hblWqiLDMTg8dJmg/screenshots.firefox.com) screenshot. On the preview page the image is given a really odd width via an inline style `height:auto;width:350.6333312988281px;max-width:100%;`

As such, it appears blurry (you'll want to open up these images below to full size to compare):

Original

The actual width is `350px` and if I modify the style to this, then it looks sharp as you'd expect:

With width changed

chenba:
Fix #3541

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -40,7 +40,7 @@ class Clip extends React.Component {
     if (!isValidClipImageUrl(clip.image.url)) {
       return null;
     }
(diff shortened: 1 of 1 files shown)


lol768:
Thanks, looks much better now 👍

PR #3647, opened 2017-10-17
Really fix selenium

ianb:


test/test.js
```diff
@@ -246,6 +252,11 @@ describe("Test Screenshots", function() {
     }).then(() => {
       return focusIframe(driver, PRESELECTION_IFRAME_ID);
     }).then(() => {
+      // This avoids a problem where the UI has been instantiated, and handlers
+      // added, but not everything is fully setup yet; by waiting we give it
+      // time to set everything up
+      return setTimeoutPromise(500);
```
6a68: Since we're here again, the right way to fix this, according to webdriver best practices, is to use an implicit or explicit wait, rather than bake a long delay into the tests. @davehunt any suggestions on how to ensure webdriver waits for iframed UI to be fully ready?

6a68:
eh, let's just land this and do the even better thing in a followup bug

test/test.js
ianb: If we want to fix this we'd have to do something a bit fancier. The elements appear (it does wait), but if you click them right away there are problems.

This display code is the issue: https://github.com/mozilla-services/screenshots/blob/2abb4de30f12c77aacf0d36dc3ed124851aa0453/addon/webextension/selector/ui.js#L414-L418

*All* of `iframeSelection.display`, `iframePreSelection.display` and `iframePreview.display` have to complete before you click on elements. They all complete in a reasonable time, but not reasonable for a computer that's polling for the interface.

To fix this properly we'd have to keep that promise, and use it to guard all the actual functions in `installHandlerOnDocument` and `standardOverlayCallbacks`. It felt a bit too complicated, though I suppose it's only 15ish lines of code when I think about it.

ianb: Followup in #3653

PR #3649, opened 2017-10-18
[do not merge] Put save/download buttons in the shoulder, not near the image

6a68:
(Code works, but I want to wait to merge until this gets UX feedback.)

@johngruen @youwenliang @ianb mind taking a look?

When looking at issues related to selection button positioning, it
occurred to me that the UI would be more consistent if we always kept
the controls positioned at the top corner of the screen. This commit
moves the selection buttons (cancel, save, download) to the top-right
corner of the overlay, and adjusts their style to match the
pre-selection shoulder buttons (save full page, save visible, my shots).

The shoulder button fixed positioning is a little janky if the page is
fast-scrolled when there's a selection. `Position: sticky` would be ideal
here, as the jank is caused by async scrolling, but `sticky` doesn't work
inside a positioned iframe, same problem as trying to use `position:
fixed` inside an absolutely positioned iframe :-\

It does seem unlikely that a user would fast-scroll a page while
adjusting the screenshot selection's dimensions, so hopefully this jank
won't be noticeable to everyday users.

Still todo: need to tidy up CSS. Also, the label text looks a little small.

-----

Screenshots of the new buttons in the creation flow:

screen shot 2017-10-17 at 5 05 29 pm
screen shot 2017-10-17 at 5 05 39 pm

-----

Close-up screenshots of the button strip and hover states:

screen shot 2017-10-17 at 5 06 05 pm
screen shot 2017-10-17 at 5 06 39 pm
screen shot 2017-10-17 at 5 06 28 pm
screen shot 2017-10-17 at 5 06 34 pm

### DIFF
--- a/addon/webextension/selector/ui.js
+++ b/addon/webextension/selector/ui.js
@@ -504,6 +504,16 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
       } else {
         this.download.style.display = "none";
       }
+      // Now that the button container has a spacer and some margins, it needs
+      // to be separately hidden when the callbacks are unset.
+      // TODO: Seems like either all callbacks are present, or none are.
+      // Would it be simpler to just display:none the whole button container,
+      // rather than individual buttons?
+      if (!callbacks) {
+        this.buttons.style.display = "none";
+      } else {
+        this.buttons.style.display = "";
+      }
       let bodyRect = getBodyRect();
       // Note, document.documentElement.scrollHeight is zero on some strange pages (such as the page created when you load an image):
       let docHeight = Math.max(document.documentElement.scrollHeight || 0, document.body.scrollHeight);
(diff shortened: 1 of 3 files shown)


addon/webextension/selector/ui.js
```diff
@@ -610,9 +606,34 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
       this.bgBottom = makeEl("div", "bghighlight");
       iframe.document().body.appendChild(this.bgBottom);
       iframe.document().body.appendChild(boxEl);
+      iframe.document().body.appendChild(this.buttons);
```
6a68: Hmm. Might be a little weird that `this.buttons` is created in `_createButtons` but assumed to exist here. maybe `_createButtons` should return `buttons`, then this function can set `this.buttons = buttons` and append it to the DOM?

locales/en-US/webextension.properties
```diff
@@ -6,6 +6,7 @@ contextMenuLabel = Take a Screenshot
 myShotsLink = My Shots
 screenshotInstructions = Drag or click on the page to select a region. Press ESC to cancel.
 saveScreenshotSelectedArea = Save
+saveScreenshotSelectedAreaToCloud = Save to Cloud
```
6a68: `make clean && make addon` wasn't enough to make this new string appear, for some odd reason

flodolo: Uhm, what's "Cloud" in this context? I feel like we should add an explanation in a comment

youwenliang:
Quick thoughts on this PR, I understand the idea for consistency but most of the time keeping the main functions at the same location is also an important part to keep consistency. In this case, the original three buttons ("full page", "visible", "my shots") is always on the top right corner, and if we swap out the functions with "cancel", "download", "save" while users are taking shots, this could confuse users.

Another thing to consider is that we want to keep the action buttons close within user's reach when taking shots. If you think about larger screens, after a user selects an area, they have to move all the way to the top right corner to do the desired actions, which is not ideal.

image

Any other thoughts? @johngruen @ianb

johngruen:
Agree with @youwenliang's concerns here. I am skeptical of UIs where buttons change their interaction patterns mid-flow. I also think we're losing a lot by moving the contextual controls away from the selection context.

We already do some math when selecting off the right hand side of the page...It seems like we should just put together a proposal for what should happen when buttons flow off the right or bottom of the page.

6a68:
Cool, thanks for the feedback. I can just tweak the existing button logic to make sure the buttons are drawn on the page. Closing this PR.

PR #3655, opened 2017-10-18
Download only mode

6a68:
@youwenliang @johngruen Mind taking a look?

Updated, now showing the download-only mode warning in the full-page preview; see screenshot below.

~~Nearly done, but just realized I've got to add the warning to the full-page preview frame. Need to do a round of todo cleanups, too.~~

Screenshots of the private browsing mode screens:

screen shot 2017-10-18 at 3 05 16 pm
screen shot 2017-10-18 at 3 05 14 pm
screen shot 2017-10-22 at 5 03 50 pm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 /Profile
 .DS_Store
+*.swp
+*.swo
 screenshots.xpi
 *.pyc
 dist
--- a/addon/webextension/background/selectorLoader.js
+++ b/addon/webextension/background/selectorLoader.js
@@ -84,7 +84,10 @@ this.selectorLoader = (function() {
   function incognitoCheck(tabId) {
     return browser.tabs.get(tabId).then(tab => {
       return browser.tabs.executeScript(tabId, {
-        code: `window.incognito = ${tab.incognito};`,
+        // Note: `window` here refers to a global accessible to content
+        // scripts, but not the scripts in the underlying page. For more
+        // details, see https://mdn.io/WebExtensions/Content_scripts#Content_script_environment
+        code: `window.downloadOnly = ${tab.incognito}`,
         runAt: "document_start"
       });
     });
--- a/addon/webextension/background/selectorLoader.js
+++ b/addon/webextension/background/selectorLoader.js
@@ -65,12 +65,12 @@ this.selectorLoader = (function() {
   let loadingTabs = new Set();
 
   exports.loadModules = function(tabId, hasSeenOnboarding) {
-    let promise;
     loadingTabs.add(tabId);
+    let promise = incognitoCheck(tabId);
     if (hasSeenOnboarding) {
-      promise = executeModules(tabId, standardScripts.concat(selectorScripts));
+      promise = promise.then(executeModules(tabId, standardScripts.concat(selectorScripts)));
     } else {
-      promise = executeModules(tabId, standardScripts.concat(onboardingScripts).concat(selectorScripts));
+      promise = promise.then(executeModules(tabId, standardScripts.concat(onboardingScripts).concat(selectorScripts)));
     }
     return promise.then((result) => {
       loadingTabs.delete(tabId);
@@ -80,6 +80,15 @@ this.selectorLoader = (function() {
       throw error;
     });
   };
+
+  function incognitoCheck(tabId) {
+    return browser.tabs.get(tabId).then(tab => {
+      return browser.tabs.executeScript(tabId, {
+        code: `window.incognito = ${tab.incognito};`,
+        runAt: "document_start"
+      });
+    });
+  }
 
   function executeModules(tabId, scripts) {
     let lastPromise = Promise.resolve(null);
--- a/addon/webextension/background/selectorLoader.js
+++ b/addon/webextension/background/selectorLoader.js
@@ -68,9 +68,13 @@ this.selectorLoader = (function() {
     loadingTabs.add(tabId);
     let promise = incognitoCheck(tabId);
     if (hasSeenOnboarding) {
-      promise = promise.then(executeModules(tabId, standardScripts.concat(selectorScripts)));
+      promise = promise.then(() => {
+        return executeModules(tabId, standardScripts.concat(selectorScripts));
+      });
     } else {
-      promise = promise.then(executeModules(tabId, standardScripts.concat(onboardingScripts).concat(selectorScripts)));
+      promise = promise.then(() => {
+        return executeModules(tabId, standardScripts.concat(onboardingScripts).concat(selectorScripts));
+      });
     }
     return promise.then((result) => {
       loadingTabs.delete(tabId);
--- a/addon/webextension/selector/ui.js
+++ b/addon/webextension/selector/ui.js
@@ -92,27 +92,17 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
   }
 
   function isDownloadOnly() {
-    return window.incognito;
+    return window.downloadOnly;
   }
 
   function renderDownloadNotice() {
     let notice = makeEl("div", "download-only");
-    notice.textContent = browser.i18n.getMessage("downloadOnlyNotice");
-    let questionMark = makeEl("span", "circle");
-    questionMark.textContent = "?";
-    notice.appendChild(questionMark);
-
-    let tooltip = makeEl("div", "download-only-details");
-    let details = makeEl("p");
-    details.textContent = browser.i18n.getMessage("downloadOnlyDetails");
-    let detailsList = makeEl("ul");
-    let detailsPrivate = makeEl("li");
-    detailsPrivate.textContent = browser.i18n.getMessage("downloadOnlyDetailsPrivate");
-
-    detailsList.appendChild(detailsPrivate);
-    details.appendChild(detailsList);
-    tooltip.appendChild(details);
-    notice.appendChild(tooltip);
+    notice.innerHTML = `
+      <span data-l10n-id="downloadOnlyNotice"></span><span class="circle">?</span>
+      <div class="download-only-details">
+        <p data-l10n-id="downloadOnlyDetails">
+          <ul><li data-l10n-id="downloadOnlyDetailsPrivate">`;
+    localizeText(notice);
     return notice;
   }
 
--- a/addon/webextension/selector/ui.js
+++ b/addon/webextension/selector/ui.js
@@ -91,6 +91,31 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
     return result;
   }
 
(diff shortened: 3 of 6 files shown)


6a68:
I'm not sure why, but the strings aren't showing up when I build this locally. That's why they are commented out in the code. That's one of my stray remaining todos :-P

6a68:
Ah, looks like I need to fix the download-only notice to the bottom of the viewport, not the bottom of the iframe. It's out of view on scrollable pages.

6a68:
Seems ready to land. I'll add the other PBM case (history disabled) in a followup.

johngruen:
looking

johngruen:
@6a68 a few tweaks: 
- the background color for the notice should be: `rgba(255, 255, 255, 0.1);`

6a68:
@johngruen Yeah, the notice can't be transparent, because the selection can actually go underneath it:

screen shot 2017-10-23 at 10 24 14 am

johngruen:
@6a68 okay well grey is bad...let me figure out a middle ground...

ALSO, the ? is missing two of the use cases. I tested with never remember history and 3rd party cookies disabled and DL only works in both...we should add those to the list

6a68:
One thing I'm not sure about is whether we should keep the download-only notice fixed to the bottom of the window if the user scrolls around. The download-only notice is currently redrawn whenever the selection is adjusted, but not if the screen is scrolled. Seems a little surprising to me, but edge-casey enough that the perf hit of an onScroll listener may not matter. Thoughts?

6a68:
> ALSO, the ? is missing two of the use cases. I tested with never remember history and 3rd party cookies disabled and DL only works in both...we should add those to the list

Yeah, I added the strings, but wanted to wait to display them until the behavior was actually implemented. Also, I think 3rd party cookies may not be an issue any longer, now that we have 3rd party cookie support, but need to double-check.

johngruen:
@6a68 do #737373 that's the photon approved middle grey

.gitignore
```diff
@@ -1,5 +1,7 @@
 /Profile
 .DS_Store
+*.swp
+*.swo
```
ianb: What kind of files are these?

static/css/inline-selection.scss
```diff
@@ -291,6 +359,13 @@ $very-light-grey: #ededed;
   display: block;
   margin: 5px;
   width: 40px;
+
+  &.download-only-button {
+    width: auto;
+    background-position: 7px;
+    margin-left: 10px;
```
johngruen: Kill this...it makes the buttons look a little gap-toothed

static/css/inline-selection.scss
```diff
@@ -291,6 +359,13 @@ $very-light-grey: #ededed;
...
+    padding-left: 30px;
```
johngruen: 32px

addon/webextension/background/selectorLoader.js
```diff
@@ -65,12 +65,12 @@ this.selectorLoader = (function() {
...
   exports.loadModules = function(tabId, hasSeenOnboarding) {
-    let promise;
     loadingTabs.add(tabId);
+    let promise = incognitoCheck(tabId);
     if (hasSeenOnboarding) {
-      promise = executeModules(tabId, standardScripts.concat(selectorScripts));
+      promise = promise.then(executeModules(tabId, standardScripts.concat(selectorScripts)));
     } else {
-      promise = executeModules(tabId, standardScripts.concat(onboardingScripts).concat(selectorScripts));
+      promise = promise.then(executeModules(tabId, standardScripts.concat(onboardingScripts).concat(selectorScripts)));
```
ianb: This isn't right, is it? That is, you want `promise` (aka `incognitoCheck`) to complete, *then* call `executeModules` – but this code doesn't do that. It does `incognitoCheckPromise.then(executeModulesPromise)` instead of `incognitoCheckPromise.then(callbackThatRunsExecuteModules)`

addon/webextension/background/selectorLoader.js
```diff
@@ -81,6 +81,15 @@ this.selectorLoader = (function() {
...
   };
 
+  function incognitoCheck(tabId) {
+    return browser.tabs.get(tabId).then(tab => {
+      return browser.tabs.executeScript(tabId, {
+        code: `window.incognito = ${tab.incognito};`,
+        runAt: "document_start"
+      });
+    });
+  }
```
ianb: I think this can be:

```js

function incognitoCheck(tabId) {
  return browser.tabs.get(tabId).then(tab => {
    return tab.incognito;
  });
}
```

See also [tabs.Tab](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/Tab)

6a68:
@johngruen thanks, updated. loooks like I forgot to push some linter fixes, added those too

addon/webextension/selector/ui.js
```diff
@@ -91,6 +91,31 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
     return result;
   }
 
+  function isDownloadOnly() {
+    return window.incognito;
+  }
```
ianb: I think we want the background page to control/detect this, not the content script. Or at least we have a couple other instances like some privacy settings where we might disable it. (If it's hard this could be a followup bug)

addon/webextension/selector/ui.js
```diff
@@ -91,6 +91,31 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
...
+    details.textContent = browser.i18n.getMessage("downloadOnlyDetails");
+    let detailsList = makeEl("ul");
+    let detailsPrivate = makeEl("li");
+    detailsPrivate.textContent = browser.i18n.getMessage("downloadOnlyDetailsPrivate");
+
+    detailsList.appendChild(detailsPrivate);
+    details.appendChild(detailsList);
+    tooltip.appendChild(details);
+    notice.appendChild(tooltip);
+    return notice;
```
ianb: Isn't there a function that can localize this?  I.e., we do:

```js
let notice = `<div class="download-only">
<p><span data-l10n-id="downloadOnlyNotice"></span><span class="circle">?</span>
...
`;
```

That's how we setup the iframes I think.

.gitignore
6a68: vim swapfiles

johngruen:
Sure...Last note...i'm not super stoked about how that element moves on scroll, but i'm not sure i see a way around it right now.

addon/webextension/background/selectorLoader.js
6a68: Ah, actually this function has to assign `window.incognito` inside the content script context, so that the ui.js scripts can detect private browsing mode. The window global is the content script global, which is [different](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_scripts#Content_script_environment) from the background page global and from the underlying website's global.

addon/webextension/selector/ui.js
6a68: Yeah, per my last comment, the background page does detect and assign this state. Borrowing 'incognito' from Tabs.tab probably made this harder to understand at a glance.

I think this should be ok from a security standpoint, as the incognito state isn't leaked to scripts running in the underlying page:

screen shot 2017-10-23 at 11 21 28 am

ianb: Oh, I get it then. Maybe a name like downloadOnly would be better? Also you can assign it to `this` which for subtle reasons I don't understand is slightly cleaner than `window`.

johngruen:
One last note for @youwenliang...i think the warning overall feels a little big relative to other screen UI. What do you think could we make it a bit smaller?

6a68:
> Sure...Last note...i'm not super stoked about how that element moves on scroll, but i'm not sure i see a way around it right now.

We could add a scroll handler that keeps the element at the bottom of the window. I can add this to the PR if you'd like to see if it's nicer.

addon/webextension/background/selectorLoader.js
6a68: I *think* it's correct as written, since `executeModules` returns a promise. Seems like the code works. Am I missing something?

addon/webextension/selector/ui.js
6a68: Updated. Oddly, assigning to `this` instead of `window` doesn't seem to work, so I've left that with an explanatory note.

ianb: Oh sure, that whole thing was a little superstitious anyway (well, based on feedback from @kmaglione, I think based around the levels of wrappers and performance, which wouldn't really matter here). Also of course in the module closures `this` gets mapped to something else, so it's probably all just a bit of a nuisance.

youwenliang:
@johngruen I agree we should make it smaller. Maybe a font size 16 or 14px and a padding 10px 20px instead of 12px 25px?

johngruen:
@6a68 cool i can check again later today. thanks for being patient here. this is actually the first new UI we've added in a long time!

6a68:
@johngruen sure :-) added the CSS changes @youwenliang mentioned.

I'm wondering if `white-space: nowrap` is a good idea for the message at the bottom of the screen. I wanted to avoid wrapping onto two lines, which would require extra work to position the circled question mark. The German translation ("Sie befinden sich derzeit im Nur-Download-Modus", according to google translate) doesn't seem much longer than the English, so maybe it's fine?

6a68:
@johngruen @ianb mind taking another look?

addon/webextension/background/selectorLoader.js
ianb: I don't know what the behavior of `.then()` is supposed to be when it gets a promise, but at best this is equivalent to:

```
promise = Promise.all([promise, executeModules(...)])
```

That is, as soon as the promise is created it's going to "start". Probably the promises will run in the right order because they have to run in some order, and order of creation is mostly likely, but it's undefined behavior. If you want executeModules to run *after* the other promise, then you have to only create the executeModules promise after the first promise completes.

I think the behavior will be clearer if you put an error into the incognito code – which should (but currently I believe won't) stop the rest from running.

ianb:
I'm still pretty sure the promise thing is wrong, but otherwise :+1:

addon/webextension/background/selectorLoader.js
6a68: ah, I see what you mean. `.then` expects a function which returns a promise when invoked, not the promise itself

6a68:
@ianb updated, mind taking a quick look?

@johngruen it would be great to land this for 58. how about we land this and add any tweaks in a followup?

ianb:
Looks good, @6a6: do you want to squash and merge?

6a68:
sure, thanks!

PR #3660, opened 2017-10-19
Use pg connection pooling.

### LINKED ISSUE #3371: Review pg connection pooling
Donovan thinks we never turned pg connection pooling on, because it had some potential issues and wasn't important for our performance. We should revisit this and consider turning it back on.

chenba:
Fix #3371


ianb:
We should make sure @relud is aware of this config in the next deploy.

PR #3672, opened 2017-10-25
Allow full size downloaded shots.

### LINKED ISSUE #3506: Remove 5000px limit for downloaded shots
I don't think we have as much of a reason to limit the resolution on downloaded shots.

chenba:
Fix #3506

### DIFF
--- a/addon/webextension/selector/uicontrol.js
+++ b/addon/webextension/selector/uicontrol.js
@@ -161,12 +161,12 @@ this.uicontrol = (function() {
     onClickFullPage: () => {
       sendEvent("capture-full-page", "selection-button");
       captureType = "fullPage";
-      let width = documentWidth
+      let width = getDocumentWidth();
       if (width > MAX_PAGE_WIDTH) {
         captureType = "fullPageTruncated";
       }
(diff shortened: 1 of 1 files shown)


addon/webextension/selector/uicontrol.js
```diff
@@ -161,20 +161,12 @@ this.uicontrol = (function() {
     onClickFullPage: () => {
       sendEvent("capture-full-page", "selection-button");
       captureType = "fullPage";
-      let width = Math.max(
-        document.body.clientWidth,
-        document.documentElement.clientWidth,
-        document.body.scrollWidth,
-        document.documentElement.scrollWidth);
+      let width = documentWidth
```
ianb: I had forgotten these variables exist. Unfortunately they aren't kept updated when the document resizes. I'm guessing this isn't going to work if you start Screenshots, resize the window, then capture the page.  `ui.js` keeps track of the page size for other reasons, but I don't know if that's helpful. I think I saved these as variables because I felt like the scroll tracking shouldn't re-calculate these on every event, but that doesn't count for this case.

Simply putting these in a function would be good though.

Issue #3574, opened 2017-10-02
Create "download only" mode


ianb:
There are a couple situations where we want to let the user use Screenshots and download their shot, but not allow them to upload the shot.

The three cases we are specifically thinking about:

1. From a Private Browsing window (note we also need #3274)
2. When third-party cookies are disabled (see #3572)
3. When "never remember history" is enabled (see #3196)

To do this bug, potentially only one of these cases might be handled, and we can do a follow-on for the other cases.

The second two are based on Firefox preferences, and we would have to read these preferences in bootstrap.js. This would be sent to the WebExtension in the same manner (or at the same time) as the Telemetry/FHR preference.

johngruen:
@clouserw why does this need product? We've already discussed this in our standup.

wresuolc:
@johngruen only if there is anything around when to show the label.  Eg. number 2 up there -- do we want to warn about 3rd party cookies?

Sounds like needs:ux is most appropriate now.

youwenliang:
Two quick proposals here:
https://mozilla.invisionapp.com/share/WJDTYPTGE

Option A shows a simple message about "Download-only mode" and users can click on the info button to open the tooltip and learn more about what download-only mode is.

Option B shows the whole message in a block placing at the bottom left corner.

This is a temporary copy and it should be modified by copywriters to help users better understand the situation.

johngruen:
@youwenliang lets go with option A with the caveat that i think that tooltip should just show up on hover, so we can ditch the [x] control

johngruen:
@youwenliang can you provide a measurement spec for option A?

youwenliang:
Here's the measurement: https://mozilla.github.io/testpilot-assets/Firefox_Screenshots/FFSS-3574/

PR #3687, opened 2017-10-27
Use download only mode if 'never remember history' is checked

6a68:
Since we're going to allow uploads when third-party cookies are disabled, this finishes off #3574.

Two open TODOs in the code. Otherwise, seems to work for me :beers:

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -5,6 +5,7 @@ const ADDON_ID = "screenshots@mozilla.org";
 const TELEMETRY_ENABLED_PREF = "datareporting.healthreport.uploadEnabled";
 const PREF_BRANCH = "extensions.screenshots.";
 const USER_DISABLE_PREF = "extensions.screenshots.disabled";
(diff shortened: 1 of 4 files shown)


addon/bootstrap.js
```diff
@@ -49,6 +50,21 @@ const prefObserver = {
   }
 };
 
+let historyEnabled = prefs.getBoolPref(HISTORY_ENABLED_PREF);
+// TODO: since checking "never remember history" requires a restart,
+// is it necessary to observe the places.history.enabled pref?
```
6a68: I think it might be good to observe the pref, as it's not a lot of code, and the pref _could_ be changed in about:config without triggering a restart. If nothing else, this might make QA's life a little easier.

ianb: We could also fetch it at the same time we fetch the telemetry pref, which we do before each major interaction. We could simply re-read it each time.

ianb:
Followup in #3700

PR #3721, opened 2017-11-03
Use fixed position for selection frame.

### LINKED ISSUE #3256: Does not work correctly on Firefox Send
STR:
1) Open send.firefox.com.
2) Click the icon in toolbar to take a screenshot.
3) Select any area, either by explicit mouse select or just by clicking and element.
4) Click the download button.
5) The saved screenshot contains a different page area.

Mozilla/5.0 (X11; Linux x86_64; rv:56.0) Gecko/20100101 Firefox/56.0 ID:20170801100311 CSet: 51ffb9283f0c7c00e08eb8c39b33fbee218c370d

chenba:
Fix #3256

### DIFF
--- a/addon/webextension/selector/ui.js
+++ b/addon/webextension/selector/ui.js
@@ -137,7 +137,7 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
           this.element = initializeIframe();
           this.element.id = "firefox-screenshots-selection-iframe";
           this.element.style.display = "none";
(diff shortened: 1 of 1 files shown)


6a68:
@chenba i'm pretty sure the selection iframe is not fixed so that users can scroll around and change their selection

chenba:
I'm deleting this PR (and my branch) since the alternative approach won't involve changing the `position` attr of the selection frame, so there's no need to keep that commit.

PR #3723, opened 2017-11-03
Fix #3674, put 50% of people into control

### LINKED ISSUE #3674: Make ab-tests create 50/50 cohorts
We should change ab-tests.js to do cohort selection, then 50/50 control/experiment (or 33/33/33 for three tests, etc).  The cohort selection might be 99/1, 90/10, 90/10 except only en-US, etc.

I think we should change from just `("control", experiment_name)` to `("excluded", "control", experiment_name)`.

ianb:
This separates out people who are excluded from a test, from those who are in the control for the test. 50% of people always go to control, and 50% go into some experiment branch. People who aren't in either go into the 'exclude' category.

### DIFF
--- a/server/src/ab-tests.js
+++ b/server/src/ab-tests.js
@@ -1,8 +1,15 @@
+/** A/B test assignment support
+
+This assigns users to tests using the `abTests.updateAbTests(obj)` function
+
+Tests are defined directly in this file, in the `allTests` variable.  A comment
+below shows what these test objects look like.
+*/
 // Note: these get turned into Test objects:
 let allTests = {
 };
 
-/* Example of how this could be set (until we have real tests to serve as docs): */
+/* Example of how this could be set: */
 /*
 let allTests = {
   autoOpenSharePanel: {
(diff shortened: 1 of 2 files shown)


server/src/ab-tests.js
```diff
@@ -39,7 +49,8 @@ let allTests = {
     version: 1,
     exclude: ["highlightButtonOnInstall", "autoOpenSharePanel"],
     options: [
-      // Note no one will end up in control in this example:
+      // Note no one will end up in exclude in this example (but 50% will still
+      // be control):
```
6a68: @ianb If I'm understanding this correctly, since 10% of the population is in the `autoOpenSharePanel` test, and another 10% is in the `highlightButtonOnInstall` test, and people in those tests are excluded from this `myShotsDisplay` test, then, 50% of the remaining total 80% of users will be in the control group in this test, while the other half will be 9:1 split between the 'intropopup' and 'blink' tests. Does that seem correct?

PR #3726, opened 2017-11-06
Fix #3718, send a response after incrementing counts

6a68:
@chenba Mind giving this a try? After applying this patch, dev edition should still complain about an "unknown scalar", but should otherwise work

PR #3727, opened 2017-11-07
Fix #3673, instrument response times in shot creation flow

### LINKED ISSUE #3673: Instrument interaction time in shot creation flow
As part of the RAIL perf guidelines, we want to ensure the app responds in <100ms when the user clicks on a button.

I don't think it's possible to measure time from the page action click to the overlay being drawn, but we should be able to instrument time from click to paint for the other steps. Since all the frames live inside the same window, the `Performance` API should work fine.

I'm not sure how hard it'll be to instrument click to paint, but click to next callback might be a good enough proxy.

6a68:
The metrics docs changes cover the basic overview.

The iframe unhide events seem to fire slightly before the UI is updated, but given how horrifyingly slow some of the steps are (especially full-page previews & downloads), I suspect these start and end points will be good enough for a while.

Easiest way to test this is to filter on 'perf' in the browser console, then run through various shot creation + download/upload flows:

screen shot 2017-11-06 at 4 22 07 pm

### DIFF
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -8,6 +8,33 @@ this.analytics = (function() {
   let telemetryPrefKnown = false;
   let telemetryPref;
 
+  function sendTiming(timingLabel, timingVar, timingValue) {
+    // sendTiming is only called in response to sendEvent, so no need to check
+    // the telemetry pref again here.
+    let timingCategory = "addon";
+    let url = main.getBackend() + "/timing";
+    let req = new XMLHttpRequest();
+    req.open("POST", url);
+    req.setRequestHeader("content-type", "application/json");
+    req.onload = catcher.watchFunction(() => {
+      if (req.status >= 300) {
+        let exc = new Error("Bad response from POST /timing");
+        exc.status = req.status;
+        exc.statusText = req.statusText;
+        console.error(exc);
+        // TODO: send the error to sentry, maybe?
+      }
+    });
+    log.info(`sendTiming ${timingCategory}/${timingLabel}/${timingVar}: ${timingValue}`);
+    req.send(JSON.stringify({
+      deviceId: auth.getDeviceId(),
+      timingCategory,
+      timingLabel,
+      timingVar,
+      timingValue
+    }));
+  }
+
   exports.sendEvent = function(action, label, options) {
     let eventCategory = "addon";
     if (!telemetryPrefKnown) {
@@ -183,7 +210,7 @@ this.analytics = (function() {
       } else if (timingData[r.name] && match(r.end, action, label)) {
         let endTime = Date.now();
         let elapsed = endTime - timingData[r.name];
-        exports.sendEvent('perf-response-time', r.name, {cd1: elapsed});
+        sendTiming("perf-response-time", r.name, elapsed);
         delete timingData[r.name];
       }
     });
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -194,11 +194,7 @@ this.analytics = (function() {
   }
 
   function anyMatches(filters, action, label) {
-    let matched = false;
-    filters.forEach(filter => {
-      matched = matched || match(filter, action, label);
-    });
-    return matched;
+    return !!filters.find(filter => match(filter, action, label));
   }
 
   function measureTiming(action, label) {
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -18,6 +18,12 @@ this.analytics = (function() {
       log.info(`Cancelled sendEvent ${eventCategory}/${action}/${label || 'none'} ${JSON.stringify(options)}`);
       return Promise.resolve();
     }
+    measureTiming(action, label);
+    // Internal-only events are used for measuring time between events,
+    // but aren't submitted to GA.
+    if (action === 'internal') {
+      return Promise.resolve();
+    }
     if (typeof label == "object" && (!options)) {
       options = label;
       label = undefined;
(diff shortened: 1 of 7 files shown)


addon/bootstrap.js
```diff
@@ -230,8 +230,10 @@ function handleMessage(msg, sender, sendReply) {
     sendReply({type: "success", value: historyEnabled});
   } else if (msg.funcName === "incrementDownloadCount") {
     Services.telemetry.scalarAdd('screenshots.download', 1);
+    sendReply({type: "success", value: true});
   } else if (msg.funcName === "incrementUploadCount") {
     Services.telemetry.scalarAdd('screenshots.upload', 1);
+    sendReply({type: "success", value: true});
```
6a68: The changes in this file are the patch from #3726

6a68:
It occurred to me in the shower this morning that it might be nicer to keep the times inside the uicontrol state object, rather than overload sendEvent to turn it into a quasi-messaging broker with the internal-only events and separate time tracking. Might be something to fix in a refactoring bug later?

wresuolc:
This is important to land for our OKR :)

6a68:
@ianb cool, restored `/timing` (slightly modified) and updated the addon code accordingly. R?

addon/webextension/background/analytics.js
```diff
@@ -77,5 +110,111 @@ this.analytics = (function() {
...
+    return filter.label ?
+      filter.action === action && filter.label === label :
+      filter.action === action;
+  }
+
+  function anyMatches(filters, action, label) {
+    let matched = false;
+    filters.forEach(filter => {
+      matched = matched || match(filter, action, label);
+    });
```
chenba: Nitpick: since we are only interested in the first matching filter, we can use `array.find()` here.

server/src/server.js
```diff
@@ -451,6 +451,29 @@ app.post("/event", function(req, res) {
...
+      userAnalytics = userAnalytics.debug();
+    }
+    let params = {
+      userTimingCategory: bodyObj.timingCategory,
+      userTimingVariableName: bodyObj.timingVar,
+      userTimingTime: bodyObj.timingValue,
+      userTimingLabel: bodyObj.timingLabel
+    };
+    userAnalytics.timing(params).send();
+    simpleResponse(res, "OK", 200);
```
chenba: Previously `/timing` allowed batching.  That would make the code on the addon side a little more complicated though.  Perhaps keep the completed timing data in one object and the in progress ones in another, so whenever teh in progress list is empty and there are completed timing measurements, send those to `/timing`?

6a68: Yeah, I noticed that too. Since `sendEvent` fires many more pings, and fires each one individually, it didn't seem worth bothering with the extra state management.

If there's a good reason to batch events in general, maybe we can adjust both endpoints in a followup bug?

ianb: I made a followup in #3757

addon/webextension/background/analytics.js
```diff
@@ -8,6 +8,33 @@ this.analytics = (function() {
   let telemetryPrefKnown = false;
   let telemetryPref;
 
+  function sendTiming(timingLabel, timingVar, timingValue) {
+    // sendTiming is only called in response to sendEvent, so no need to check
+    // the telemetry pref again here.
+    let timingCategory = "addon";
+    let url = main.getBackend() + "/timing";
+    let req = new XMLHttpRequest();
+    req.open("POST", url);
```
ianb: We should probably start using `fetch()` (though I also keep falling back to XHR too ;)

addon/webextension/background/analytics.js
```diff
@@ -8,6 +8,33 @@ this.analytics = (function() {
...
+    req.setRequestHeader("content-type", "application/json");
+    req.onload = catcher.watchFunction(() => {
+      if (req.status >= 300) {
+        let exc = new Error("Bad response from POST /timing");
+        exc.status = req.status;
+        exc.statusText = req.statusText;
+        console.error(exc);
+        // TODO: send the error to sentry, maybe?
```
ianb: This should reject a promise. Also it should create a promise! Same as sendEvent. I'm not sure how consistently we watch the sendEvent promise, arguably we should have a wrapped version of these functions for the not-uncommon case where we want to send-and-forget (with `catcher.watchFunction(sendTiming, true)`)

addon/webextension/selector/ui.js
```diff
@@ -446,6 +448,7 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
 
     unhide() {
       this.element.style.display = "";
+      catcher.watchPromise(callBackground("sendEvent", "internal", "unhide-preview-frame"));
```
ianb: I'm a little concerned that callBackground is going to confound the timing measurements a little. But... eh?

docs/METRICS.md
```diff
@@ -177,6 +179,75 @@ The onboarding slides have some events:
...
 sendEvent("click-install-firefox-home", {useBeacon: true});
+
+#### Add-on performance measurements
+
+
+##### Internal-only events
+
+Internal-only events are used to measure the time from user input to user-visible UI response.
+
+*NOTE: Internal-only events are not submitted to GA.*
```
ianb: Probably these don't even need to be documented, but... well, no harm.

docs/METRICS.md
```diff
@@ -177,6 +179,75 @@ The onboarding slides have some events:
...
+
+1. [x] New tab opened for newly-created shot page `addon/internal/open-shot-tab`
+1. [x] Screenshots UI hidden by uicontrol `addon/internal/deactivate`
+1. [x] Pre-selection iframe shown `addon/internal/unhide-preselection-frame`
+1. [x] Selection iframe shown `addon/internal/unhide-selection-frame`
+1. [x] Preview iframe shown `addon/internal/unhide-preview-frame`
+
+##### First step: starting the shot
+
+1. [x] Time from clicking the page action (or toolbar button) to displaying the preselection iframe, `addon/perf-response-time/page-action` with `cd1: {ms response time}`
```
ianb: I think these need to be adjusted to reflect they aren't events, but timings.

addon/webextension/selector/ui.js
6a68: Yeah, the time to send a message seems pretty small, <=10ms (for short messages), and some of these steps take several seconds, so it seemed fine for now.

addon/webextension/background/analytics.js
6a68: sure, already filed as #2261

addon/webextension/background/analytics.js
```diff
@@ -12,27 +12,30 @@ this.analytics = (function() {
...
-        exc.statusText = req.statusText;
-        console.error(exc);
-        // TODO: send the error to sentry, maybe?
-      }
+    return new Promise((resolve, reject) => {
+      let url = main.getBackend() + "/timing";
+      let req = new XMLHttpRequest();
+      req.open("POST", url);
+      req.setRequestHeader("content-type", "application/json");
+      req.onload = catcher.watchFunction(() => {
```
ianb: Technically I don't think this will reject the promise, if there's an exception in the function then the promise will just hang. But it will report something to Sentry.

PR #3739, opened 2017-11-09
Properly set the Photon page action's title and icon.

0c0w3:
Bug 1395387 changed the Photon page action API so that `title`
and `iconURL` are no longer properties but methods, setTitle()
and setIconURL().  We need to update our consumer.  Right now,
setting these properties isn't doing anything at all.


ianb:
Thanks! I've made a note we need to get this exported into Firefox 58

PR #3743, opened 2017-11-09
Use download only notice style for cropped image warning.

### LINKED ISSUE #3701: Use download-only warning style for height limit warning
I think we have two styles of warnings for these two cases

chenba:
Fix #3701

### DIFF
--- a/addon/webextension/selector/ui.js
+++ b/addon/webextension/selector/ui.js
@@ -108,8 +108,8 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
         </ul>
       </div>
       <tbody>
-        <tr class="download-only-wrapper">
-          <td class="download-only-content" data-l10n-id="downloadOnlyNotice"></td>
+        <tr class="notice-wrapper">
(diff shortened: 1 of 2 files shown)


addon/webextension/selector/ui.js
```diff
@@ -841,9 +841,15 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
       img.src = dataUrl;
       iframe.document().querySelector(".preview-image").appendChild(img);
       if (showCropWarning && !(isDownloadOnly())) {
-        let imageCroppedEl = makeEl("DIV");
-        imageCroppedEl.id = "imageCroppedWarning";
-        imageCroppedEl.textContent = browser.i18n.getMessage("imageCroppedWarning", buildSettings.maxImageHeight);
+        let imageCroppedEl = makeEl("table", "image-cropped-warning");
```
6a68: Looks like you've left the 'image-cropped-warning' class on this table, and the other table has the 'download-only' class applied to it. Any reason a 'notice' class couldn't be applied to both?

PR #3750, opened 2017-11-11
Fix #3745, enforce image size limitations for drag selection

### LINKED ISSUE #3745: The Image Height Limit Can Be Bypassed by Dragging the Hightlight Rectangle
### STR

1. Go to a loooong page (some page with a lot of user comments should work...)
1. Start Screenshots
1. Start a selection rectangle by clicking on an element or dragging.
1. Now drag the bottom edge of rectangle down by dragging
1. Save the shot

### Result

Shot is saved as selected.

### Expected

Shot is cropped to the image height limit.

testeaxeax:
Fix #3745

### DIFF
--- a/addon/webextension/selector/uicontrol.js
+++ b/addon/webextension/selector/uicontrol.js
@@ -130,6 +130,13 @@ this.uicontrol = (function() {
       exports.deactivate();
     }, save: () => {
       sendEvent("save-shot", "overlay-save-button");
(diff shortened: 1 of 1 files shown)


ianb:
Thanks for the patch @testeaxeax. We decided not to place any size restriction on people manually selecting a region – if someone puts in all the effort to make a giant selection, we're going to let them upload it regardless.

Issue #3661, opened 2017-10-20
webextension strings sometimes get perma-cached


6a68:
Not sure what the underlying issue is, but it seems like changing the webextension strings and bumping the version number isn't enough to pick up new strings. However, if I navigate to the messages.json URL, I do see the correct strings. Some cache seems to be stale. Switching between dev edition and nightly clears the cache.

kmag suggested that we need to pass the version number to the startup method; need to dig into this and ask him when I understand the code a little better.

This line might be relevant: https://dxr.mozilla.org/mozilla-central/source/toolkit/mozapps/extensions/internal/XPIProvider.jsm#2029

johngruen:
"someone ask aswan to ask kris to explain it"

chenba:
I've found that the strings get updated reliably on the second restart.  That is, make some changes, `bin/run-addon`, exit, `bin/run-addon` once more.  I noticed that the version string passed in the startup data does not change while the version string in the manifest changes during the restarts.  (Occasionally I saw that the version string from deviceInfo, which uses browser.runtime.getManifest(), being different than the version in manifest.json on disk...)

I don't understand how XPIProvider's startup would help here.  That's what is calling bootstrap's startup looks like?  And bootstrap's startup receives the version number in the startup data so I don't understand the part where "we need to pass the version number to the startup method".

6a68:
bootstrap's startup receives the version number, but we don't pass that number into the webextension's startup method. we probably need to ask in #teamaddons to understand how to correctly pass in that number

6a68:
probably [this](https://github.com/mozilla-services/screenshots/blob/master/addon/bootstrap.js#L182) is the line that needs to be changed - but there are just a handful of 'embedded webextensions' in existence, so the (internal) APIs we're using aren't really documented in the same way that public webextension APIs are

chenba:
EmbeddedExtension.startup doesn't take a version number as one of its params though.  It only takes the startup reason.

6a68:
You're right @chenba, the version number is stale. I think this may be an artifact of using the unmaintained `jpm` tool inside `./bin/run-addon`. If I bump the version number, create the xpi via `make unsigned_bootstrap_xpi`, then manually install it via `about:addons`, the version number passed to startup is correct.

The best long-term solution is to get away from jpm by removing the bootstrapped part of the webextension (#3875). Closing this, as it seems unlikely to affect real world users.

Issue #1646, opened 2016-09-29
Add "copy image" to shot pages


ianb:
Sometimes, like with GitHub issues, I'd really rather copy the image directly.  We have the URL copy but not the image copy.

johngruen:
Kicking this out of blue sky since we're looking at copying to clipboard from the add-on. This would make sense as a follow up.

chenba:
This is a dup of #2582

johngruen:
@youwenliang  we need a spec to get started on this one!

youwenliang:
@johngruen I had the measurements here:
-> https://github.com/mozilla-services/screenshots/issues/2582#issuecomment-337473248
Should we do an A/B testing for the button (icon-only & icon+text)?

chenba:
AFAIK copying images to the clipboard (outside of an addon) is not possible.  😕

PR #3758, opened 2017-11-13
Copy image to clipboard

### LINKED ISSUE #2582: Add option to copy image (not just link) *to* clipboard
For pasting page shots into github or irccloud (for example), it would be pretty nifty to have an option that just copied the image content directly to the clipboard.

chenba:
Fix #2582

### DIFF
--- a/locales/en-US/webextension.properties
+++ b/locales/en-US/webextension.properties
@@ -24,6 +24,14 @@ notificationLinkCopiedTitle = Link Copied
 # modifier key used in the shortcut (do not translate it): for example, Ctrl-V
 # on Windows systems.
 notificationLinkCopiedDetails = The link to your shot has been copied to the clipboard. Press {meta_key}-V to paste.
+# This is a verb, as in "Copy image to the system clipboard."
+copyScreenshot = Copy
+notificationImageCopiedTitle = Shot Copied
+# The string "{meta_key}-V" should be translated to the region-specific
+# shorthand for the Paste keyboard shortcut. {meta_key} is a placeholder for the
+# modifier key used in the shortcut (do not translate it): for example, Ctrl-V
+# on Windows systems.
+notificationImageCopiedDetails = Your shot has been copied to the clipboard. Press {meta_key}-V to paste.
 # This is displayed as a warning when a Full Page save will be cropped.
 # {pixels} will be a number like 5000.
 imageCroppedWarning = This image has been cropped to {pixels}px.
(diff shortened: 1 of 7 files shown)


locales/en-US/webextension.properties
```diff
@@ -24,6 +24,9 @@ notificationLinkCopiedTitle = Link Copied
 # modifier key used in the shortcut (do not translate it): for example, Ctrl-V
 # on Windows systems.
 notificationLinkCopiedDetails = The link to your shot has been copied to the clipboard. Press {meta_key}-V to paste.
+copyScreenshot = Copy
```
flodolo: This string needs a localization comment, since it's unclear if it's a noun or a verb without context.

johngruen:
@chenba gonna pull this down and check it out now

johngruen:
@chenba 3 minor things:

1. For download only mode, keep Download as the default option
screen shot 2017-11-14 at 7 10 20 pm

2. For normal mode, flip save and copy
screen shot 2017-11-14 at 7 09 35 pm

3. The SVG fill color should match the other two...make `fill: #3e3d40`

johngruen:
ALSO...left need a string change on the message

chenba:
@johngruen Need to confirm: do you want to swap the download and copy buttons _only_ for download-only mode?  Or, we keep them consistent, always Cancel, Copy, Download [, Save]?

chenba:
> ALSO...left need a string change on the message

@johngruen I don't understand this comment.  Left?

chenba:
@johngruen Swapped the position of the copy and download buttons.  Take another look?

ianb:
The code looks good, though for some reason the translations aren't working for me. I've even done a `make clean`. The translations are obviously there, but I get warnings like `Unknown localization message copyscreenshot`

chenba:
@ianb Yeah, this is related to #3661.  If you load the addon through `about:debugging` then you'll definitely see the new l10n messages.

ianb:
OK, we can merge (should also get squashed) once @johngruen gives his 👍

chenba:
@flodolo Updated a couple of l10n messages in the latest commit (https://github.com/mozilla-services/screenshots/pull/3758/commits/8e33cb73f9cf8b6d9aa0a7ae09d85ba329479dec).  Could you take a quick look?

eyaler:
is there a copy button also in the my shots library?
how about binding with ctrl-c?

chenba:
@eyaler ctrl-c is in this patch.  There's no copy (to clipboard) button in My Shots; an issue's been filed for adding one to the page of an individual shot (#1646).  Could you file one for My Shots?

eyaler:
here: https://github.com/mozilla-services/screenshots/issues/3792

PR #3768, opened 2017-11-14
Fix bug when UITour-lib.js loads after homepage-bundle.js

ianb:
The UITour-lib.js library is loaded async, and the homepage-bundle can potentially load and run before it (we saw this in Sentry). This puts in a 1 second retry when that happens.

### DIFF
--- a/server/src/pages/homepage/controller.js
+++ b/server/src/pages/homepage/controller.js
@@ -29,6 +29,7 @@ document.addEventListener("addon-present", () => {
             }
             return;
           }
+          clearTimeout(interval);
           Mozilla.UITour.showHighlight("screenshots");
         }, 100);
       } else {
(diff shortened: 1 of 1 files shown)


server/src/pages/homepage/controller.js
```diff
@@ -17,7 +17,23 @@ document.addEventListener("addon-present", () => {
...
+        let interval = setInterval(() => {
+          if (typeof Mozilla == "undefined") {
+            count--;
+            if (count <= 0) {
+              clearTimeout(interval);
+              throw new Error("UITour-lib.js didn't load up after 1 second");
+            }
+            return;
+          }
+          Mozilla.UITour.showHighlight("screenshots");
```
chenba: Shouldn't `clearInterval` be called here?

PR #3775, opened 2017-11-14
Use var instead of let in reactrender

ianb:
This code is run literally without any babel translation, and shouldn't use fancy JavaScript


chenba:
👍

PR #3795, opened 2017-11-16
s/Photon/WebExtension/ for page action.

### LINKED ISSUE #3756: Remove iconPath-related messages from bootstrap.js
This is a followup to #3747 (00f04091ffa4200009dd6b9bfe38ccdc34370157).

Now that we don't dynamically set the icon for onboarding, we can remove all icon-resetting code from bootstrap.js.

See also [Bug 1412377](https://bugzilla.mozilla.org/show_bug.cgi?id=1412377) (change it all to a pageAction)

chenba:
Fix #3756

### DIFF
--- a/addon/webextension/background/startBackground.js
+++ b/addon/webextension/background/startBackground.js
@@ -31,7 +31,7 @@ this.startBackground = (function() {
   // Maximum milliseconds to wait before checking for migration possibility
   const CHECK_MIGRATION_DELAY = 2000;
 
-  browser.tabs.onUpdated.addListener((id, changeInfo, tab) => {
+  browser.tabs.onCreated.addListener(tab => {
     browser.pageAction.show(tab.id);
   });
 
--- a/addon/webextension/background/startBackground.js
+++ b/addon/webextension/background/startBackground.js
@@ -1,6 +1,6 @@
 /* globals browser, main, communication */
 /* This file handles:
-     clicks on the Photon page action
+     clicks on the WebExtension page action
      browser.contextMenus.onClicked
      browser.runtime.onMessage
    and loads the rest of the background page in response to those events, forwarding
@@ -31,6 +31,18 @@ this.startBackground = (function() {
   // Maximum milliseconds to wait before checking for migration possibility
   const CHECK_MIGRATION_DELAY = 2000;
 
+  browser.tabs.onUpdated.addListener((id, changeInfo, tab) => {
+    browser.pageAction.show(tab.id);
+  });
+
+  browser.pageAction.onClicked.addListener(tab => {
+    loadIfNecessary().then(() => {
+      main.onClicked(tab);
+    }).catch(error => {
+      console.error("Error loading Screenshots:", error);
+    });
+  });
+
   browser.contextMenus.create({
     id: "create-screenshot",
     title: browser.i18n.getMessage("contextMenuLabel"),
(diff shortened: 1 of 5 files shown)


addon/webextension/background/startBackground.js
```diff
@@ -31,6 +31,18 @@ this.startBackground = (function() {
   // Maximum milliseconds to wait before checking for migration possibility
   const CHECK_MIGRATION_DELAY = 2000;
 
+  browser.tabs.onUpdated.addListener((id, changeInfo, tab) => {
+    browser.pageAction.show(tab.id);
+  });
```
ianb: Do we need this? I'm pretty sure there's a substantial performance impact to using `browser.tabs.unUpdated`

chenba:
Good catch.  Thanks!  Switch to `onCreated`.

chenba:
Unfortunately `onCreated` does not work consistently.  The page action button is clickable for Hacker News and NPR sometimes.  Seems like the button is less likely to be clickable for the first NPR or Hacker News tab.  With Google News, the Screenshots page action button is never clickable.

ianb:
I think if `page_action` gets added to the manifest, then the page actions will just show up, and we'll never have to call `.show()`. ([Per these docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/page_action))

chenba:
`page_action` is in the manifest with this patch (https://github.com/mozilla-services/screenshots/pull/3795/files#diff-edd690115570044e76445f43adc8ba8bR33).  And the doc says "While browser actions are displayed by default, page actions are hidden by default. They can be shown for a particular tab by calling pageAction.show(), passing in the tab's ID."

ianb:
There's a [treeherder build in progress](https://treeherder.mozilla.org/#/jobs?repo=try&revision=4df69ca1cf33a7f47aac68d2afc62c43367bb3d9) with a [Talos performance test](https://treeherder.mozilla.org/perf.html#/comparechooser?newProject=try&newRevision=4df69ca1cf33a7f47aac68d2afc62c43367bb3d9)

It'll take a while to run, but we can look at the performance together when it finishes. (The output requires some interpretation)

chenba:
Couple more notes regarding `tabs.onCreated`.  How the tab is created makes a difference.  If a blank tab is opened (ctrl-t, + button, etc.) and then the user go to a site, the Screenshots page action button will not be clickable.  This make sense since no page actions are displayed on a blank page, and Screenshots' `pageAction.show()` is not called on the subsequent updates.  But even this behavior is not totally consistent.  Occasionally, I can ctrl-t, type in a URL, and the Screenshots button is clickable.

If the user opens a tab with a URL (e.g. by middle-clicking), then the page action button is clickable.  

Also, any tab(s) that starts with the browser will not have a clickable Screenshots button.

ianb:
The last Try build failed a couple times, basic build stuff so I assume nothing related to what I'm doing, I rebased against central again and am retrying:

* Try build: https://treeherder.mozilla.org/#/jobs?repo=try&revision=2adf735fe1686490e288b33e699ea5b92d13a9e5
* Talos/perf: https://treeherder.mozilla.org/perf.html#/comparechooser?newProject=try&newRevision=2adf735fe1686490e288b33e699ea5b92d13a9e5

ianb:
We're seeing a notable performance regression:

image

Note this test was taken with 4975a71e2810c63d56dcce56534f3247750a7150 (using onCreated and not onUpdated), *but* we would actually need to use onUpdated, which would make the impact worse.

After some discussion on IRC with @kmaglione the performance hit seems to be inevitable with the current APIs. I have opened [Bug 1419842](https://bugzilla.mozilla.org/show_bug.cgi?id=1419842) to propose an API change to alleviate this.

I'm closing this PR now, as we'll need some kind of WebExtension changes before this can land.

PR #3811, opened 2017-11-20
Batch event and timing analytics data.

### LINKED ISSUE #3757: Batch events and timings
Right now we get events (and with #3727 timings) one-by-one to our /events and /timing endpoints. It would be better to batch these, since some events happen very quickly in succession. Batching can be as simple as:

```js
let pendingEvents = [];
let eventTimer = null;
function sendEvent(event) {
  // do normal stuff to prepare the event
  // set the time of the event
  pendingEvents.push(event);
  if (!eventTimer) {
    eventTimer = setTimeout(() => {
      eventTimer = null;
      flushAllEvents();
      pendingEvents = [];
    }, 1000);
  }
}
```

We should set [queueTime](https://github.com/peaksandpies/universal-analytics/blob/master/AcceptableParams.md#queue-time) to backdate the events. Note that means we should record how long it was between `sendEvent()` being called, and `flushAllEvents()` (per-event).

chenba:
Fix #3757 

This uses `window.requestIdleCallback`.  For the current user timings, it makes no difference because those are relatively far apart.  If the user actions are quick enough, some batching occurs with event tracking.  Of course, the slower/busier the computer, the more batching potential.  Everyone's Mileage May Vary.

The server side of this patch is backwards compatible with older clients.

### DIFF
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -8,34 +8,69 @@ this.analytics = (function() {
   let telemetryPrefKnown = false;
   let telemetryPref;
 
+  const EVENT_BATCH_DURATION = 1000; // ms for setTimeout
+  let pendingEvents = [];
+  let pendingTimings = [];
+  let eventsTimeoutHandle, timingsTimeoutHandle;
+  const fetchOptions = {
+    method: "POST",
+    mode: "cors",
+    headers: { "content-type": "application/json" },
+    credentials: "include"
+  };
+
+  function flushEvents() {
+    if (pendingEvents.length === 0) {
+      return;
+    }
+
+    let eventsUrl = `${main.getBackend()}/event`;
+    let deviceId = auth.getDeviceId();
+    let sendTime = Date.now();
+
+    pendingEvents.forEach(event => {
+      event.queueTime = sendTime - event.eventTime
+      log.info(`sendEvent ${event.event}/${event.action}/${event.label || 'none'} ${JSON.stringify(event.options)}`);
+    });
+
+    let body = JSON.stringify({deviceId, events: pendingEvents});
+    let fetchRequest = fetch(eventsUrl, Object.assign({body}, fetchOptions));
+    fetchWatcher(fetchRequest);
+    pendingEvents = [];
+  }
+
+  function flushTimings() {
+    if (pendingTimings.length === 0) {
+      return;
+    }
+
+    let timingsUrl = `${main.getBackend()}/timing`;
+    let deviceId = auth.getDeviceId();
+    let body = JSON.stringify({deviceId, timings: pendingTimings});
+    let fetchRequest = fetch(timingsUrl, Object.assign({body}, fetchOptions));
+    fetchWatcher(fetchRequest);
+    pendingTimings.forEach(t => {
+      log.info(`sendTiming ${t.timingCategory}/${t.timingLabel}/${t.timingVar}: ${t.timingValue}`);
+    });
+    pendingTimings = [];
+  }
+
   function sendTiming(timingLabel, timingVar, timingValue) {
     // sendTiming is only called in response to sendEvent, so no need to check
     // the telemetry pref again here.
     let timingCategory = "addon";
-    return new Promise((resolve, reject) => {
-      let url = main.getBackend() + "/timing";
-      let req = new XMLHttpRequest();
-      req.open("POST", url);
-      req.setRequestHeader("content-type", "application/json");
-      req.onload = catcher.watchFunction(() => {
-        if (req.status >= 300) {
-          let exc = new Error("Bad response from POST /timing");
-          exc.status = req.status;
-          exc.statusText = req.statusText;
-          reject(exc);
-        } else {
-          resolve();
-        }
-      }, true);
-      log.info(`sendTiming ${timingCategory}/${timingLabel}/${timingVar}: ${timingValue}`);
-      req.send(JSON.stringify({
-        deviceId: auth.getDeviceId(),
-        timingCategory,
-        timingLabel,
-        timingVar,
-        timingValue
-      }));
+    pendingTimings.push({
+      timingCategory,
+      timingLabel,
+      timingVar,
+      timingValue
     });
+    if (!timingsTimeoutHandle) {
+      timingsTimeoutHandle = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


addon/webextension/background/analytics.js
```diff
@@ -8,34 +8,70 @@ this.analytics = (function() {
...
+      event.queueTime = sendTime - event.eventTime
+      log.info(`sendEvent ${event.event}/${event.action}/${event.label || 'none'} ${JSON.stringify(event.options)}`);
+    });
+
+    let body = JSON.stringify({deviceId, events});
+    let fetchRequest = fetch(eventsUrl, Object.assign({body}, fetchOptions));
+    fetchWatcher(fetchRequest);
+
+    events = [];
+    eventsIdleCallbackHandle = null;
```
ianb: I don't think there's any good way for this to happen, but potentially if `events` is empty then the handle won't be nulled out. I'd be inclined to reset `eventsIdleCallbackHandle` in the same place it is initialized. But as a separate issue it's not initialized.

addon/webextension/background/analytics.js
```diff
@@ -8,34 +8,70 @@ this.analytics = (function() {
...
-        timingValue
-      }));
+    timings.push({
+      timingCategory,
+      timingLabel,
+      timingVar,
+      timingValue
     });
+    if (!timingsIdleCallbackHandle) {
+      window.requestIdleCallback(flushTimings);
```
ianb: Reading about [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback), it doesn't really seem like the right thing for what we're doing.

We *want* to batch timings, not just to avoid doing work when the code isn't idle, but also to reduce overall network. There is a substantial HTTP overhead to each request after all.

My reading of requestIdleCallback is that it can be run extremely soon, and doesn't have any real delay. There's a `timeout` parameter, but that's a maximum time, not a minimum time. If we were being fancy we'd actually want to do both `setTimeout` (to establish a minimum) and `requestIdleCallback` (to schedule the task during idle time).

addon/webextension/background/analytics.js
```diff
@@ -8,34 +8,70 @@ this.analytics = (function() {
...
-        timingVar,
-        timingValue
-      }));
+    timings.push({
+      timingCategory,
+      timingLabel,
+      timingVar,
+      timingValue
     });
+    if (!timingsIdleCallbackHandle) {
```
ianb: `timingsIdleCallbackHandle` (and the other handle) aren't initialized at all. I'd rather they be both initialized and reset in the same place, like:

```js
if (!timingsIdleCallbackHandle) {
  timingsIdleCallbackHandle = setTimeout(() => {
    timingsIdleCallbackHandle = null;
    flushTimings();
  }, EVENT_BATCH_TIME);
}
```

addon/webextension/background/analytics.js
```diff
@@ -209,11 +234,26 @@ this.analytics = (function() {
...
+      request.then(response => {
+        if (response.status >= 400) {
+          let exc = new Error("Bad response from POST /timing");
+          exc.status = response.status;
+          exc.statusText = response.statusText;
+          reject(exc);
+        }
+      }).catch(error => {
+        reject(error);
+      });
```
ianb: This could be simpler, like:

```js
function fetchWatcher(request) {
  return request.then((resp) => {
    if (!resp.ok) {
      throw new Error(`Bad response: ${resp.status} ${resp.statusText}`); // Maybe include request.url? Or just set these things on the exception like you did above
    }
    return resp;
  });
}
```

Throwing an exception in a then handler is the same as rejecting it.

addon/webextension/background/analytics.js
chenba: At first I had a hard time understanding your comment about not initializing the handles, but then I saw that I didn't actually assign the return value of the line below to the handle var.  D:

chenba:
@ianb Changed to `setTimeout`.  I'd like your take on flushing on browser shutdown vs. losing some data because the browser exited before the setTimeout delay expires.  Technically it's possible, but during normal workflow it seems very unlikely.

addon/webextension/background/analytics.js
```diff
@@ -8,34 +8,68 @@ this.analytics = (function() {
   let telemetryPrefKnown = false;
   let telemetryPref;
 
+  const EVENT_BATCH_DURATION = 1000; // ms for setTimeout
+  let events = [];
+  let timings = [];
```
ianb: I think calling these "pendingEvents" or "batchedEvents" or something like that would be clearer.

addon/webextension/background/analytics.js
```diff
@@ -8,34 +8,68 @@ this.analytics = (function() {
...
+  function flushTimings() {
+    if (timings.length === 0) {
+      return;
+    }
+
+    let timingsUrl = `${main.getBackend()}/timing`;
+    let deviceId = auth.getDeviceId();
+    let body = JSON.stringify({deviceId, timings});
+    let fetchRequest = fetch(timingsUrl, Object.assign({body}, fetchOptions));
+    fetchWatcher(fetchRequest);
```
ianb: This does not reset `timings` to `[]`! Ditto `events`

chenba:
@ianb Really truly emptying out those arrays this time.  Also squashed the commits.  r?  Thanks!

addon/webextension/background/analytics.js
```diff
@@ -8,34 +8,70 @@ this.analytics = (function() {
...
+    let timingsUrl = `${main.getBackend()}/timing`;
+    let deviceId = auth.getDeviceId();
+    let body = JSON.stringify({deviceId, timings: pendingTimings});
+    let fetchRequest = fetch(timingsUrl, Object.assign({body}, fetchOptions));
+    fetchWatcher(fetchRequest);
+    pendingTimings = [];
+
+    pendingTimings.forEach(t => {
+      log.info(`sendTiming ${t.timingCategory}/${t.timingLabel}/${t.timingVar}: ${t.timingValue}`);
+    });
```
ianb: Oops, the order here doesn't make sense, since you cleared the variable first ;)

chenba:
😢 I just don't even... well you know what they say, fourth time's a charm.  (Squashed commit, but only change is the ill-placed line.)

PR #3866, opened 2017-11-30
Replace CDN with SITE_CDN and CONTENT_CDN.

### LINKED ISSUE #3590: Prod is not using a CDN
I did a new [perf test](https://www.webpagetest.org/result/171003_SH_38eac1cd8480fe2b2da3b13a2955d883/1/performance_optimization/) (while looking at #3202) and notice that the static resources are not coming off our CDN.

Looking through the code, it seems to be based pretty directly on `$CDN` in the environment. If that is empty, then there won't be any prefix, and it will load the resources locally. @relud says that we're setting `$SITE_CDN` and `$CONTENT_CDN`, which sounds right – there should be two CDN paths. We'll need to look up the changes to figure out what happened here.

chenba:
Fix #3590


6a68:
Not sure about that CI failure with the CDN PR. I created a shot using the dev xpi just now, and the Content-Type looks correct on the shot page: https://screenshots.dev.mozaws.net/eoBGlCxKeTK4ioYo/screenshots.dev.mozaws.net

Rerunning the test

6a68:
looks fixed with a retry :+1:

PR #3868, opened 2017-11-30
Share icon A/B test

6a68:
Show the '<' share icon to 10% of users.

Setting probability to 0.9 seems to be the nicest way to trigger the new icon.


server/src/ab-tests.js
```diff
@@ -7,6 +7,15 @@ below shows what these test objects look like.
 */
 // Note: these get turned into Test objects:
 let allTests = {
+  shotShareIcon: {
+    description: "Use a different share icon",
+    gaField: "cd7",
```
6a68: @ianb Did we decide if a/b tests will each use individual custom dimensions, or if we'll have one custom dimension used for every a/b test?

ianb: We didn't authoritatively decide this. I think after some discussion we came around to just sticking with the way we've been doing it, and allocating new ones. (Note they have to be configured in GA as well.) It should be safe to reuse them after an old test is closed down for a while.

6a68:
Holding this PR till GA is configured. I don't seem to have admin access

Issue #3830, opened 2017-11-22
[Annotations][Enhancement] Add a colour picker 


niharikak101:
Enable users to pick a stroke colour

PR #3880, opened 2017-12-04
Annotations - Add colour picker

niharikak101:
A colo(u)r picker for the pen!


chenba:
This fixes #3830

chenba:
@niharikak101 Is there going to be a future patch where the highlighter also uses this color picker?  If not, then the placement of the color picker is a little confusing.  Moreover, if the highlighter is not going to use the color picker, then I don't think the user should be able to pick a color while the highlighter is selected.

niharikak101:
@chenba good point, I think we should probably extend the highlighter to use the color picker, because the yellow doesn't always show up well against certain colors. @youwenliang will probably have more insight.

server/src/pages/shot/editor.js
```diff
@@ -140,3 +146,54 @@ exports.Editor = class Editor extends React.Component {
...
+          <div className="swatch" style={{backgroundColor: "rgb(142, 68, 173)"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" style={{backgroundColor: "rgb(26, 188, 156)"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" style={{backgroundColor: "rgb(52, 73, 94)"}} onClick={this.onClickSwatch.bind(this)}></div>
+        </div>
+      </div>
+    } else {
+      colorBoard = null;
+    }
+
+    return <div><button className={`color-button`} id="color-picker" onClick={this.onClickColorPicker.bind(this)} title="color-picker" style={{"backgroundColor": this.state.color, "border": `solid ${this.state.color}`}}></button>
```
chenba: The title attr can be "color picker".

server/src/pages/shot/editor.js
```diff
@@ -140,3 +146,54 @@ exports.Editor = class Editor extends React.Component {
...
+        </div>
+        <div className="row">
+          <div className="swatch" style={{backgroundColor: "rgb(46, 204, 113)"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" style={{backgroundColor: "rgb(52, 152, 219)"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" style={{backgroundColor: "rgb(241, 196, 15)"}} onClick={this.onClickSwatch.bind(this)}></div>
+        </div>
+        <div className="row">
+          <div className="swatch" style={{backgroundColor: "rgb(142, 68, 173)"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" style={{backgroundColor: "rgb(26, 188, 156)"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" style={{backgroundColor: "rgb(52, 73, 94)"}} onClick={this.onClickSwatch.bind(this)}></div>
```
chenba: Would be cool if we have sweet names for these pleasant colors!

chenba: Do they need to be `rgb(...)`?

youwenliang:
@niharikak101 Yes the highlighter should also use the color picker but with a transparency.

niharikak101:
@chenba @youwenliang All done!

server/src/pages/shot/editor.js
```diff
@@ -140,3 +146,54 @@ exports.Editor = class Editor extends React.Component {
...
+        </div>
+        <div className="row">
+          <div className="swatch" title="Emerald" style={{backgroundColor: "#2ECC71"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" title="Peter River" style={{backgroundColor: "#3498DB"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" title="Yellow" style={{backgroundColor: "#FF0"}} onClick={this.onClickSwatch.bind(this)}></div>
+        </div>
+        <div className="row">
+          <div className="swatch" title="Wisteria" style={{backgroundColor: "#8E44AD"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" title="Turquoise" style={{backgroundColor: "#1ABC9C"}} onClick={this.onClickSwatch.bind(this)}></div>
+          <div className="swatch" title="Wet Asphalt" style={{backgroundColor: "#34495E"}} onClick={this.onClickSwatch.bind(this)}></div>
```
chenba: Hah, I was kind of kidding about color names, but this is great.  Not sure how localization is going to work for something like this.  Maybe @flodolo can enlighten us.

server/src/pages/shot/editor.js
```diff
@@ -140,3 +146,54 @@ exports.Editor = class Editor extends React.Component {
...
+        </div>
+      </div>
+    } else {
+      colorBoard = null;
+    }
+
+    return <div><button className={`color-button`} id="color-picker" onClick={this.onClickColorPicker.bind(this)} title="Color Picker" style={{"backgroundColor": this.state.color, "border": `solid ${this.state.color}`}}></button>
```
chenba: `color-button` can be a regular string here.

chenba:
Noticed a few things.  Maybe @johngruen and/or @youwenliang will have some input.

The mouse pointer no longer changes based on the selected tool.

When white is selected, the color picker button disappears because the toolbar background color is also white.  Perhaps use a border color that's not white?

The white highlighter acts as an eraser for other highlight colors.  It's not intuitive.

The pen drawings will always be above the highlights regardless of the order in which I use the tools.  E.g. if I draw a few lines in Alizarin with the pen, and then highlight with Peter River, the Alizarin lines are displayed solidly above the highlight.

server/src/pages/shot/editor.js
flodolo: Oh boy, they're going to have fun… (I need to google Wisteria)

server/src/pages/shot/editor.js
```diff
@@ -36,20 +37,24 @@ exports.Editor = class Editor extends React.Component {
...
           <canvas className="image-holder centered" id="image-holder" ref={(image) => { this.imageCanvas = image }} height={ canvasHeight } width={ canvasWidth } style={{height: canvasHeight, width: canvasWidth}}></canvas>
           <canvas className="highlighter centered" id="highlighter" ref={(highlighter) => { this.highlighter = highlighter }} height={canvasHeight} width={canvasWidth}></canvas>
-          <canvas className={"editor centered " + this.state.tool} id="editor" ref={(editor) => { this.editor = editor }} height={canvasHeight} width={canvasWidth}></canvas>
+          <canvas className="editor centered" id="editor" ref={(editor) => { this.editor = editor }} height={canvasHeight} width={canvasWidth}></canvas>
         </div>
       </div>
     </div>
   }
 
+  getColor(color) {
```
ianb: This should be named `onSetColor` or `setColor` – it definitely doesn't get the color!

server/src/pages/shot/editor.js
```diff
@@ -140,3 +146,54 @@ exports.Editor = class Editor extends React.Component {
     this.drawContext.stroke();
   }
 }
+
+class ColorPicker extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      picker: 'inactive',
```
ianb: This should be a boolean, like `pickerActive`

server/src/pages/shot/editor.js
```diff
@@ -140,3 +146,54 @@ exports.Editor = class Editor extends React.Component {
...
+      color: '#000'
+    };
+  }
+
+  render() {
+    let colorBoard;
+    if (this.state.picker == 'active') {
+      colorBoard = <div className="color-board">
+        <div className="row">
```
ianb: This is a big logic chunk... I think this should be done in a later method, and included like `{this.state.pickerActive ? this.renderColorBoard : null}`

server/src/pages/shot/editor.js
niharikak101: @flodolo @chenba We can stick with blues, greens and yellows!

niharikak101:
@chenba I made some updates based on your helpful comments: 

- The mouse pointer changes based on the tool selected
- White has a border colour so that it does not blend in with the background when it is selected
- The white highlighter no longer acts as an eraser for other highlighter tools
- Drawings are displayed in the order in which tools are used

chenba:
@niharikak101 Thanks for those changes.  I just tried it out and noticed that the white highlighter does not have any effect.  That feels like a bug since nothing appears to be happening while I was performing some action with the tool.  But I don't know what the expected effect is with a "white" highlighter.  Reducing opacity?  @johngruen can chime in.

PR #3883, opened 2017-12-04
Do not copy while copying

### LINKED ISSUE #3878: "Whoa! Firefox Screenshots went haywire..." error message is displayed when you rapidly click 3 times the "Copy" button
[Notes]:
- This issue is reproducible only on pages with a large amount of data(eg.:https://www.youtube.com/user/smileyomul, https://imgur.com/, http://edition.cnn.com/), when you select the "Save full page" option.

[Affected versions]:
- Nightly 59.0a1
- Screenshots 25.0.0 

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Open Firefox browser and navigate to a page that has a large amount of data (eg.:https://www.youtube.com/user/smileyomul).
2. Click the "Take a Screenshot" button from the "Page Actions" menu.
3. Click the "Save Full page" button from the right part of the page.
4. Rapidly click 3 times the "Copy" button.

[Expected result]:
- The "Shot Copied" notification appears.

[Actual result]:
- The "Whoa! Firefox Screenshot went haywire..." is displayed ("RefferenceError: uicontrol is not defined) and the elements from the page are selected.

[Additional notes]:
screenshots - uicontrol error

- Here is a screenshot with the error from the Browser console:
screenshots - uicontrol error- browser console

chenba:
Fix #3878

### DIFF
--- a/addon/webextension/selector/shooter.js
+++ b/addon/webextension/selector/shooter.js
@@ -196,12 +196,30 @@ this.shooter = (function() { // eslint-disable-line no-unused-vars
     }));
   };
 
+  let copyInProgress = null;
   exports.copyShot = function(selectedPos) {
+    // This is pretty slow. We'll ignore additional user triggered copy events
+    // while it is in progress.
+    if (copyInProgress) {
+      return;
+    }
+    // A max of five seconds in case some error occurs.
+    copyInProgress = setTimeout(() => {
+      copyInProgress = null;
+    }, 5000);
+
+    let unsetCopyInProgress = () => {
+      if (copyInProgress) {
+        clearTimeout(copyInProgress);
+        copyInProgress = null;
+      }
+    }
     let dataUrl = screenshotPage(selectedPos);
     let blob = blobConverters.dataUrlToBlob(dataUrl);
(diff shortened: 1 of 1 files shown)


addon/webextension/selector/shooter.js
```diff
@@ -196,11 +196,25 @@ this.shooter = (function() { // eslint-disable-line no-unused-vars
...
+      copyInProgress = null;
+    }, 5000);
     let dataUrl = screenshotPage(selectedPos);
     let blob = blobConverters.dataUrlToBlob(dataUrl);
     catcher.watchPromise(callBackground("copyShotToClipboard", blob).then(() => {
       uicontrol.deactivate();
+      if (copyInProgress) {
+        clearTimeout(copyInProgress);
+        copyInProgress = null;
+      }
```
ianb: If this project rejects then it should also clear the timeout (though generally that shouldn't happen). If `Promise.finally` was available that's what should really get used. I don't know if we have a good module to put a finally implementation into (there's one in test.js).

Maybe also add `disabled` to the button while it's in progress? I think we should probably add a different icon while it is in progress, but that applies to all the buttons so we should do that as a separate bug (plus we need an icon).

Issue #3831, opened 2017-11-22
[Annotations][Enhancement] Enable users to crop shots


niharikak101:
Crop tool!

youwenliang:
@niharikak101 Here's a crop tool icon: https://www.dropbox.com/s/hx1jb09202u9gva/crop.svg?dl=0

Adding those tools(crop, color picker, clear) the layout should look something like this:
![annotation-tools](https://user-images.githubusercontent.com/8919308/33158430-dd3cb92e-d044-11e7-943a-9a0290f84e1f.png)

(Fang is on PTO for a week so I'll need to wait for her feedback)

chenba:
This can be closed now. 🎉

PR #3892, opened 2017-12-06
Annotations - Add cropping

niharikak101:
Crop tool for Annotations!

### DIFF
--- a/server/src/pages/shot/editor.js
+++ b/server/src/pages/shot/editor.js
@@ -1,14 +1,31 @@
 const React = require("react");
 const sendEvent = require("../../browser-send-event.js");
+let lineOffset = 4;
+
+let mousedown = false;
+let x1 = -1;
+let x2 = -1;
+let y1 = -1;
+let y2 = -1;
+
+let boxes = [];
 
 exports.Editor = class Editor extends React.Component {
   constructor(props) {
     super(props);
+    this.tempBox = null;
+    this.mouseout = this.mouseout.bind(this);
+    this.mousedown = this.mousedown.bind(this);
+    this.mouseup = this.mouseup.bind(this);
+    this.mousemove = this.mousemove.bind(this);
     this.draw = this.draw.bind(this);
     this.setPosition = this.setPosition.bind(this);
+
+    this.canvasWidth = this.props.clip.image.dimensions.x;
+    this.canvasHeight = this.props.clip.image.dimensions.y;
     this.state = {
       tool: 'pen',
-      color: '#000000',
+      color: '#000',
       size: '5'
     };
   }
@@ -16,16 +33,16 @@ exports.Editor = class Editor extends React.Component {
   render() {
     let penState = this.state.tool == "pen" ? 'active' : 'inactive';
     let highlighterState = this.state.tool == "highlighter" ? 'active' : 'inactive';
-    let canvasHeight = this.props.clip.image.dimensions.y;
-    let canvasWidth = this.props.clip.image.dimensions.x;
+    let toolBar = (<div className="annotation-tools">
+      <button className={`button transparent crop-button`} id="crop" onClick={this.onClickCrop.bind(this)} title="Crop"></button>
+      <button className={`button transparent pen-button ${penState}`} id="pen" onClick={this.onClickPen.bind(this)} title="Pen"></button>
+      <button className={`button transparent highlight-button ${highlighterState}`} id="highlight" onClick={this.onClickHighlight.bind(this)} title="highlighter"></button>
+      <button className={`button transparent clear-button`} id="clear" onClick={this.onClickClear.bind(this)} title="Clear"></button>
+    </div>)
     return <div>
       <div className="editor-header default-color-scheme">
         <div className="shot-main-actions annotation-actions">
-          <div className="annotation-tools">
-            <button className={`button transparent pen-button ${penState}`} id="pen" onClick={this.onClickPen.bind(this)} title="pen"></button>
-            <button className={`button transparent highlight-button ${highlighterState}`} id="highlight" onClick={this.onClickHighlight.bind(this)} title="highlighter"></button>
-            <button className={`button transparent clear-button`} id="clear" onClick={this.onClickClear.bind(this)} title="clear"></button>
-          </div>
+        {this.cropToolBar || toolBar}
         </div>
         <div className="shot-alt-actions annotation-alt-actions">
           <button className="button primary save" id="save" onClick={ this.onClickSave.bind(this) }>Save</button>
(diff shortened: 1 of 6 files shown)


chenba:
This fixes #3831

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+      <button className={`button transparent confirm-crop`} id="confirm-crop" onClick={this.onClickConfirmCrop.bind(this)} title="confirm-crop">Confirm</button>
+      <button className={`button transparent cancel-crop`} id="cancel-crop" onClick={this.onClickCancelCrop.bind(this)} title="cancel-crop">Cancel</button>
+    </div>
+  }
+
+  onClickConfirmCrop() {
+    let x1 = boxes[0].x1;
+    let x2 = boxes[0].x2;
+    let y1 = boxes[0].y1;
+    let y2 = boxes[0].y2;
```
chenba: Nitpick: these names can be confusing since they are not descriptive and the outer scope are using them.

server/src/pages/shot/editor.js
```diff
@@ -1,31 +1,48 @@
...
       size: '5'
     };
   }
 
   render() {
     let penState = this.state.tool == "pen" ? 'active' : 'inactive';
     let highlighterState = this.state.tool == "highlighter" ? 'active' : 'inactive';
-    let canvasHeight = this.props.clip.image.dimensions.y;
-    let canvasWidth = this.props.clip.image.dimensions.x;
+    let toolBar = (<div className="annotation-tools">
```
ianb: I guess this will conflict with or require localization?

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+
+  onClickConfirmCrop() {
+    let x1 = boxes[0].x1;
+    let x2 = boxes[0].x2;
+    let y1 = boxes[0].y1;
+    let y2 = boxes[0].y2;
+    x1 = x1 < 0 ? 0 : x1;
+    y1 = y1 < 0 ? 0 : y1;
+    x2 = x2 > this.canvasWidth ? this.canvasWidth : x2;
+    y2 = y2 > this.canvasHeight ? this.canvasHeight : y2;
```
chenba: `Math.min` and `Math.max` are easier to read and shorter.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+    }
+    this.edit();
+  }
+
+  onClickCrop() {
+    this.setState({tool: 'crop'});
+    this.cropToolBar = <div className="annotation-tools">
+      <button className={`button transparent confirm-crop`} id="confirm-crop" onClick={this.onClickConfirmCrop.bind(this)} title="confirm-crop">Confirm</button>
+      <button className={`button transparent cancel-crop`} id="cancel-crop" onClick={this.onClickCancelCrop.bind(this)} title="cancel-crop">Cancel</button>
+    </div>
```
ianb: missing semicolon

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+    let y2 = boxes[0].y2;
+    x1 = x1 < 0 ? 0 : x1;
+    y1 = y1 < 0 ? 0 : y1;
+    x2 = x2 > this.canvasWidth ? this.canvasWidth : x2;
+    y2 = y2 > this.canvasHeight ? this.canvasHeight : y2;
+    let cropWidth = x2 - x1;
+    let cropHeight = y2 - y1;
+    let croppedImage = document.createElement('canvas');
+    croppedImage.width = parseInt(cropWidth, 10);
+    croppedImage.height = parseInt(cropHeight, 10);
```
ianb: These are already numbers, so you should use `Math.floor()` instead of parseInt.

chenba: I don't think `parseInt`s are needed.  But if they are, then the results should be assigned into `cropWidth` and `cropHeight` since those are assigned to other variables later.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+    let croppedContext = croppedImage.getContext("2d");
+    croppedContext.drawImage(this.imageCanvas, x1, y1, croppedImage.width, croppedImage.height, 0, 0, croppedImage.width, croppedImage.height);
+    croppedContext.drawImage(this.editor, x1, y1, croppedImage.width, croppedImage.height, 0, 0, croppedImage.width, croppedImage.height);
+    croppedContext.globalCompositeOperation = 'multiply';
+    croppedContext.drawImage(this.highlighter, x1, y1, croppedImage.width, croppedImage.height, 0, 0, croppedImage.width, croppedImage.height);
+    this.canvasWidth = cropWidth;
+    this.canvasHeight = cropHeight;
+    let img = new Image();
+    let imageContext = this.imageCanvas.getContext('2d');
+    img.crossOrigin = 'use-credentials';
```
chenba: This shouldn't be necessary as the image will be from a dataurl.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+    let width = cropWidth;
+    let height = cropHeight;
+    img.onload = () => {
+      imageContext.drawImage(img, 0, 0, width, height);
+    }
+    this.imageContext = imageContext;
+    img.src = croppedImage.toDataURL("image/png");
+    this.setState({image: 'cropped'});
```
chenba: I couldn't find where `image` in the component's state is used.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
     this.edit();
```
chenba: This call is not needed since the following line will change state and that will lead to `this.edit()` being called.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
 
+  onClickCancelCrop() {
+    this.cropToolBar = null;
+    this.cropper.getContext('2d').clearRect(0, 0, this.cropper.width, this.cropper.height);
+    boxes = [];
+    this.setState({tool: 'pen'});
+  }
+
+  mouseup(e) {
+    if (this.clickedArea.box == -1 && this.tempBox != null && boxes.length < 1) {
```
chenba: `this.clickedArea.box` is either `-1` or `0` and act as an index into `boxes`; again, I don't think `boxes` needs to be an array.  The only element in `boxes` is `this.tempBox`, so we can simply use one variable for the crop box.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+      boxes.push(this.tempBox);
+    } else if (this.clickedArea.box != -1) {
+      let selectedBox = boxes[this.clickedArea.box];
+      if (selectedBox.x1 > selectedBox.x2) {
+        let previousX1 = selectedBox.x1;
+        selectedBox.x1 = selectedBox.x2;
+        selectedBox.x2 = previousX1;
```
chenba: You could simply do `[selectedBox.x1, selectedBox.x2] = [selectedBox.x2, selectedBox.x1]`;

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+      }
+      if (selectedBox.y1 > selectedBox.y2) {
+        let previousY1 = selectedBox.y1;
+        selectedBox.y1 = selectedBox.y2;
+        selectedBox.y2 = previousY1;
+      }
+    }
+
+    this.clickedArea = {box: -1, pos: 'o'};
```
chenba: I couldn't find how the pos of 'o' is used.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+        let previousX1 = selectedBox.x1;
+        selectedBox.x1 = selectedBox.x2;
+        selectedBox.x2 > previousX1;
+      }
+      if (selectedBox.y1 > selectedBox.y2) {
+        let previousY1 = selectedBox.y1;
+        selectedBox.y1 = selectedBox.y2;
+        selectedBox.y2 > previousY1;
+    }
+  }
```
chenba: Indentation is off here.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+    mousedown = false;
+    this.clickedArea = {box: -1, pos: 'o'};
+    this.tempBox = null;
+  }
+
+  findCurrentArea(x, y) {
+    let box = boxes[0];
+    if (box) {
+      let xCenter = box.x1 + (box.x2 - box.x1) / 2;
+      let yCenter = box.y1 + (box.y2 - box.y1) / 2;
```
chenba: The results of the calculations below are re-used, so it'd be easier to read if we assign and name them.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+      imageContext.drawImage(img, 0, 0, width, height);
+    }
+    this.imageContext = imageContext;
+    img.src = url || this.props.clip.image.url;
+  }
+
+
   onClickClear() {
     this.setState({tool: this.state.tool});
+    this.setState({image: 'cleared'});
```
chenba: This is not necessary.

server/src/pages/shot/editor.js
```diff
@@ -110,6 +374,14 @@ exports.Editor = class Editor extends React.Component {
       this.canvasContainer.removeEventListener("mousemove", this.draw);
```
chenba: I know this is not new code, but I don't understand how `this.state.tool == 'none'` is ever possible.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+      }
+      this.redraw();
+    }
+  }
+
+  redraw() {
+    let boxX1 = x1 < x2 ? x1 : x2;
+    let boxY1 = y1 < y2 ? y1 : y2;
+    let boxX2 = x1 > x2 ? x1 : x2;
+    let boxY2 = y1 > y2 ? y1 : y2;
```
chenba: `Math.min` and `Math.max`?

ianb:
There's a general UX issue here around selection, and confirm/cancel save/cancel:

* The selection works differently than Screenshots itself works. Unfortunately we can't just share code between the two, but we should use the same styles. I'd be okay following up with that.
* In Screenshots selection we use a lightbox effect instead of a translucent overlay. That's more conventional (and looks better).
* The resize targets in Screenshots selection are actually largish areas, with the handles as just visual hints.
* Clicking the background doesn't cancel the selection

Anyway, that could be fixed up later, it's not a blocker for me. BUT: I am uncomfortable with the overall flow here.

When you start cropping you get this:

image

This is very confusing! What does Confirm mean? Which Cancel means what? What tool is active? The second time I tested it I clicked crop, selected a region, clicked Save, and was certain that there was a bug because my crop was lost.

Arguably crop doesn't have to be applied at all, except lazily on Save. That is, hitting the Crop button would let you drag out a region, which would be lightboxed. There would be some "remove crop" button (which should be explicit, so you can distinguish it from canceling edit), but no confirm button. If you select another tool, the crop just stays there, and you draw on the entire image as normal. I think it may even be feasible to make it resizable while another tool is activated (though that's going to get into trouble if the resize targets are too large, and certainly there'd be no move).

Maybe a simple enough change for now would simply be to make sure Save effectively confirms any active crop. I'd also rename the Confirm button as simply Done or Crop. I think Cancel needs to be "Cancel crop" or maybe "Undo": two "Cancel" buttons is too many.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+    let cropWidth = x2 - x1;
+    let cropHeight = y2 - y1;
+    let croppedImage = document.createElement('canvas');
+    croppedImage.width = parseInt(cropWidth, 10);
+    croppedImage.height = parseInt(cropHeight, 10);
+    let croppedContext = croppedImage.getContext("2d");
+    croppedContext.drawImage(this.imageCanvas, x1, y1, croppedImage.width, croppedImage.height, 0, 0, croppedImage.width, croppedImage.height);
+    croppedContext.drawImage(this.editor, x1, y1, croppedImage.width, croppedImage.height, 0, 0, croppedImage.width, croppedImage.height);
+    croppedContext.globalCompositeOperation = 'multiply';
+    croppedContext.drawImage(this.highlighter, x1, y1, croppedImage.width, croppedImage.height, 0, 0, croppedImage.width, croppedImage.height);
```
ianb: It looks like this flattens the image when cropping. Is there a reason to do this? Couldn't all the canvases be flattened at the same time?

I'm a little unclear why the highlight isn't applied twice, as I don't see `this.highlighter` cleared out. Though in experimentation it seems to work fine.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+    this.canvasWidth = cropWidth;
+    this.canvasHeight = cropHeight;
+    let img = new Image();
+    let imageContext = this.imageCanvas.getContext('2d');
+    img.crossOrigin = 'use-credentials';
+    let width = cropWidth;
+    let height = cropHeight;
+    img.onload = () => {
+      imageContext.drawImage(img, 0, 0, width, height);
+    }
```
ianb: I'd like to see the resizing all factored out into a stand-alone function that returns a promise. The resizing is complicated code, so keeping it separated from the editing would be nice.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+  mousedown(e) {
+    mousedown = true;
+    this.clickedArea = this.findCurrentArea(e.offsetX, e.offsetY);
+    x1 = e.offsetX;
+    y1 = e.offsetY;
+    x2 = e.offsetX;
+    y2 = e.offsetY;
+  }
+
+  mouseout(e) {
```
ianb: Does there need to be a `mouseout`? I guess this catches when you drag outside of the canvas. But really that should be fine, it should just get truncated to the canvas size. Right now it's actually not easy to crop to the edge (e.g., if I want to move the selection in on only a couple sides), because the selection will typically get stopped a couple pixels from the edge.

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+        selectedBox.y1 = selectedBox.y2;
+        selectedBox.y2 > previousY1;
+    }
+  }
+    mousedown = false;
+    this.clickedArea = {box: -1, pos: 'o'};
+    this.tempBox = null;
+  }
+
+  findCurrentArea(x, y) {
```
ianb: There's a pattern in `ui.js` and `uicontrol.js` for this. E.g.: https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/selector/uicontrol.js#L104-L114

and: https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/selector/uicontrol.js#L800-L823

This method is a bit like `ui.Box.draggerDirection` but it's got a bit too much going on. Also `draggerDirection` doesn't do math on the pixel location, but uses the fact that the box has target elements: https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/selector/ui.js#L783-L799

server/src/pages/shot/editor.js
```diff
@@ -34,23 +51,278 @@ exports.Editor = class Editor extends React.Component {
...
+      if (boxX2 - boxX1 > lineOffset * 2 && boxY2 - boxY1 > lineOffset * 2) {
+        this.tempBox = {x1: boxX1, y1: boxY1, x2: boxX2, y2: boxY2}
+      }
+      if (this.tempBox) {
+        this.drawCropBox(this.tempBox, context);
+      }
+    }
+  }
+
+  drawCropBox(box, context) {
```
ianb: I don't think this needs to happen on the canvas at all, it can be an element on top of the canvas. This makes it easier to style and control.

niharikak101:
@ianb and @chenba - your comments were incredibly useful and therefore I decided to give this an overhaul.
I've stuck to the how the Screenshots selection works - in terms of the UI and the actual code.  The overall flow works something like this:
Clicking the crop button activates the crop toolbar. The extra save and cancel buttons are no longer visible and the user can choose to either confirm or cancel the crop. Confirming the crop puts the user in an intermediate state where they can still go back to the original image by clearing their edits (clicking on the clear button). This crop is only confirmed by saving. This is similar to the original UI spec provided by @youwenlang.

chenba:
@niharikak101 I haven't looked through the code yet, but from trying it out I noticed the following:

- The selection rectangle is not anchored.  I cannot start a selection by dragging left or up.  And I can keep pushing the selection left and up until it's out of view (without releasing the initial mousedown to start the selection).  And, if I keep dragging left and back down, the selection rectangle enters from the right.  Confirming that selection gives me a blank image.
- Similarly, once I have a selection, if I drag a border towards its opposite border, when I reached the opposite border, I'd start pushing the selection farther in that direction.
- Once I've made a selection, if I click and drag outside of the selection, it immediately switches to the Pen tool and draws a line from the top left corner of the shot to where the mousedown occurred.  The line continues with the dragging until the mouse button is released.

I like that the selection style now matches the addon!

server/src/pages/shot/editor.js
```diff
@@ -61,10 +95,239 @@ exports.Editor = class Editor extends React.Component {
...
+    }
+    this.cropBox = null;
+    selectedPos = {};
+    movingPos = {};
+    resizeDirection = null;
+    resizing = false;
+    creating = false;
+  }
+
+  displayCropBox(pos, direction) {
```
chenba: `direction` is not used in this function.

server/src/pages/shot/editor.js
```diff
@@ -61,10 +95,239 @@ exports.Editor = class Editor extends React.Component {
...
+    this.bgLeft.style.left = "0px";
+    this.bgLeft.style.width = pos.left + "px";
+    this.bgRight.style.top = pos.top + "px";
+    this.bgRight.style.height = pos.bottom - pos.top + "px";
+    this.bgRight.style.left = pos.right + "px";
+    this.bgRight.style.width = "100%";
+  }
+
+  createCropBox() {
+    let movements = ["topLeft", "top", "topRight", "left", "right", "bottomLeft", "bottom", "bottomRight"];
```
chenba: This line is identical to the one in `findClickedArea`, and both have the same name as another variable in the outer scope.  Perhaps rename it and move it to the outer scope.

niharikak101:
@chenba the changes take care of thumbnail regeneration; the selection box is now anchored, and clicking outside the selection removes the selection box, but the crop tool remains active.

chenba:
If I start the selection from somewhere within the right 1/3 (?) portion of the image, I'd have a selection box that will not allow its left and right borders meet and switch sides.  Easiest way to see this is to start the selection by dragging left and up from the lower-right of the image.

server/src/pages/shot/editor.js
```diff
@@ -2,53 +2,168 @@ const React = require("react");
...
+      let tmp = this.x2;
+      this.x2 = this.x1;
+      this.x1 = tmp;
+    }
+    if (this.y1 > this.y2) {
+      let tmp = this.y2;
+      this.y2 = this.y1;
+      this.y1 = tmp;
+    }
+  }
```
chenba: If you call `sortCoords()` at the end of the constructor and at the beginning of the setters, then we won't need the `Math.min()` and `Math.max()` calls on the getters, and the setters won't need their own `if`s.

server/src/pages/shot/editor.js
```diff
@@ -61,10 +176,246 @@ exports.Editor = class Editor extends React.Component {
...
+        moveY = moveY == "*" ? ["y1", "y2"] : [moveY];
+        for (let moveDir of moveY) {
+          selectedPos[moveDir] = this.truncateY(resizeStartSelected[moveDir] + diffY);
+        }
+      }
+    this.displayCropBox(selectedPos);
+  }
+
+  truncateX(x) {
+    let max = Math.max(this.canvasWidth);
```
chenba: Don't need the `Math.max()` since there's only number here.

server/src/pages/shot/editor.js
```diff
@@ -61,10 +176,246 @@ exports.Editor = class Editor extends React.Component {
...
+    if (x < 0) {
+      return 0;
+    } else if (x > max) {
+      return max;
+    }
+    return x;
+  }
+
+  truncateY(y) {
+    let max = Math.max(this.canvasHeight);
```
chenba: Don't need this `Math.max()` either.

server/src/pages/shot/editor.js
```diff
@@ -61,10 +176,246 @@ exports.Editor = class Editor extends React.Component {
...
+  resizeCropBox(event, direction) {
+    let rect = this.cropContainer.getBoundingClientRect();
+    let diffX = event.clientX - rect.left - resizeStartPos.x;
+    let diffY = event.clientY - rect.top - resizeStartPos.y;
+    let movement = movementPositions[resizeDirection];
+    if (movement[0]) {
+        let moveX = movement[0];
+        moveX = moveX == "*" ? ["x1", "x2"] : [moveX];
+        for (let moveDir of moveX) {
+          selectedPos[moveDir] = this.truncateX(resizeStartSelected[moveDir] + diffX);
```
chenba: This should use the `left` and `right` setters in `selectedPos`.

server/src/pages/shot/editor.js
```diff
@@ -61,10 +176,246 @@ exports.Editor = class Editor extends React.Component {
...
+        }
+      }
+      if (movement[1]) {
+        let moveY = movement[1];
+        moveY = moveY == "*" ? ["y1", "y2"] : [moveY];
+        for (let moveDir of moveY) {
+          selectedPos[moveDir] = this.truncateY(resizeStartSelected[moveDir] + diffY);
```
chenba: This should use setters as well.

server/src/pages/shot/editor.js
niharikak101: This builds on the code for the Screenshots selection box - the Coords are sorted based on the values of `x1`, `x2` and `y1`, `y2`. Using getters and setters breaks resizing.

niharikak101:
The last commit adds a fix for #3974

ianb:
With the default drawing tool I managed to get it into a buggy state with a little scribbling. It would draw for just a short while (20ish pixels) and then stop, with error messages `TypeError: this.imageCanvas is null` then `TypeError: null has no properties`. But then I tried repeating it with an unminified build and was able to, but without any errors. This error persisted when I changed tools. But this probably isn't related to cropping. Here's a little odd thing that does happen on the first resize after making a selection:

image

Not a blocker, but maybe a followup. Otherwise functionality-wise it's working really great! (I still have to look at the code...)

chenba:
I get `TypeError: this.imageCanvas is null` whenever I start the edit tool, cancel, and start editing again.  Saving the image will remove the error for the next edit button click.

PR #3904, opened 2017-12-08
Fix #3894, implement A/B test for text-free Download button on shot page

### LINKED ISSUE #3894: A/B test textless download button on shot page
Eventually we are going to want to start rearranging buttons in the add-on. In anticipation of this, I want to see how much effect removing the text cue from a button will have.

Let's conduct a test where we hide the word `download` from the download button on shot pages for some users and measure the relative click thru rate.

Current:
image

Test Branch:
image

I would like to measure whether taking out the text depresses clicks on this button or if it does not matter.

One gotcha to note: the diff here includes removing an `&nbsp;` hiding at the end of this line:
https://github.com/mozilla-services/screenshots/blob/master/server/src/pages/shot/view.js#L415

6a68:


server/src/pages/shot/view.js
```diff
@@ -386,8 +386,13 @@ class Body extends React.Component {
...
 
-    const shotPageDownload = <Localized id="shotPageDownload"><span className="download-text">Download</span></Localized>;
-
+    const noText = this.props.abTests && this.props.abTests.downloadText
+                           && this.props.abTests.downloadText.value === "no-download-text";
+    const downloadText = [<img src={ this.props.staticLink("/static/img/download-white.svg") } width="20" height="20"/>,
+                         '\u00a0',
+                          <Localized id="shotPageDownload"><span className="download-text">Download</span></Localized>];
+    const downloadNoText = [<img src={ this.props.staticLink("/static/img/download-white.svg") } width="20" height="20"/>,
+                    <span className="download-text"></span>];
```
chenba: eslint is not happy with these arrays.  Do they need to be arrays?

6a68: yeah, I got slightly obsessed with finding a way to wedge a non-breaking space into jsx. I'll just put some right-margin on the img instead

PR #3920, opened 2017-12-19
Use fixed height for thumbnail container

chenba:
Fix #3950


chenba:
With this patch, some of the generated thumbnails will not be the _exact_ size and aspect ratio at which the CSS will display them.  E.g. a thumbnail at 210 x 90 can be 135 x 90 after this change.  We are leaving a few kb on the table here and there, but it does allow the thumbnail generation code to be simpler.  (Also, we could tinker more with the thresholds for generating a thumbnail if needed; currently it is very simple.)

PR #3926, opened 2017-12-27
Skip test_file_managemnt tests by default

### LINKED ISSUE #3421: Turn off test_file_management.py from running by default
This test is really slow, it's probably best that it not run when `npm run test:server` is run

pradeepgangwar:
Add a marker for slow tests. And skip test_file_managements.py with that marker.

Closes #3421


ianb:
Oh cool, I figured py.test must have a way to do this, but couldn't figure it out from the docs. Thanks!

PR #3941, opened 2018-01-04
Use an object url for the preview image.

### LINKED ISSUE #3940: Use Object URL for Preview Image
Related to bug 1424870

chenba:
Fix #3940

### DIFF
--- a/addon/webextension/selector/ui.js
+++ b/addon/webextension/selector/ui.js
@@ -1,4 +1,4 @@
-/* globals log, util, catcher, inlineSelectionCss, callBackground, assertIsTrusted, assertIsBlankDocument, buildSettings */
+/* globals log, util, catcher, inlineSelectionCss, callBackground, assertIsTrusted, assertIsBlankDocument, buildSettings blobConverters */
 
 "use strict";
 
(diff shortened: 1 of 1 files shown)


addon/webextension/background/main.js
```diff
@@ -325,5 +325,11 @@ this.main = (function() {
     }));
   });
 
+  // Used for the preview images. Can be removed once #3508
+  // (https://git.io/v5Nkc) is done.
+  communication.register("dataUrlToBlob", (sender, url) => {
```
ianb: Let's just include blobConverters in content worker and call this function directly instead of using the background worker

addon/webextension/selector/ui.js
```diff
@@ -1,4 +1,4 @@
-/* globals log, util, catcher, inlineSelectionCss, callBackground, assertIsTrusted, assertIsBlankDocument, buildSettings */
+/* globals log, util, catcher, inlineSelectionCss, callBackground, assertIsTrusted, assertIsBlankDocument, buildSettings blobConverters */
```
chenba: @ianb You were right about `blobConverters` being in the content worker already. 😳

PR #3957, opened 2018-01-10
Do not send events while in private browsing.

### LINKED ISSUE #3900: Confirm disabled event collection while in Private Browsing
While we don't allow upload, and so we never login to the website and have data collection there, I think we are still collecting events for Private Browsing sessions.

chenba:
Fix #3900 

Note that user timings are still sent.

### DIFF
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -94,6 +94,15 @@ this.analytics = (function() {
       label = undefined;
     }
     options = options || {};
+
+    // Don't send events if in private browsing.
+    if (options.incognito) {
+      return Promise.resolve();
+    }
+
+    // Don't include in event data.
+    delete options.incognito;
+
     let di = deviceInfo();
     options.applicationName = di.appName;
     options.applicationVersion = di.addonVersion;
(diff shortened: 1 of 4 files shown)


addon/webextension/background/analytics.js
```diff
@@ -94,6 +94,17 @@ this.analytics = (function() {
       label = undefined;
     }
     options = options || {};
+
+    // Don't send events if in private browsing.
+    if (options.hasOwnProperty('incognito')) {
```
ianb: What's the reason for `hasOwnProperty` here?

addon/webextension/selector/shooter.js
```diff
@@ -221,6 +221,15 @@ this.shooter = (function() { // eslint-disable-line no-unused-vars
   };
 
   exports.sendEvent = function(...args) {
+    let maybeOptions = args[args.length - 1];
+
+    if (typeof maybeOptions === "object") {
+      maybeOptions.incognito = browser.extension.inIncognitoContext;
+      args[args.length - 1] = maybeOptions;
```
ianb: This line isn't necessary – you're modifying the object in-place, so it'll stay in the arguments in its modified form.

addon/webextension/background/analytics.js
ianb: You could do `options.incognito !== undefined` – but I think simply testing for truthiness would do what's needed here?

chenba: If it's falsey because it has a value of `false`, then `delete` below won't happen.  I suppose I could call the `delete` without a check since that won't throw an error.

PR #3958, opened 2018-01-10
Check if thumbnail exist before use.

chenba:
Because that attribute does not exist on old versions of the shot model.


chenba:
Closing this because it's not a bug.

PR #3965, opened 2018-01-11
Add never expire indicator to My Shots

### LINKED ISSUE #3500: Add an indicator that a shots have been saved indefinitely
In order to highlight indefinitely saved shots, we should add some kind of indicator to to _My Shots_ on each indefinitely saved shot. This could be a little badge or something. Seems like low hanging fruit and an easy way to call out favorites.

As a follow on, we might consider moving indefinitely saved shots to the top of _My Shots_.

chenba:
Fix #3500

### DIFF
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -136,7 +136,10 @@ shotPageDMCAIncludeLink = Please include the URL of this shot in your email: { $
 shotPageKeepFor = How long should this shot be retained?
 // Note: shotPageSelectTime is a placeholder label for the time selection dropdown.
 shotPageSelectTime = Select time
-shotPageKeepIndefinitely = Indefinitely
+// The ∞ is used to indicate that the shot won't expire. It is also used in
+// shotIndexNoExpirationSymbol. Try to use the same symbol in both strings, or
+// if no such symbol is available for a language/culture, simply leave it out.
+shotPageKeepIndefinitely = Indefinitely ∞
 shotPageKeepTenMinutes = 10 Minutes
 shotPageKeepOneHour = 1 Hour
 shotPageKeepOneDay = 1 Day
@@ -217,7 +220,12 @@ shotIndexPagePreviousPage
     .title = Previous page
 shotIndexPageNextPage
     .title = Next page
-
+// This symbol is used in the lower right corner of the card for a shot on the
+// My Shots page to indicate that the shot does not expire. It should be a
+// single character (or simply nothing if no such symbol is available for a
+// language/culture).
+shotIndexNoExpirationSymbol = ∞
+    .title = This shot does not expire
 
 // all metrics strings are optional for translation
 [[ Metrics page ]]
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -549,7 +549,7 @@ class ExpireWidget extends React.Component {
         <Localized id="shotPageKeepFor"><span>How long should this shot be retained?</span></Localized>
         <select ref={expireTime => this.expireTime = expireTime}>
           <Localized id="shotPageSelectTime"><option value="cancel">Select time</option></Localized>
-          <Localized id="shotPageKeepIndefinitely"><option value="0">Indefinitely</option></Localized>
+          <Localized id="shotPageKeepIndefinitely"><option value="0">Indefinitely &infin;</option></Localized>
           <Localized id="shotPageKeepTenMinutes"><option value={ 10 * minute }>10 Minutes</option></Localized>
           <Localized id="shotPageKeepOneHour"><option value={ hour }>1 Hour</option></Localized>
           <Localized id="shotPageKeepOneDay"><option value={ day }>1 Day</option></Localized>
--- a/server/src/pages/shotindex/controller.js
+++ b/server/src/pages/shotindex/controller.js
@@ -17,7 +17,11 @@ const queryParamModelPropertyMap = {
 exports.launch = function(m) {
   if (m.hasDeviceId) {
     if (m.shots) {
-      m.shots = m.shots.map((shot) => new AbstractShot(m.backend, shot.id, shot.json));
+      m.shots = m.shots.map((shot) => {
+        let s = new AbstractShot(m.backend, shot.id, shot.json);
+        s.expireTime = shot.expireTime;
+        return s;
+      });
     }
     Object.assign(m, extractQueryParamValues(queryParamModelPropertyMap));
     model = m;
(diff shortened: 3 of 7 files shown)


locales/en-US/server.ftl
```diff
@@ -217,7 +218,8 @@ shotIndexPagePreviousPage
     .title = Previous page
 shotIndexPageNextPage
     .title = Next page
-
+shotIndexNoExpirationSymbol = ∞
```
flodolo: Can you add some context in a comment on where this symbol is used? In particular, make it clear that there is not space available for more text, if necessary it should be replaced by another single character like `-`.

locales/en-US/server.ftl
```diff
@@ -136,7 +136,8 @@ shotPageDMCAIncludeLink = Please include the URL of this shot in your email: { $
 shotPageKeepFor = How long should this shot be retained?
 // Note: shotPageSelectTime is a placeholder label for the time selection dropdown.
 shotPageSelectTime = Select time
-shotPageKeepIndefinitely = Indefinitely
+// The ∞ is also used in shotIndexNoExpirationSymbol.  Please ensure they match.
```
flodolo: Do they really need to match?

locales/en-US/server.ftl
chenba: Good point.  Will do.

locales/en-US/server.ftl
chenba: If I understand the UI/UX folks' idea correctly, we want that symbol in the expiration selection and My Shots, so the users can draw the connection between them.

flodolo: True. But, if for any reason the symbol doesn't work for a language/culture, this might not be possible.

Maybe it's enough to change "Please ensure" with "Try to use the same symbol in both strings"?

chenba: Oh I figured that if there isn't an appropriate symbol for a language/culture, we can leave it out altogether.  I'll mention that and use your wording as well.

chenba:
@flodolo Added more comments.  Please take another look.

locales/en-US/server.ftl
```diff
@@ -217,7 +220,12 @@ shotIndexPagePreviousPage
     .title = Previous page
 shotIndexPageNextPage
     .title = Next page
-
+// This symbol is used in the lower right corner of the card for a shot on the
+// My Shots page to indicate that the shot does not expire. It should be a
+// single character (or simply nothing if no such symbol is available for a
+// languge/culture).
```
flodolo: typo: language

locales/en-US/server.ftl
```diff
@@ -217,7 +220,12 @@ shotIndexPagePreviousPage
     .title = Previous page
 shotIndexPageNextPage
     .title = Next page
-
+// This symbol is used in the lower right corner of the card for a shot on the
+// My Shots page to indicate that the shot does not expire. It should be a
+// single character (or simply nothing if no such symbol is available for a
```
flodolo: I wonder if we risk breaking something without a value. Maybe suggest `or a character like - if…`

chenba: Nothing should break if we don't have a value.  That's why I've left out the default content at https://github.com/mozilla-services/screenshots/pull/3965/files#diff-d45944c551eead471a128c9a716e2a79R309

flodolo: OK, then feel free to keep the existing comment

ianb:
@chenba Can you put a screenshot on this for @ChenMorpheus or @johngruen or someone on the design side to look at?

chenba:
@johngruen, @ChenMorpheus:

image

chenba:
Note that screenshot is from a minimal Linux set up.  The font used on that infinity character is arial (`font-family` inherited from `<body>` and none of the other fonts are installed on that OS).

fangshih:
@chenba Thanks for the screenshot! I've discussed with UXs, we don't need to show the indicator when we hover the shot. So we should keep the hover buttons in the same position here. 

Also, Is that possible to use this gray color #9E9E9E for the indicator? And would be great if we can make it slightly bigger with 8px height to see how it looks. Thanks! Thanks!

Oh! Can we also have a screenshot of the dropdown menu? We'd like to see how the indicator looks in the dropdown menu. Thanks!

chenba:
@fangshih I've updated the hover style and the font color.  I did not quite understand the 8px height request, but I've included a screenshot with multiple font sizes.  I also made the domain names longer so you can see where it could collide with the symbol; I can move the symbol down and right more to compensate.

image

The `<option>` text:

image

chenba:
D'oh!  I should've added a `[WIP]` to the title of this PR.  :no_mouth:  @fangshih Please leave feedback based on the two images above here.  I will created another PR to follow up once you do.  Thanks!

ianb:
Whoops, sorry!

fangshih:
@chenba  Thanks for the previews! I think I'll go for the ∞ Semibold 18pt in system font. 
For the longer domain name, I've attached a screenshot. We may need to make the name no longer than this size. I'd like to keep some space between ∞ and the domain name. 
And for the dropdown part, the icon seems a little different compared to the MyShot one? Just want to make sure it is using the same font same style like Myshot ∞, and it will be good to go. And thank you so much for all the updates! 
 
example-icon

locales/en-US/server.ftl
```diff
@@ -136,7 +136,10 @@ shotPageDMCAIncludeLink = Please include the URL of this shot in your email: { $
 shotPageKeepFor = How long should this shot be retained?
 // Note: shotPageSelectTime is a placeholder label for the time selection dropdown.
 shotPageSelectTime = Select time
-shotPageKeepIndefinitely = Indefinitely
+// The ∞ is used to indicate that the shot won't expire. It is also used in
+// shotIndexNoExpirationSymbol. Try to use the same symbol in both strings, or
+// if no such symbol is available for a language/culture, simply leave it out.
+shotPageKeepIndefinitely = Indefinitely ∞
```
flodolo: Ugh, I completely missed that you changed an existing string. Was it like this from the beginning? All locales won't pick up the change without a new string ID

chenba: It was just `Indefinitely` before the patch.

flodolo: Yes, what I (likely) missed is that the string changed but the ID didn't
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings

Unfortunately, we need a patch that changes the ID, to force all locales to look at that string and retranslate it.

PR #3975, opened 2018-01-17
Update style for indefinitely saved symbol.

### LINKED ISSUE #3500: Add an indicator that a shots have been saved indefinitely

chenba:
Fix #3500 some more

### DIFF
--- a/static/css/shot-index.scss
+++ b/static/css/shot-index.scss
@@ -155,6 +155,9 @@ h1 {
     .alt-actions-container {
       opacity: 1;
     }
+    .never-expires {
+      display: none;
+    }
   }
 
   .shot-image-container {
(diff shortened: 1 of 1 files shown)


chenba:
@fangshih This patch contains the style changes for the ∞ on My Shots.  Since `<option>` do not allow HTML tags, we cannot style the symbol in the select separately.  (We could switch to some other markup, but that'd be a bigger change and should have its own issue.)

PR #3992, opened 2018-01-19
Change l10n string id.

chenba:
Needed for #3500


flodolo:
Thanks for fixing the ID.

PR #4006, opened 2018-01-24
Disable uploads for ESR or when upload-disabled pref is set

### LINKED ISSUE #3996: Disable uploading on ESR
Firefox 60 is an ESR.  To avoid problems down the line, we'd like to disable the ability to save screenshots to our servers on ESR releases (note: just ESR, not plain Firefox 60).  The screenshots functionality will still be there, but the Save button will be omitted.

From our discussion:
- We should add a pref to do this and convert our current disable-save-button logic to use it
- Add ESR detection to flip the pref
- Add the pref to our documentation and/or SUMO

6a68:
Fixes #3996, and includes some refactoring as well.

Might be more readable commit-by-commit, rather than as unified diff.

### DIFF
--- a/addon/webextension/background/analytics.js
+++ b/addon/webextension/background/analytics.js
@@ -6,7 +6,7 @@ this.analytics = (function() {
   let exports = {};
 
   let telemetryPrefKnown = false;
-  let telemetryPref;
+  let telemetryEnabled;
 
   const EVENT_BATCH_DURATION = 1000; // ms for setTimeout
   let pendingEvents = [];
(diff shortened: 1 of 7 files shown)


6a68:
I've filed [bug 1432694](https://bugzil.la/1432694) to add a default value for this pref to about:config.

addon/webextension/background/analytics.js
```diff
@@ -129,24 +129,24 @@ this.analytics = (function() {
...
+      telemetryEnabled = false;
       throw error;
     });
   };
 
-  exports.getTelemetryPrefSync = function() {
+  exports.isTelemetryEnabled = function() {
     catcher.watchPromise(exports.refreshTelemetryPref());
-    return !!telemetryPref;
+    return !!telemetryEnabled;
```
chenba: Nit (and not even part of these changes): if `telemetryEnabled` defaults to false, then we don't need the `!!` here.

alexhass:
Why is this ESR only per title? I need this in RR, too.

PR #4065, opened 2018-02-06
Upstream changes

6a68:
Two refactorings in mozilla-central affected our code. See the individual commit messages for details.

One of the refactorings touched raven.js. I'm not sure if eslint can auto-rewrite that particular line or not, but it'll be something to watch for in future exports.


ianb:
Someday (per manifest discussions in the GitHub group) we should be able to mark Raven.js as external and not editable.

PR #4075, opened 2018-02-08
Change Xvfb display number.

chenba:
Because root on CircleCI is a jerk.


chenba:
Merging this since it's worked multiple times now. (Failure now from linting issues not Selenium.)

PR #4083, opened 2018-02-09
Update to fluent 0.4.3

stasm:
Please update the `fluent` dependency to 0.4.3.  It has important fixes to the FTL parser.

While the main development of `fluent` continues with the recent 0.6.x versions, I also published 0.4.3 which is API-compatible with other 0.4.x versions. The upgrade should be painless.

`fluent` 0.4.3 supports the recent changes to the Fluent Syntax. It also supports the syntax Screenshots and Pontoon use right now. We expect Pontoon to upgrade their Fluent dependencies early next week, at which point all translations saved back to GitHub will be serialized using the new syntax. This PR will minimize the risk of regressions from that update (which is already low because the changes to the syntax are small).

Some of the syntax changes will require an update to the en-US reference file in this repo, too. We'll open a new PR for that once Pontoon is capable of reading it.

cc @flodolo and @zbraniecki who will be around next week to assist in case of questions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "envc": "2.5.0",
     "escape-html": "1.0.3",
     "express": "4.16.2",
-    "fluent": "0.4.1",
+    "fluent": "0.4.3",
     "fluent-intl-polyfill": "0.1.0",
     "fluent-langneg": "0.1.0",
     "fluent-react": "0.4.1",


stasm:
Unfortunately I wasn't able to test this locally. I ran into trouble with setting up postgres on Windows.

flodolo:
I tested locally on macOS, switching a few languages (it, pl, ach as an incomplete one), and everything seems to work normally.

flodolo:
Flagging a couple of potential reviewers to make sure this doesn't fall off the radar. It would be great if someone can take a look at this early next week.

P.S. also retrigged the CircleCI build, hoping the failure will go away

flodolo:
To give you the full picture and help prioritizing this PR: we're starting to use Fluent in Firefox, and we're waiting for the 0.6.x patch to be uplifted to mozilla-beta. We hope that will happen Today or Tomorrow.

Once that happens, we need to deploy a new version of Pontoon to unblock other pending work for Firefox. Unfortunately, that means that any update done by localizers will serialize a file with the new syntax, potentially causing issues in projects using an old version of Fluent (like Screenshots).

In order to avoid that, before deploying the new version of Pontoon, we will disable sync (read/write to the repository) for this project, and re-enable it as soon as the new version of Fluent is in use (i.e. this PR is merged).

flodolo:
Thanks Ian!

PR #4106, opened 2018-02-13
Fix #3771 - Check for S3 errors when accessing bucket

### LINKED ISSUE #3771: Catch NoSuchKey from AWS
Example in [this report](https://sentry.prod.mozaws.net/operations/pageshot-prod/issues/628348/)

The traceback is not very useful. I'm pretty sure it's some case where the S3 file is gone, but the database entry still exists. This happens a lot! That might be a indication of a problem itself, we should confirm with @relud if there's any cases where the S3 image can get automatically deleted.

The error should also bubble up to the code that is trying to do a get, and looking at the code it appears to do so: https://github.com/mozilla-services/screenshots/blob/0b4f49aa6888bc15e46efd2e4442a886401ef283/server/src/servershot.js#L115-L129

But it doesn't?

6a68:
Re: description of #3771, I'm not sure this error check is worth merging.

I looked at where the `get()` function is used, and didn't see a spot where a promise rejection might get dropped. Noticed that we don't check errors in the outer `createBucket` call, so, _maybe_ that's relevant? Our usage of the aws-sdk API diverges from the docs on the aws site (we don't pass params into the `createBucket` call, but into the constructor), so maybe there's some other undocumented behavior where errors get handled in the inner callback?

### DIFF
--- a/server/src/servershot.js
+++ b/server/src/servershot.js
@@ -112,7 +112,11 @@ if (!config.useS3) {
 
   get = (uid, contentType) => {
     return new Promise((resolve, reject) => {
-      s3bucket.createBucket(() => {
+      s3bucket.createBucket((err) => {
+        if (err) {
+          mozlog.error("error-getting-bucket", {err});
+          reject(err);
+        }
         const params = {Key: uid};
         s3bucket.getObject(params, function(err, data) {
           if (err) {


chenba:
Agreed with 6a68 that this probably isn't the right patch for #3771.

PR #4107, opened 2018-02-13
Fix #3642, Require node 8 for docker and in package.json

### LINKED ISSUE #3642: Make node 8 a package.json requirement
I recently encountered a problem running under Node v6. So... somewhere along the way v8 became a hard requirement. I think there's something like an "engine" requirement we can put in package.json?

6a68:
@oremj I'm not running Docker locally. Do you know of any issues to worry about with bumping our Docker base image from node:7 to node:8, maybe bugs other services have run into?

Also, I'm guilty of skipping the eslint-related changes @pdehaan suggested in #3642...we can add those to a separate good-first-bug for node linting, maybe?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:7
+FROM node:8
 
 COPY package.json /app/
 COPY build/server /app/build/server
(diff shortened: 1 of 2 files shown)


oremj:
We haven't run in to any issues. As long as your code works with node 8, you should be good to go.

PR #4110, opened 2018-02-14
change flag to survey

### LINKED ISSUE #3654: Reconsider copy on flag email
Currently when you flag something the email starts out like:

```
Subject: Flagging shot for abuse
Body:
Flagging shot for abuse: (url)
```

We can include any copy we want here. We could include things like: "Can't delete your own shot? See {link} for help". Also we can ask for particular information.

Note this isn't translated. But I think that's okay because the people monitoring it only speak English, so we don't really want to invite non-English submissions.

johngruen:
Fixes #3654 
We should wait for approval from legal


6a68:
@johngruen Ping me in here when this is ready to land

PR #4114, opened 2018-02-15
Wait until save button is visible in a test 🤷‍♂️ 

### LINKED ISSUE #4108: intermittent selenium bug: ElementNotInteractableError
[This build](https://circleci.com/gh/mozilla-services/screenshots/5140) failed with this error message:

`screenshots-server.console.ERROR: error {"msg":"(node:23714) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): ElementNotInteractableError"}`

I wonder if doing something like (1) wait to find the element, then (2) wait for the element to be enabled, would fix this.

chenba:
Fixes #4108, maybe?

Since the issue is intermittent and occurs only on CircleCI, it's difficult to say with some degree of certainty that this fixes #4108 (even if the tests have passed in five builds for #4114 on CircleCI).

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -45,7 +45,7 @@ function getDriver() {
 
   const driver = new webdriver.Builder()
     // .setFirefoxService(servicebuilder)
-    .withCapabilities({"moz:webdriverClick": false})
+    .withCapabilities({"moz:webdriverClick": true})
     .forBrowser("firefox")
     .setFirefoxOptions(options)
     .build();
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -224,6 +224,10 @@ describe("Test Screenshots", function() {
       return driver.wait(
         until.elementLocated(By.css(".preview-button-save"))
       );
+    }).then((saveButton) => {
+      return driver.wait(
+        until.elementIsVisible(saveButton)
```
6a68: If this isn't enough to eliminate intermittent failures, there's an `element.isEnabled()` method which polls for a visible form element to not have the `disabled` attribute on it. Some discussion in [this SO answer](https://stackoverflow.com/questions/26241495/how-do-i-wait-for-an-element-to-be-clickable-in-webdriverjs). We'd have to start toggling the `disabled` attribute when frames are loaded, but it could be worth it.

PR #4115, opened 2018-02-15
Switch en-US FTL file to new syntax

flodolo:
* `#` for comment sigil
* no more sections,`##` for section comments
* equal sign on empty value

CC @pike

### DIFF
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -1,21 +1,16 @@
-// Localization for Server-side strings of Firefox Screenshots
-//
-// Please don't localize Firefox, Firefox Screenshots, or Screenshots
+### Localization for Server-side strings of Firefox Screenshots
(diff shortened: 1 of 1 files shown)


locales/en-US/server.ftl
```diff
@@ -1,21 +1,16 @@
-// Localization for Server-side strings of Firefox Screenshots
-//
-// Please don't localize Firefox, Firefox Screenshots, or Screenshots
-
-
-
-// Global phrases shared across pages, prefixed with 'g'
-[[ global ]]
+### Localization for Server-side strings of Firefox Screenshots
+### Please don't localize Firefox, Firefox Screenshots, or Screenshots
```
Pike: Should we add Terms for these?

flodolo: That would invalidate a ton of existing translations. So, I wouldn't be happy about it.

We might think about introducing terms and using them in new strings for the future.

PR #4123, opened 2018-02-16
Use done cb in Selenium tests to fail faster

### LINKED ISSUE #4109: intermittent selenium bug: UnhandledPromiseRejectionWarning
Related to #4108, I thought mocha should gracefully handle uncaught Promise failures, but I guess not. Maybe we need to add a generic `.catch(err => fail(err))` or something throughout?

chenba:
Fixes #4109

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -179,9 +179,9 @@ describe("Test Screenshots", function() {
     return null;
   });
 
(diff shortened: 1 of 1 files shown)


6a68:
🎉  nice find! totally overlooked that

PR #4154, opened 2018-02-26
Add download shot Selenium test.

### LINKED ISSUE #4026: Add Selenium Test for Downloading a Shot
Add a Selenium test to verify that a shot has been successfully downloaded.

chenba:
Fixes  #4026

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -17,6 +17,8 @@ const { By, until } = webdriver;
 // Uncomment the next line and others with `ServiceBuilder` to enable trace logs from Firefox and Geckodriver
 // const { ServiceBuilder } = firefox;
 const path = require("path");
+const fs = require("fs");
+const util = require("util");
 
 const SHOOTER_BUTTON_ID = "pageAction-panel-screenshots";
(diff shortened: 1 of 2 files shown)


chenba:
On hold until I update the filename checking to account for the change in #4199

test/test.js
```diff
@@ -310,6 +322,43 @@ describe("Test Screenshots", function() {
...
+          return false;
+        }
+        const stat = fs.statSync(path.join(downloadDir, filename));
+        return ((stat.mtime.getTime() - Date.now()) < 2000);
+      });
+      assert(matches.length, `Shot was not downloaded to ${downloadDir}`);
+      // clean up the dir
+      const unlinkPromise = util.promisify(fs.unlink);
+      files.forEach(filename => {
+        filename !== ".gitignore" && unlinkPromise(path.join(downloadDir, filename));
```
ianb: Given that the promise isn't being used for anything, I guess promisifying isn't really needed here (though it doesn't hurt either). `return Promise.all(files.map(...))` would actually block on the cleanup, though I don't think that really matters either.

PR #4189, opened 2018-03-06
Fix DPI scaling in download/copy 

### LINKED ISSUE #4174: Screenshots doesn't account for DPI scaling anymore
STR:

1. Have dpi scale >100%
2. Take a screenshot using either the drag / click method or the 'Save full page' button. Download it somewhere.

AR:

1. screenshot turns out to be tinier than what you had on screen
2. You now need to zoom in in your image viewer to make things readable but also blurry in the process

ER: it should be the same size, like it used to be until recently

The 'Save visible' button does still respect DPI. Issue seems to be with only drag, select and the 'Save full page' button.

I'm on Nightly 60.0a1 (2018-03-03), Screenshots v30; Windows 10.

6a68:
Fixes #4174. Changes:

* don't overwrite correctly-scaled canvas dimensions
* pass the shot type to `downloadShot` and `copyShot`

@chenba Mind taking a look? I git bisected the bug back to commit 6fa7f3a8a, but I may not have all the context around the reasons for the changes in that commit. In particular, I'm curious if "fullPage" was hard-coded for any particular reason.

### DIFF
--- a/addon/webextension/selector/shooter.js
+++ b/addon/webextension/selector/shooter.js
@@ -84,9 +84,7 @@ this.shooter = (function() { // eslint-disable-line no-unused-vars
     }
     const canvas = captureToCanvas(selectedPos, captureType);
     ui.iframe.showLoader();
-    const width = selectedPos.right - selectedPos.left;
-    const height = selectedPos.bottom - selectedPos.top;
-    const imageData = canvas.getContext("2d").getImageData(0, 0, width, height);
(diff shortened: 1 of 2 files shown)


chenba:
Well, there's always the simple, most likely explanation that I screwed up.  I believe the thinking was that only fullPage shots would use the the promise route as the other shots can use the preview image.

I've filed #4193 as a related follow-up.

6a68:
lol, no worries! just wanted to make sure I wasn't missing any edge cases when doing manual testing of the patch. thanks for the quick review :beers:

PR #4199, opened 2018-03-07
Standard ISO 8601 date for filename, fixes #4197

### LINKED ISSUE #4197: Format date of title/name to YYYY-MM-DD
I'm not sure, but I think it's [this line of code responsible](https://github.com/mozilla-services/screenshots/blob/4e093dd8caaf241463b0b8a222fca2f47cd16602/shared/shot.js#L377) for it:

```
let clipFilename = `Screenshot-${date.getFullYear()}-${date.getMonth() + 1}-${date.getDate()} ${filenameTitle}`;
```

Currently, the date in a screenshot's name is not following an ISO Standard. This is confusing, e.g. screenshots saved today will have the title "Screenshot-2018-3-7 whatever.png". I'd like to have a consistent scheme, e.g. "Screenshot-2018-03-07 whatever.png" (this should be [ISO 8601](https://xkcd.com/1179/)).

The only easy way of doing this in JavaScript might be this piece of code (date formatting in JavaScript is … confusing. There's not really an easy way to do this, right?):

```
var today = new Date();
today.toISOString().substring(0, 10);
```

I can't test that (don't want to set everything up), but maybe someone could make a PR using this?

Edit: Sorry, toISOString() won't work properly because of GMT?! Any other solution?

KopfKrieg:
I couldn't test it yet, but this should fix issue #4197. Maybe someone can test it?

### DIFF
--- a/shared/shot.js
+++ b/shared/shot.js
@@ -374,7 +374,7 @@ class AbstractShot {
     // eslint-disable-next-line no-control-regex
     filenameTitle = filenameTitle.replace(/[:\\<>/!@&?"*.|\x00-\x1F]/g, " ");
     filenameTitle = filenameTitle.replace(/\s{1,4000}/g, " ");
-    let filenameDate = new Date(date.getTime() - date.getTimezoneOffset() * 60 * 1000).toISOString().substring(0,10);
+    const filenameDate = new Date(date.getTime() - date.getTimezoneOffset() * 60 * 1000).toISOString().substring(0, 10);
     let clipFilename = `Screenshot_${filenameDate} ${filenameTitle}`;
     const clipFilenameBytesSize = clipFilename.length * 2; // JS STrings are UTF-16
     if (clipFilenameBytesSize > 251) { // 255 bytes (Usual filesystems max) - 4 for the ".png" file extension string
--- a/shared/shot.js
+++ b/shared/shot.js
@@ -374,8 +374,7 @@ class AbstractShot {
     // eslint-disable-next-line no-control-regex
     filenameTitle = filenameTitle.replace(/[:\\<>/!@&?"*.|\x00-\x1F]/g, " ");
     filenameTitle = filenameTitle.replace(/\s{1,4000}/g, " ");
-    let today = new Date();
-    let filenameDate = new Date(today.getTime() - today.getTimezoneOffset() * 60 * 1000).toISOString().substring(0,10);
+    let filenameDate = new Date(date.getTime() - date.getTimezoneOffset() * 60 * 1000).toISOString().substring(0,10);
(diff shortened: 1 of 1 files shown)


ianb:
Thanks for the patch @KopfKrieg! If you look at the failures you'll see they are lint errors – you should be able to see those locally too if you run `npm run lint`. It might just be the undeclared `today` variable?

6a68:
@KopfKrieg should be possible to test it by doing `./bin/run-server` in one terminal, and `./bin/run-addon` in another, then take a shot and see if the date looks right

KopfKrieg:
Whoops, sorry. Made a few errors. Will try again.

shared/shot.js
```diff
@@ -374,7 +374,9 @@ class AbstractShot {
     // eslint-disable-next-line no-control-regex
     filenameTitle = filenameTitle.replace(/[:\\<>/!@&?"*.|\x00-\x1F]/g, " ");
     filenameTitle = filenameTitle.replace(/\s{1,4000}/g, " ");
-    let clipFilename = `Screenshot-${date.getFullYear()}-${date.getMonth() + 1}-${date.getDate()} ${filenameTitle}`;
+    today = new Date()
```
6a68: @KopfKrieg note that `date` is the created date of the shot. you'll want to base the filename on that date, as in the line you're replacing

shared/shot.js
```diff
@@ -374,7 +374,9 @@ class AbstractShot {
     // eslint-disable-next-line no-control-regex
     filenameTitle = filenameTitle.replace(/[:\\<>/!@&?"*.|\x00-\x1F]/g, " ");
     filenameTitle = filenameTitle.replace(/\s{1,4000}/g, " ");
-    let clipFilename = `Screenshot-${date.getFullYear()}-${date.getMonth() + 1}-${date.getDate()} ${filenameTitle}`;
+    let today = new Date();
+    let filenameDate = new Date(today.getTime() - today.getTimezoneOffset() * 60 * 1000).toISOString().substring(0,10);
```
6a68: whoops! should have commented on the whole PR, not that (now outdated) commit. Note that the line you replaced uses `date`, which is based on the `createdDate` of the shot (line 373 above). You'll want to use that date for generating the `filenameDate`.

KopfKrieg: Sure!

KopfKrieg:
Hm, can't test this local, ./bin/run-server (as suggested) fails, and my JS knowledge is … very basic.

6a68:
@KopfKrieg What's going wrong with the server? Happy to help you debug :-)

6a68:
Actually, I guess if this is just for a downloaded file, `./bin/run-addon` is really all you need

PR #4210, opened 2018-03-09
Use string functions instead of regex for isUrl.

### LINKED ISSUE #2660: Use better validation in shot.js isUrl
The validation is just a regex.  There's not a single technique for validation across the browser and Node, so we have to find some way to handle both cases.

chenba:
Fixes #2660


chenba:
Ah, my bad, I should've checked.   Much shorter patch is up.

Issue #4240, opened 2018-03-19
Investigate Failing Selenium Tests in CircleCI


chenba:
Starting with [this build](https://circleci.com/gh/mozilla-services/screenshots/5392), three Selenium tests consistently fail in CircleCI.  No changes have been made to the tests.  And they do not fail locally.

chenba:
I get the same failures locally once I updated my Nightly.  However, they occur only in headless mode (or running in Xvfb).

chenba:
Last good revision: b61bf51fd0c2f788ce98c96d7f205cc4d5c48424
First bad revision: ad5d0d34bddb427b2b89d218d7761c472a64428a
[Pushlog: https://hg.mozilla.org/integration/mozilla-inbound/pushloghtml?fromchange=b61bf51fd0c2f788ce98c96d7f205cc4d5c48424&tochange=ad5d0d34bddb427b2b89d218d7761c472a64428a](https://hg.mozilla.org/integration/mozilla-inbound/pushloghtml?fromchange=b61bf51fd0c2f788ce98c96d7f205cc4d5c48424&tochange=ad5d0d34bddb427b2b89d218d7761c472a64428a)

chenba:
Using the latest geckodriver (1.11.0) did not make a difference.

chenba:
Filed [bug 1447366](https://bugzilla.mozilla.org/show_bug.cgi?id=1447366).

In the meantime, I'll temporarily disabled those three failing tests.

PR #4242, opened 2018-03-20
Disable three tests temporarily.

chenba:
Disable three tests that are failing because of changes in Firefox.  (See #4240)


chenba:
Oh yeah, skipping is way better.

PR #4248, opened 2018-03-20
Use backend env var in Selenium tests

### LINKED ISSUE #4247: Run Selenium Test with Other Server Backends
While we can build an addon with the backend set to the SCREENSHOTS_BACKEND env var, there's some hardcoding in test.js preventing people from running the tests without running the dev server.

chenba:
Fixes #4247

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -10,6 +10,16 @@ NO_CLOSE = if not empty then when the test is finished, the browser will not be 
 
 */
 
+const backend = process.env.SCREENSHOTS_BACKEND || "http://localhost:10080";
+const { URL } = require("url");
+let backendUrl;
+
+try {
+  backendUrl = new URL(backend);
+} catch (e) {
+  throw new Error(`SCREENSHOTS_BACKEND is not a valid URL: ${e.message}`);
+}
+
 const assert = require("assert");
 const firefox = require("selenium-webdriver/firefox");
 const webdriver = require("selenium-webdriver");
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -259,10 +269,7 @@ describe("Test Screenshots", function() {
       });
     }).then((shotUrl) => {
       assert(shotUrl.startsWith(backend), `Got url ${shotUrl} that doesn't start with ${backend}`);
```
6a68: Probably makes sense to move this part of the check inside the `verifyShotUrl` method, too, since it's duplicated in both places, and should be easy to compare the `URL` for expected vs actual backend.

test/test.js
```diff
@@ -182,6 +183,15 @@ function expectCreatedShot(driver, creator) {
   });
 }
 
+function verifyShotUrl(shotUrl) {
+  const restUrl = shotUrl.substr(backend.length);
+  const hostname = new URL(backend).hostname;
```
6a68: `new URL()` throws in a lot of cases. It might make sense to try/catch and rethrow with an error like "backend is not a valid URL, please check the value of `process.env.SCREENSHOTS_BACKEND`"

test/test.js
```diff
@@ -182,6 +183,15 @@ function expectCreatedShot(driver, creator) {
...
+  const regex = new RegExp(`^/[^/]+/${hostname}$`)
+  if (!regex.test(restUrl)) {
```
6a68: Since we're already using `URL` here, it might be more consistent and more readable to get rid of `restUrl` and `regex` and instead do something like,

```
let url = new URL(shotUrl);
if (!url.pathname.endsWith(hostname)) {
```

test/test.js
chenba: We won't arrive here if the backend URL is not valid.  I will put in a validation earlier.

PR #4250, opened 2018-03-21
Match CircleCI node version to the Docker image's.

### LINKED ISSUE #2216: React warning while loading shot page
add-ons_manager

I'm seeing an error when I try and load the following shot page: https://pageshot.net/PNDTvTF1Q6jgffrZ/testpilot.firefox.com

> _Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:_
> 
> &nbsp;&nbsp; (client) ="Page-Shot-2017-2-23 Firefox Test Pilot
> &nbsp;&nbsp; (server) ="Page-Shot-2017-2-24 Firefox Test Pilot

chenba:
Fixes #2216's deployment issue

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:8.10.0
 
 COPY package.json /app/
 COPY build/server /app/build/server
@@ -7,6 +7,7 @@ COPY build/shared /app/build/shared
 COPY node_modules /app/node_modules
 COPY bin/_run-docker /app/bin/
 COPY build/screenshots.xpi /app/build/xpi/screenshots.xpi
+ENV NODE_ICU_DATA="/app/node_modules/full-icu"
 RUN cd app && npm install
 
 CMD /app/bin/_run-docker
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -7,6 +7,7 @@ COPY build/shared /app/build/shared
 COPY node_modules /app/node_modules
 COPY bin/_run-docker /app/bin/
 COPY build/screenshots.xpi /app/build/xpi/screenshots.xpi
+ENV NODE_ICU_DATA="/app/node_modules/full-icu"
```
6a68: @chenba I'm a little confused by the comments in bug #2216. If the problem was a node version mismatch, do you still need to add this env var?

chenba: Yes.  Still need this for node.  The path was correct but the version of ICU data was incorrect last time around because CircleCI was using a different version of node than Docker.  (The data is pulled down during a postinstall script into `node_modules/full-icu`; it's not dependent on the full-icu module version.)  If I've left that env var along with the wrong version of ICU data, node won't run at all, so it was removed.

PR #3277, opened 2017-08-04
Resilient delete fix

ianb:
We should wait until a release to merge this PR

PR #4251, opened 2018-03-21
Resilient deletes

### LINKED ISSUE #3275: Delete images even when there are temporary failures
Right now we delete expired images loosely, and don't handle or track failures. We should do it better!

chenba:
Fixes #3275 

This follows the (unmerged) work done in #3277.  The sleep wait time during a server (re)start in the tests has been increased because of problem described in #2208.


6a68:
@ianb would you mind reviewing this one?

PR #4259, opened 2018-03-22
Fixes #4182 - Highlight entire word when editing title

### LINKED ISSUE #4182: When editing title of screenshot, highlight the entire word.
Currently, when editing a Screenshot’s title, the cursor is placed before all the text forcing the user to do extra work to erase the title or append something to the end.

screen shot 2018-03-05 at 10 17 37 am

Fix: We should highlight the entire text so when a user types it overrides everything (or they can arrow left or right to place the cursor where they need it).

punamdahiya:


punamdahiya:
@ianb Submitting patch with the fix, please review. Thanks!

PR #4263, opened 2018-03-23
Wait for onboarding status.

### LINKED ISSUE #3954: Screenshots 'forgets' that I have seen its tour
I'm unsure if this coincides with updates to the browser (I'm on Nightly, and update infrequently), or updates to the Screenshots add-on or both, but clicking the page action after a browser start-up often launches that tour.

I have 'Clear history when Nightly closes' enabled, but I don't clear cookies, site prefs or OWD, which AFAIK, are the only ones known to cause webext to lose their settings:
capture
Additionally, this doesn't happen on every browser restart, so that should probably means that the above setting is not to be blamed.

chenba:
Fixes #3954

### DIFF
--- a/addon/webextension/background/main.js
+++ b/addon/webextension/background/main.js
@@ -11,11 +11,9 @@ this.main = (function() {
   const manifest = browser.runtime.getManifest();
   let backend;
 
-  let hasSeenOnboarding;
-
-  browser.storage.local.get(["hasSeenOnboarding"]).then((result) => {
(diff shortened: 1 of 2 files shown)


SoftVision-CosminMuntean:
@6a68 unfortunately I haven't managed to reproduce the reported issue #3954, but I started to test different scenarios related to the onboarding using the latest Screenshots dev version. 
Until now I haven't found any issues, but I will continue to test it.

PR #3967, opened 2018-01-12
s/Photon/WebExtension/ for page action

chenba:
Fix #3756

PR #4264, opened 2018-03-23
Enable context menu icon.

### LINKED ISSUE #3858: Add icon to context menu item
It was [pointed out](https://twitter.com/mohnessem/status/935552509942353926) that we should have the screenshots icon next to the context menu item when right clicking the page.
image

chenba:
Fixes #3858


6a68:
I forgot to mention that the export documentation is inside /docs.

chenba:
Yep, no prob.  Already used the export script for #3967 

What's the alternative if there's a performance regression?   Try different image formats?

6a68:
> What's the alternative if there's a performance regression? Try different image formats?

Probably best to just ask the addons team if we see a regression; they might fix it themselves, or offer us a workaround. In the past, regressions have tended to be code that runs on each page load, like `tabs.onUpdated`, and not code that seems like it would only run once, like creating a context menu item with an icon - but you never know :-P

chenba:
[Talos results](https://treeherder.mozilla.org/perf.html#/compare?originalProject=mozilla-central&newProject=try&newRevision=dc749072cb75babc79fc1c3edd2277b4809cde27&framework=1&showOnlyComparable=1&showOnlyImportant=1&selectedTimeRange=604800).

6a68:
@chenba Great! Would you mind pinging jmaher in #perf to see if he thinks the results are ok?

chenba:
jmaher said: "I think it is ok- the good news it is gave us data to look at- the orange/red items are that color because it looks like a possible or likely regression; in these cases by looking at the graphs and zooming in, you can see that most data points are bi-modal and on the try push it falls heavier on a single mode- therefore making the numbers different"

😌

PR #4272, opened 2018-03-26
Do not uglify when NO_UGLIFY is set.

### LINKED ISSUE #4271: Do Not Run Bundled JS Through uglify When NO_UGLIFY is Non-Empty
The `bundle_dependencies` script is compressing and mangling the bundle even when NO_UGLIFY is set.  It shouldn't

chenba:
Fixes #4271


chenba:
`NO_UGLIFY` need be exported as an environment variable in this case.  (Or you can keep it as a shell var and call `bundle_dependencies` directly.)

Issue #3681, opened 2017-10-26
[FxA] The Firefox Accounts sign in no longer works


SoftVision-CosminMuntean:
[Affected versions]:
- Nightly 58.0a1
- Screenshots 22.0.0

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" boolean pref is set to "false".
- The "extensions.legacy.enabled" boolean pref is set tot "true".
- Have the Screenshots dev version installed on a new clean profile.
- The Screenshots Onboarding tour was already visited.

[Steps to reproduce]:
1. Start the browser with the profile from prerequisites and navigate to "https://screenshots.dev.mozaws.net/shots" page.
2. Click the "cogwheel" from the right corner of the page.
3. Click the "Sign In" button and enter a valid account.
4. Click the "Sign in" button.
5. Observe the browser behavior.

[Expected result]:
- You are successfully logged in.

[Actual results]:
- A blank page is displayed.

[Notes]:
- Also, you can't create a new account.
- If you refresh the blank page you get the following errors: 
![errors](https://user-images.githubusercontent.com/19226534/32057030-cc13dd44-ba6e-11e7-8973-a9abb0fca9d4.png)

- Here is a screenshot with browser console errors:
![fxa brwoser console errors](https://user-images.githubusercontent.com/19226534/32057034-cfa13718-ba6e-11e7-9a70-6227eb3053a4.png)

niharikak101:
@SoftVision-CosminMuntean This should be working now! Can you test it? Thank you.

chenba:
Adding this to current sprint since we already merged some code. 🤸‍♂️

SoftVision-CosminMuntean:
This issue is no longer reproducible on latest Firefox Screenshots (31.3.0) dev version. Tested on Windows 10 x64, 7 x64, Mac 10.13 and Ubuntu 14.0.4.

PR #4274, opened 2018-03-27
Fix last step in fxa oauth flow

6a68:
*Edited: this PR is now a complete fix for #3681.*

Partial fix for #3681. The implementation of the `request` and `readBody` methods inside `helpers.js` assumed that the wrapped `wreck` methods were callback-based, but those methods actually return promises, so the promise chain wasn't terminated properly. I don't see any need for the helpers, so just invoked the `wreck` methods directly.

@niharikak101 Does this unblock you with fxa debugging?

Without this patch, when I do the fxa login flow on localhost, it hangs forever after entering the password and confirmation code. (When I do the same on the dev server, as in the STR for #3681, I see a 504 Gateway Timeout after a 60 second wait, which I think is nginx timing out the connection.)

With this patch, when I do the fxa login flow on localhost, I immediately get a 403 error with message "Invalid OAuth access token".

Doesn't fix FxA, but does move the debugging one step forward :+1:

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "morgan": "1.9.0",
     "mozlog": "2.1.1",
     "multer": "1.3.0",
+    "node-fetch": "2.1.2",
     "node-statsd": "0.1.1",
     "nodemon": "1.12.1",
     "nodify-uuid": "0.0.1",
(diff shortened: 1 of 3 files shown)


niharikak101:
Thank you for taking the time to look at this, @6a68! Definitely helped to identify the root of the problem. I tried using `node-fetch` instead of `wreck` which is a lot simpler and also has support for `promises`. The oAuth flow is working as expected; I'm adding a new commit to this PR.

6a68:
sweet! Thanks for adding that commit, @niharikak101.

6a68:
wow! this indeed fixes fxa locally: I can connect one profile to fxa, then log in from a new scratch profile, log in with that account, and access the shots.

ianb:
Looks good – too late now, but it would have been a reasonable place to start using async/await.

Issue #3482, opened 2017-09-11
Write tests for invalid image files


g-k:
Write tests to cover the assertPng invalid-data-url, invalid-data-image, and invalid-data-image-decoded paths.

https://codecov.io/gh/mozilla-services/screenshots/src/a82b616ae48fbab057722efbeab649e765b540a6/server/src/servershot.js#L16

The tests could probably go in a new file in `test/server` and hit the PUT /data/:id/:domain route.

This might be a good starter bug.

leni1:
Hello. Interested in working on this issue. I wanted to know if there's any documentation regarding tests that I should look at before I start e.g. expected test frameworks to use and some sample tests that that the one I'm working on should mirror.

g-k:
Hi @leni1, thanks for taking a look at this!

I'd tackle this by:

1. copying one of the server test python files from https://github.com/mozilla-services/screenshots/tree/93997ed0fb6fed3d9d6ad52329139a1604596b4d/test/server to something like `test_image_validation.py`
2. renaming or deleting the other test functions so only tests for the invalid-data-url, invalid-data-image, and invalid-data-image-decoded cases remain
3. writing tests that POST data that doesn't match the content type
4. checking that the tests cover the above cases by any of the following:

* seeing the server log warnings for those error types
* running the server with `NODE_ENV=dev ENABLE_COVERAGE=1` and checking the local coverage report
* making a PR and checking the codecov.io coverage report

Let me know if you have any questions or get stuck.

leni1:
So looking at the function `assertPng`, I would like to know what `dataUrl` should be. Should I write it as is into the test or there's something else I should use?

g-k:
Good question! In the `assertPngOrJpeg` function the `dataUrl`  follows the https://en.wikipedia.org/wiki/Data_URI_scheme

There are examples in: https://github.com/mozilla-services/screenshots/blob/5bd458345a0613565bc8f0401904582ce50ca55c/bin/load_test_exercise_images.py

I'd recommend using [the test client library](https://github.com/mozilla-services/screenshots/blob/5bd458345a0613565bc8f0401904582ce50ca55c/test/server/clientlib.py) to generate valid images then changing the mimetype or magic bytes in the data url to something invalid.

leni1:
[...]

PR #4298, opened 2018-04-03
Add sample tests for bug #3482

leni1:
Added sample tests for bug #3482.


leni1:
@g-k this is an incomplete patch. I'm not sure how to get an invalid image from `example images`. Should I import it from `clientlib` or rely on `clientlib`'s methods to pull the image? Also I haven't been able to setup screenshots on my computer so I'll just have to hope that my code logic is sound.

6a68:
@leni1 Hey, why aren't you able to get Screenshots running locally? What error have you run into?

chenba:
@leni1 I saw your question on #screenshots but you weren't in the channel at that time... you need the `-f` flag before the file path for the db migration file.

6a68:
@leni1 also, to your question about testing for invalid images: you can just load a valid image, then modify it to make it invalid, such that it triggers each of the error cases mentioned in the bug.

In a little more detail: the bug is about testing the  `invalid-data-url`, `invalid-data-image`, and `invalid-data-image-decoded` cases. Those errors are thrown by the [`assertPngOrJpeg` function](https://github.com/mozilla-services/screenshots/blob/master/server/src/servershot.js#L21-L53).

Looking at that code, `invalid-data-url` is thrown if the image data is valid base64, but doesn't start with the png or jpeg base64 headers, so you could load any one of the [example images](https://github.com/mozilla-services/screenshots/blob/master/bin/load_test_exercise_images.py#L3), then search and replace the "data/image:png" bit in the test image's `url` with something else, like "data/image:foo".

To get an `invalid-data-image` error, you might remove a couple of characters from the png or jpeg headers in the example image, so that the bit after the "base64," does not match the expected [png](https://github.com/mozilla-services/screenshots/blob/master/server/src/servershot.js#L13) or [jpeg](https://github.com/mozilla-services/screenshots/blob/master/server/src/servershot.js#L15) headers, like changing `"iVBORw0KGgo="` into `"KGgo="`.

To get an `invalid-data-image-decoded` error, you'd want the right _number_ of characters in the png/jpeg header, so that it passes the `invalid-data-image` check, but make them the wrong characters. Something like changing `"iVBORw0KGgo="` into `"testRw0KGgo="` should work.

Does this make sense? Let us know if you get stuck again, and we can help you finish up this PR :+1:

g-k:
Hey @leni1, nice start!

> Should I import it from clientlib or rely on clientlib's methods to pull the image?

Try to use clientlib's methods, but if you need to import them that's OK too. What you have for `test_invalid_image_url` looks good.

I'd recommend getting Screenshots running locally if possible. Let us know if you run into any problems following [the directions](https://github.com/mozilla-services/screenshots#installation-and-setup). 

Tests also run automatically on CircleCI so you can see if your tests pass here too. For example, the results for this branch are at https://circleci.com/gh/mozilla-services/screenshots/5625 (from this page > "Show all checks" > "Details").

leni1:
@6a68 I got the error: `ERR_STREAM_WRITE_AFTER_END` when I attempted to run the server since I already have English US set as my locale and I wasn't sure if I was to run the database migration as well. Thanks for your comments on the tests; they've cleared up what had been confusing me. 
@chenba will try that. Sorry about disappearing from the channel; wasn't sure what timezones apply and when I should expect to find people.  
@g-k will do. Unfortunately CircleCI takes a while to load for me on my connection so I think I'll have to try getting `screenshots` to work locally.

6a68:
@leni1 Just checking in. Do you have any questions, or are you stuck on anything?

leni1:
@6a68 Hi! Sorry about my silence. I have been involved in an open source sprint as a mentor, so I haven't had time to go through the comments here. I think I will be relying on CircleCI to run tests since it will take too long for me to sort out why I am unable to have the database run on my computer.

leni1:
@g-k have added the test for `invalid_data_image`. I have added an invalid image manually but I know it makes the code look messy. How would I search and replace the `"iVBORw0KGgo"` with something invalid?

leni1:
> To get an invalid-data-image-decoded error, you'd want the right number of characters in the png/jpeg header, so that it passes the invalid-data-image check, but make them the wrong characters. Something like changing "iVBORw0KGgo=" into "testRw0KGgo=" should work.
@g-k According to what @6a68 said above, I'm not sure about whether I am supposed to call the `invalid-data-image` test when running the `invalid-data-image-decoded` test or simply use the same test but modify it to check for something different.

g-k:
Hey @leni1, this is looking good!

> I'm not sure about whether I am supposed to call the invalid-data-image test when running the invalid-data-image-decoded test or simply use the same test but modify it to check for something different.

The tests shouldn't call or modify each other, but could call helper functions (for example, to create an invalid image) if you think that would be helpful.

> Not sure about whether the approach for the invalid-data-url test is correct.

That looks right. I don't have a coverage report to verify but it should result in the "Invalid PNG image submitted" log statement.

> Also the tests presently do not cover the scenario where JPEG_HEADER_BASE64 = "/9g=" like say the example image given was this.

Yes, it would be good to test JPEG validation too. Can you make an issue like https://github.com/mozilla-services/screenshots/issues/3482 for adding those tests?

6a68:
Looks like the tests failed because of circle-ci connectivity issues. I triggered a rebuild.

leni1:
@g-k wrote up some helper functions for the tests. I was wondering if I should add the tests for `JPEG` images to this pull request or wait until this is successfully merged. In addition to that, what would be a valid `JPEG` image since the constant used in the `assertPng` function is `"/9g="` yet the sample `JPEG` image we have in `example_images` uses `/9j/` as the header. If I were to test this, would I be assuming that a valid `JPEG` image begins with `/9` or it is more fine-grained than that?
Otherwise tests pass according to CircleCI and everything looks good! :smile:

g-k:
@leni1 Great! Thanks for making the issue for testing JPEG validation too.

> I was wondering if I should add the tests for JPEG images to this pull request or wait until this is successfully merged.

Let's get this PR merged and make a separate PR for JPEG validation testing. It's good to have one pull request per issue.

If you're planning to work on https://github.com/mozilla-services/screenshots/issues/4395, comment in that issue and we can assign it to you. I answered your JPEG questions in that issue too. Let me know if you have any follow up questions.

leni1:
@g-k I was wondering if this is ready to be merged. Since CircleCI says that the tests pass, I'm assuming that the bulk of the work is done, and what is left is to check that the inclusion of the tests doesn't break anything that it shouldn't.

g-k:
Hey @leni1, this is looking really good, almost there. I'm not sure why we don't have a coverage report in codecov.io, but I ran a coverage report on 727fb7767fd4075016f3a59daf39a7e35cd412d5 locally.

The invalid-data-url, invalid-data-image-decoded tests are working 👍 , but the test for invalid-data-image isn't:

screen shot 2018-05-11 at 1 39 28 pm

A truncated data url like `data:image/png;base64,tooshort` should hit that path since it's behind a length check.

leni1:
@g-k thanks for the review! So tests pass again although according to CircleCI, the surrounding setup has not worked, so I guess you can cross check by doing a rebuild. :smile:

test/server/test_image_validation.py
```diff
@@ -0,0 +1,84 @@
...
+
+def test_invalid_data_image():
+    image = image_setup()
+    (shot_data, shot_json, user) = user_setup()
+    valid_header = "data:image/png;base64,iVBORw0KGgo"
+    invalid_header = "data:image/png;base64,R0k"
+    if valid_header in image:
+        invalid_data_image = image.replace(valid_header, invalid_header)
+        for clip_id in shot_json['clips']:
+            shot_json['clips'][clip_id]['image']['url'] = invalid_data_image
```
g-k: This still ends up being too long and doesn't hit the invalid-data-image code path.

@leni1 can you set the url here to `invalid_header`? and delete the `image.replace` two lines up?

g-k:
This looks good to me. 🎉 

screen shot 2018-05-15 at 4 03 45 pm

@6a68 do you want to take a look or is it ok if I land this?

g-k:
👍 thanks @6a68 

Test failures are in the JS linting and this PR is only new python code.

Thanks @leni1!

PR #4341, opened 2018-04-14
Update delete shot confirmation dialog.

### LINKED ISSUE #4186: Inconsistent deletion messaging
When deleting an image from the library view, I see this message:

screen shot 2018-03-05 at 1 30 08 pm

When deleting from the image view, the message is:

screen shot 2018-03-05 at 1 30 24 pm

Let's pick one and use it for both areas. I feel like the latter offers more clarity, so let's use that.

chenba:
Fixes #4186

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -5,6 +5,7 @@ const { Localized } = require("fluent-react/compat");
 const { Footer } = require("../../footer-view");
 const sendEvent = require("../../browser-send-event.js");
 const { ShareButton } = require("../../share-buttons");
+const { DeleteShotButton } = require("../../delete-shot-button");
 const { TimeDiff } = require("./time-diff");
 const reactruntime = require("../../reactruntime");
 const { Editor } = require("./editor");
@@ -182,15 +183,17 @@ class Body extends React.Component {
     this.setState({closeBanner: true});
   }
 
-  onClickDelete(e) {
+  clickDeleteHandler() {
     sendEvent("start-delete", "navbar", {useBeacon: true});
-    const confirmMessage = document.getElementById("shotPageConfirmDelete").textContent;
-    if (window.confirm(confirmMessage)) {
-      sendEvent("delete", "popup-confirm", {useBeacon: true});
-      this.props.controller.deleteShot(this.props.shot);
-    } else {
-      sendEvent("cancel-delete", "popup-confirm");
-    }
+  }
+
+  confirmDeleteHandler() {
+    sendEvent("delete", "popup-confirm", {useBeacon: true});
+    this.props.controller.deleteShot(this.props.shot);
+  }
+
+  cancelDeleteHandler() {
+    sendEvent("cancel-delete", "popup-confirm");
   }
 
   onClickFlag(e) {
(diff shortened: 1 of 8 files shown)


chenba:
@flodolo Heh, those title attributes sounded pretty awkward to me when I was reviewing the commit.  I'll change them to simply match the button text.

server/src/pages/shotindex/view.js
```diff
@@ -293,6 +293,19 @@ class Card extends React.Component {
   constructor(props) {
     super(props)
     this.state = {panelOpen: "panel-closed", deleted: false};
+    this.maybeCloseDeleteConfirmation = this.maybeCloseDeleteConfirmation.bind(this);
```
6a68: There is a lot of code duplication between these two view files. Could you extract the modal code to a single shared component?

static/css/shot-index.scss
```diff
@@ -252,6 +258,17 @@ h1 {
   }
 }
 
+.delete-confirmation-dialog {
```
6a68: Maybe the default dialog style could be declared in one spot, and overridden in the other, to minimize duplication? Not sure if it's worth it, your call

static/css/partials/_delete-confirmation.scss
```diff
@@ -0,0 +1,53 @@
...
+      height: 0;
+      border-left: 8px solid transparent;
+      border-right: 8px solid transparent;
+      border-bottom: 13px solid #fff;
+      left: -8px;
+      top: 2px;
+    }
+  }
+  .delete-confirmation-message {
+    font-size: 12px;
```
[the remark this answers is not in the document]
chenba: hmm I thought I grabbed the font-size straight out of https://sevaan.github.io/fxscreenshots/menus/#artboard0, but it's 13px there (which is still small).  @sevaan what do you think of 15px?

server/src/pages/shot/view.js
```diff
@@ -184,13 +197,32 @@ class Body extends React.Component {
...
+
+  onCancelDelete(e) {
+    sendEvent("cancel-delete", "popup-confirm");
+    this.setState({ confirmDelete: false });
+  }
+
+  maybeCloseDeleteConfirmation(e) {
+    let el = e.target;
+
+    while (el) {
```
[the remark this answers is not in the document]
chenba: True.  Just need to get `ref`s to the dialog and the delete button.

static/css/shot-index.scss
chenba: Since the dialog is absolute positioned to the delete button's parent, the defaults that work for one page would need to be completely overridden for the other page anyway, so I left them out of the partial.

server/src/pages/shotindex/view.js
chenba: Yeah, that code duplication is pretty gross.  I'll try styling up a div as a button and have the confirmation dialog in there.  (Currently the dialog is actually a sibling of the button...which is the main reason they aren't in a component.)

static/css/shot-index.scss
6a68: Right, that makes sense.

static/css/partials/_delete-confirmation.scss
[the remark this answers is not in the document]
sevaan: You're right, that does seem small. 15px would work great (matching the font size that says "2 minutes ago" in your screenshot above.

server/src/pages/shot/view.js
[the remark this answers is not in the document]
6a68: Yeah. If this is getting extracted to its own component, maybe pass the delete button ref into the constructor: "dialog, attach to this element here"

PR #4342, opened 2018-04-14
Fix #4183, make onboarding behaviour consistent

### LINKED ISSUE #4183: If Screenshots is first used on a non-standard web page, the Screenshots "#hello" page is wrongly opened in a new tab
[Notes]:
- This issue is not reproducible if you use the context menu option.

[Affected versions]:
- Nightly 60.0a1
- Screenshots v31.1.0 dev

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have the "xpinstall.signatures.required" pref set to "false" and "extensions.legacy.enabled" pref set to "true".
- Have the latest Firefox Screenshots v31.1.0 installed from [here](https://screenshots.dev.mozaws.net/homepage/install-test-local.html) on a new clean profile.
- Firefox Screenshots has not been used.

[Steps to reproduce]:
1. Open Firefox browser with the profile from prerequisites and navigate to a non-standard web page such as https://addons.mozilla.org/en-US/firefox/.
2. Click the "Page actions" menu.
3. Select "Take a Screenshot" option and observe the behavior.

[Expected result]:
- "We can't screenshot this page." notification appears.

[Actual result]:
- A new tab is opened and you are redirected to the "https://screenshots.firefox.com/#hello" page.

[Additional notes]:
ss - non standard page

testeaxeax:
Fix #4183, this makes the onboarding behaviour for unsupported websites consistent between the "Page actions" menu and the context menu.

### DIFF
--- a/addon/webextension/background/main.js
+++ b/addon/webextension/background/main.js
@@ -130,19 +130,25 @@ this.main = (function() {
   }
 
   exports.onClickedContextMenu = catcher.watchFunction((info, tab) => {
-    if (!tab) {
-      // Not in a page/tab context, ignore
-      return;
-    }
-    if (!urlEnabled(tab.url)) {
-      senderror.showError({
-        popupMessage: "UNSHOOTABLE_PAGE"
-      });
-      return;
-    }
-    catcher.watchPromise(
-      toggleSelector(tab)
-        .then(() => sendEvent("start-shot", "context-menu", {incognito: tab.incognito})));
+    catcher.watchPromise(hasSeenOnboarding.then(onboarded => {
+      if (!tab) {
+        // Not in a page/tab context, ignore
+        return;
+      }
+      if (!urlEnabled(tab.url)) {
+        if (!onboarded) {
+          sendEvent("goto-onboarding", "selection-button", {incognito: tab.incognito});
+          forceOnboarding();
+          return;
+        }
+        senderror.showError({
+          popupMessage: "UNSHOOTABLE_PAGE"
+        });
+        return;
+      }
+      return toggleSelector(tab)
+        .then(() => sendEvent("start-shot", "context-menu", {incognito: tab.incognito}));
+    }));
   });
 
   function urlEnabled(url) {


6a68:
Thanks for submitting this pull request! I can take a look

sevaan:
Thanks for your help testeaxeax! I think 6a68's description of the behaviour is correct: No error pop-up for users taking their first screenshot, just pop the new tab and launch the tour. If a screenshot has already been taken successfully previously, then revert to the error message and no popped tab.

addon/webextension/background/main.js
```diff
@@ -130,19 +130,26 @@ this.main = (function() {
...
-    if (!urlEnabled(tab.url)) {
-      senderror.showError({
-        popupMessage: "UNSHOOTABLE_PAGE"
-      });
-      return;
-    }
-    catcher.watchPromise(
-      toggleSelector(tab)
-        .then(() => sendEvent("start-shot", "context-menu", {incognito: tab.incognito})));
+    catcher.watchPromise(hasSeenOnboarding.then(onboarded => {
```
6a68: I think the flow here should be something like:

```js
hasSeenOnboarding.then(onboarded => {
  if (!onboarded) { ... }
  if (!tab) { ... }
  if (!urlEnabled) { ... }
  // then, toggleSelector()...
})
```

This way, no matter how you first trigger screenshots, you'll see onboarding as a new user.

addon/webextension/background/main.js
```diff
@@ -130,19 +130,26 @@ this.main = (function() {
...
+          sendEvent("goto-onboarding", "selection-button", {incognito: tab.incognito});
+          forceOnboarding();
+          return;
+        }
+        senderror.showError({
+          popupMessage: "UNSHOOTABLE_PAGE"
+        });
+        return;
+      }
+      catcher.watchPromise(
```
6a68: `watchPromise` [manages error handling](https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/catcher.js#L65) for Promise chains. Since you've added one at line 133 above, this one's no longer needed.

I think you'll also want to `return toggleSelector...` to keep the promises chained.

addon/webextension/background/main.js
6a68: Good point! Thanks for continuing to work on this, I've gotten a bit behind on reviews. I'll take a look today or Monday :-)

testeaxeax:
@6a68 I don't know how to fix the error in automated testing without breaking the promise chain for `toggleSelector`. Thanks in advance.

6a68:
@testeaxeax Sorry for the delay in getting back to you.

> I don't know how to fix the error in automated testing without breaking the promise chain for toggleSelector. Thanks in advance.

You don't need to return the promise. This code is called in [one spot](https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/background/startBackground.js#L41) which doesn't further chain the response, and the catcher wrapper handles any returned errors.

6a68:
This is good to land, once the linter error is fixed. Nice work 🎉

6a68:
We talked this over in triage today. I'm going to land this as-is, and then I'll fix the linter error in a followup PR.

Issue #4337, opened 2018-04-13
If the Screenshots was never used, the first saved shot is wrongly displayed as you are not the full owner of it


SoftVision-CosminMuntean:
[Affected versions]:
- Nightly 61.0a1
- Screenshots 31.3.0 dev

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" boolean pref is set to "false".
- The "extensions.legacy.enabled" boolean pref is set tot "true".
- Have the Screenshots 31.3.0 dev version installed from [here](https://screenshots.dev.mozaws.net/homepage/install-test-local.html), on a new clean profile. 
- Screenshots was never used.

[Steps to reproduce]:
1. Open the Firefox browser with the profile from prerequisites and navigate to any website.
2. Click the "Take a Screenshots" button from "Page Actions" menu.
3. Skip the Screenshots Onboarding tour.
4. Perform and save a selection on the page. 
5. Observe the saved screenshot page.
6. Click the "My Shots" button from the top left corner of the page and observe the behavior.

[Expected result]:
Step 5. The saved shot page is correctly displayed as you are the owner of the shot.
Step 6. You are redirected to the "My Shots" Page.

[Actual results]:
Step 5: Wrong icon is displayed on the "My Shots" button and the "Flag" button is wrongly displayed instead of the "Delete" button.
Step 6: The "My Shots" button wrongly redirects you to the Onboarding page instead of the "My Shots" page.

[Additional Notes]:
- The issue is no longer reproducible if the page is refreshed.
- The issue is only reproducible for the first saved shot.
- The issue is not reproducible using Screenshots 31.1 production version from latest Nightly.
- Here are the errors displayed in browser console:
![owner](https://user-images.githubusercontent.com/19226534/38734532-aec20816-3f2e-11e8-8ff9-45144fda6fbc.png)
- Here is a screen recording with the issue:
![first saved shot issue](https://user-images.githubusercontent.com/19226534/38734542-b74670ee-3f2e-11e8-8fcb-db1b88b48027.gif)

chenba:
[...]

Issue #4344, opened 2018-04-18
[Annotations] Edit button does not have its badge when you first use Firefox Screenshots


Softvision-CristinaBadescu:
It seems that this is a Nightly regression. I have also logged this issue in Bugzilla: #[1454914](https://bugzilla.mozilla.org/show_bug.cgi?id=1454914).

[Affected versions]:
- Nightly 61.0a1, Build ID: 20180417225505

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Open the Firefox browser and navigate to any website.
2. Click the "Page actions" menu and select "Take a Screenshot" option.
3. Skip the onboarding tour.
4. Click the "Save visible" button and select the "Save" button.
5. Observe the "Edit" (Pencil) button.

[Expected result]:
- The button has a blue circle that has a star in it.

[Actual result]:
- The button has no badge.

[Regression]:
- It seems that this is a Nightly regression. Using mozregression tools I have found a regression window. Here are the results:
Last good revision: ae45e56c3c71ae23084e9ef549c3d58880add8a1
First bad revision: 87b552f9c09b68a255eeb84c5c554384b0c1231a
Pushlog: https://goo.gl/pYTBYX

From the pushlog it seems that bug [1452496](https://bugzilla.mozilla.org/show_bug.cgi?id=1452496) has caused this issue. 

[Notes]:
- This issue is also reproducible on an older Nightly 61.0a1 build ID 20180108220132 with latest Firefox Screenshots v32.0.0dev installed from [here](https://screenshots.dev.mozaws.net/homepage/install-test-local.html) and on the latest Beta 60.0b13 with latest Firefox Screenshots v30.1.0.
- This issue is not reproducible on the latest Firefox Beta 60.0b12 with Firefox Screenshots 30.1.0.
- This issue is not reproducible on the latest Firefox Release 59.0.2 with Firefox Screenshots 25.0.0.
- Here is a screen recording with this issue:
![screenshots - edit button](https://user-images.githubusercontent.com/31244235/38930641-420a5676-4319-11e8-9766-e85a8db7d27d.gif)

chenba:
This is a duplicate of #4337.

chenba:
@Softvision-CristinaBadescu Thanks for finding that bug!

chenba:
[...]

PR #4345, opened 2018-04-18
Downgrade react to latest 15.x.x.

chenba:
Fixes part of  #4337


chenba:
Re-opening this so we can keep stage and prod running React 15.x.x.  Related to #4344.

PR #4350, opened 2018-04-21
[HOLD] Add a keyboard shortcut for Screenshots

### LINKED ISSUE #2491: Triage: keyboard shortcut
We haven't implemented a keyboard shortcut.  For accessibility we may want to (see #2251).  If we do, we should find someone that can help us find a keyboard shortcut that won't cause too many problems, as conflicts seem common.

Note there's a metric associated with this, if we don't implement a keyboard shortcut we should update METRICS.md to show it is deprecated; if we do implement it we should use the documented metric.

6a68:
* Alt + Shift + S on all platforms, except
* Command + Shift + S on Mac

See https://github.com/mozilla-services/screenshots/issues/2491#issuecomment-383271920 for discussion of the key combination choice.

This choice doesn't seem to conflict with any standard Firefox, Devtools, Windows, or Mac shortcuts.

I also updated METRICS.md to document the new `start-shot` argument.

### DIFF
--- a/addon/webextension/background/main.js
+++ b/addon/webextension/background/main.js
@@ -130,6 +130,16 @@ this.main = (function() {
   }
 
   exports.onClickedContextMenu = catcher.watchFunction((info, tab) => {
+    onUserInput("context-menu", tab);
+  });
+
+  exports.onKeyboardShortcut = catcher.watchFunction((tab) => {
+    onUserInput("keyboard-shortcut", tab);
+  });
+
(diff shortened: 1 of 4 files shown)


6a68:
It's going to be a few weeks, at least, before we have enough data to decide whether to use the devtools Ctrl + Shift + S shortcut, or pick something else.

I'm going to close this to reduce the number of PRs hanging around.

PR #4356, opened 2018-04-23
clean up log.js

### LINKED ISSUE #4225: Remove console.log.bind(console) pattern in log.js
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1445772 to handle other cases in m-c. We don't need to manually bind console methods to the console object anymore as of Bug 
989619.

So we should be able to simplify these: https://github.com/mozilla-services/screenshots/blob/9e4ca4eebf7b4334c187e7ea5b1a11b8c56dbcc9/addon/webextension/log.js#L30-L44 to `exports.debug = console.debug`, etc.

josechoy:
fixes #4225


chenba:
@josechoy Thanks!

PR #4385, opened 2018-04-26
Set a default NODE_ENV of 'production' in the Docker image.

### LINKED ISSUE #4326: Set the NODE_ENV environment variable in prod
The `NODE_ENV` env var is not set in prod, which might not be desirable.  The only thing that this is currently affecting is the Raven configuration.

The simplest and most direct way to do this is just set it in the `_run-docker` script, where a few other env vars are set.

chenba:
Fixes #4326

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,5 +8,6 @@ COPY node_modules /app/node_modules
 COPY bin/_run-docker /app/bin/
 COPY build/screenshots.xpi /app/build/xpi/screenshots.xpi
 RUN cd app && npm install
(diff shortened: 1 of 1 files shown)


6a68:
@chenba What's the best way to test this?

chenba:
Easiest is probably building the image locally and run `docker run some_image bash -c 'echo "$NODE_ENV"'` and then run `docker run -e "NODE_ENV=somethingdifferent" some_image bash -c 'echo "$NODE_ENV"'` to see if you get something different.

Issue #4352, opened 2018-04-23
Firefox Screenshots dev version can not be installed from the website


Softvision-CristinaBadescu:
[Affected versions]:
- Nightly 61.0a1
- Screenshots 32.0.0 dev

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" boolean pref is set to "false".
- The "extensions.legacy.enabled" boolean pref is set tot "true".

[Steps to reproduce]:
1. Open the Firefox browser with the profile from prerequisites and navigate to https://screenshots.dev.mozaws.net/homepage/install-test-local.html.
2. Click the "Install" button.
3. Click the "Allow" and afterwards the "Install" button.
4. Observe the behavior.

[Expected result]:
- Firefox Screenshots dev version is installed.

[Actual results]:
- "The add-on could not be downloaded because of a connection failure" error message is displayed.

[Notes]:
- This issue is not reproducible if you use a custom build cloned from the latest master.
- Here is the error displayed in browser console:
![ss - can not install ss](https://user-images.githubusercontent.com/31244235/39116882-35bfbbd0-46ee-11e8-8f60-012ba2d4fad8.png)

- Here is a screen recording with the issue:
![ss - cannot install firefox screenshots dev version](https://user-images.githubusercontent.com/31244235/39116894-3d7eae6c-46ee-11e8-8b76-4335a1465c4d.gif)

chenba:
Seems to be working now.

chenba:
This is happening to me now.  I think the difference is that I updated to the latest Nightly.

chenba:
...and the bug has gone away in the latest Nightly.  I've verified in Windows 10 and Linux.

Softvision-CristinaBadescu:
I've noticed that eventually I can install the latest Firefox Screenshots dev version only after I restart the browser and re-install it, on latest Nightly build 61.0a1 (2018-04-24).

SoftVision-CiprianMuresan:
I think I've gotten to the bottom of this issue. We have 3 scenarios which reproduce this:

1. Create a new profile, set the prefs and trying to install the add-on from the webpage will result in the connection error message. After restarting the browser the add-on can be installed successfully.
2. Dismiss the dropdown when prompted to either install or cancel the install. Try to install the add-on from the page again and the add-on can't be installed until the browser is restarted (see 1.)
[...]

PR #4387, opened 2018-04-26
Add more info to README.

### LINKED ISSUE #4334: Make sure it's easy for designers to run the add-on locally and point it at different envs

chenba:
Fixes #4334

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,15 @@ You can find more information about Firefox Screenshots at the Mozilla Wiki page
 
 ### Installation and Setup
 
-Install [Postgres](http://www.postgresql.org/).  And do _one_ of the following:
+Install [Nightly or Developer Edition](https://www.mozilla.org/en-US/firefox/channel/desktop/).
+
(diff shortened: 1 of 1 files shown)


sevaan:
[These instructions](https://screenshots.dev.mozaws.net/homepage/install-test-local.html) should probably also link to the [Basic Instructions](https://github.com/mozilla-services/screenshots#installation-and-setup) in case the user has issues installing (like I did).

Having said that, I also found I was unable to follow all the instructions there. I think I installed Postgres (but I got an error), and I'm stuck on the "Run First Migration" step...not sure how to run it, or what to do at that point. And as for cloning a repo, do I just do that on the Github site, or is that a terminal command? I'm guessing Terminal, but am not sure what the command is...maybe spelling out all the individual steps rather than generalizing might be helpful to n00bs like me.

sevaan:
@SoftVision-CosminMuntean found a solution here: https://github.com/mozilla-services/screenshots/issues/4352#issuecomment-385913692

With that pref added, the add-on installed without issue.

6a68:
eh, let's land this

PR #4401, opened 2018-05-08
Fix #4147, Avoid including blank scrollbar area in save-visible shots

### LINKED ISSUE #4147: An extra border is wrongly displayed on the right side of the shot if the "Save visible" option is used
[Notes]:
- Is also reproducible on Mac if the "Show scrollbars" option is set to "Always".
- If the browser is resized so that the horizontal scrollbar is displayed, the issue is only reproducible if the page is scrolled to the right maximum limit.
- The issue is only reproducible for "Save visible" option.
 
[Affected versions]:
- Nightly 58.0.2 and up
- Screenshots v25.0.0
- Screenshots v29.0.0

[Affected Platforms]:
- All Windows
- All Linux
- All Mac

[Steps to reproduce]:
1. Open Firefox browser an navigate to "http://www.history.com/" website.
2. Click the "Take a Screenshot" option from "Page Actions" menu or context menu.
3. Click the "Save visible" button and save the shot
4. Observe the right part of the shot.

[Expected result]:
- The image is correctly displayed and no extra borders are displayed.

[Actual result]:
- An extra border is displayed.

[Additional Notes]:
border

6a68:
`document.documentElement.clientWidth` seems to be more reliable than `window.innerWidth` for dealing with scrollbars on both windows and osx. I'm not sure what other edge cases might be broken by this change, though :-P

For more discussion, see: https://tylercipriani.com/blog/2014/07/12/crossbrowser-javascript-scrollbar-detection/

### DIFF
--- a/addon/webextension/selector/uicontrol.js
+++ b/addon/webextension/selector/uicontrol.js
@@ -186,7 +186,7 @@ this.uicontrol = (function() {
       sendEvent("capture-visible", "selection-button");
       selectedPos = new Selection(
         window.scrollX, window.scrollY,
-        window.scrollX + window.innerWidth, window.scrollY + window.innerHeight);
+        window.scrollX + document.documentElement.clientWidth, window.scrollY + window.innerHeight);
       captureType = "visible";
       setState("previewing");
     },


6a68:
Hmm. Related, should I replace the `window.innerWidth - SCROLLBAR_WIDTH` logic with just  `document.documentElement.clientWidth` here?

https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/selector/uicontrol.js#L597

chenba:
@6a68 yeah, we should do that.  I'll file an issue.

PR #4403, opened 2018-05-08
Refactor Annotations.

### LINKED ISSUE #4389: [Annotations] Refactor editor.js
editor.js is unwieldy.  Let's refactor it before adding more annotation tools.

- Selection can be a class shared by the addon and the editor
- Move the current tools into their own components

chenba:
Fixes #4389

### DIFF
--- a/server/src/pages/shot/drawing-tool.js
+++ b/server/src/pages/shot/drawing-tool.js
@@ -1,0 +1,140 @@
+const React = require("react");
+const PropTypes = require("prop-types");
+const { Selection } = require("../../../shared/selection");
+
+exports.DrawingTool = class DrawingTool extends React.Component {
+  constructor(props) {
+    super(props);
+    this.canvas = React.createRef();
+    this.canvasWidth = parseInt(props.baseCanvas.style.width, 10);
+    this.canvasHeight = parseInt(props.baseCanvas.style.height, 10);
+  }
+
+  render() {
+    return <canvas
+      ref={this.canvas}
+      className={`image-holder centered ${this.state.classNames}`}
+      onMouseDown={this.onMouseDown.bind(this)}
+      onMouseMove={this.onMouseMove.bind(this)}
+      width={this.props.baseCanvas.width}
+      height={this.props.baseCanvas.height}
+      style={{width: this.props.baseCanvas.style.width,
+              height: this.props.baseCanvas.style.height}}></canvas>;
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
+    return {strokeStyle: nextProps.color, lineWidth: nextProps.lineWidth};
+  }
+
+  componentDidMount() {
+    this.drawingContext = this.canvas.current.getContext("2d");
+    this.drawingContext.scale(this.props.devicePixelRatio, this.props.devicePixelRatio);
+    this.setDrawingProperties();
+  }
+
+  setDrawingProperties() {
+    console.warn("Please override setDrawingProperties in your component.");
+  }
+
+  componentDidUpdate() {
+    this.setDrawingProperties();
+  }
+
+  captureMousePosition(e) {
+    const boundingRect = this.canvas.current.getBoundingClientRect();
+    return {
+      x: e.clientX - boundingRect.left,
+      y: e.clientY - boundingRect.top
+    };
+  }
+
+  onMouseDown(e) {
+    if (e.button !== 0 || this.isMousedown) {
+      return;
+    }
+    this.mousedownPosition = this.captureMousePosition(e);
+    this.isMousedown = true;
+    this.draw(this.mousedownPosition);
+    this.updateDrawnArea(this.mousedownPosition);
+  }
+
+  onMouseMove(e) {
+    if (!this.isMousedown) {
+      return;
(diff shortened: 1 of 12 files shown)


chenba:
/me add some comments to this big o' change

server/src/pages/shot/editor.js
```diff
@@ -104,61 +15,90 @@ exports.Editor = class Editor extends React.Component {
...
+          cancelCropHandler={this.onClickCancelCrop.bind(this)}
+          confirmCropHandler={this.onCropUpdate.bind(this)}
+          toolbarOverrideCallback={this.overrideToolbar.bind(this)} />;
+      default:
+        return null;
+    }
+  }
+
+  overrideToolbar() {
+    this.setState({overrideToolbar: this.state.tool});
```
chenba: This changes the state just so that `renderToolBar()` is called _after_ the selected tool's been mounted in the previous render.

server/src/pages/shot/editor.js
```diff
@@ -213,319 +144,64 @@ exports.Editor = class Editor extends React.Component {
...
+      return;
+    }
+
+    this.imageContext.drawImage(incomingCanvas,
+      affectedArea.left * this.devicePixelRatio,
+      affectedArea.top * this.devicePixelRatio,
+      affectedArea.width * this.devicePixelRatio,
+      affectedArea.height * this.devicePixelRatio,
+      affectedArea.left, affectedArea.top, affectedArea.width, affectedArea.height);
+  }
```
chenba: This saves an edit over a certain area of the image.  For tracking changes for undo/redo, we can first copy the affected area of the image into another image before applying the edit.

server/src/pages/shot/editor.js
```diff
@@ -213,319 +144,64 @@ exports.Editor = class Editor extends React.Component {
...
     const imageContext = this.imageCanvas.getContext("2d");
     this.imageContext = imageContext;
-    img.src = croppedImage.toDataURL("image/png");
-    this.removeCropBox();
-    this.cropToolBar = null;
+    img.src = incomingCanvas.toDataURL("image/png");
     this.setState({tool: this.previousTool});
-    imageIsCropped = true;
-    sendEvent("confirm-crop", "crop-toolbar");
   }
```
chenba: This replaces the image in the base canvas with the one in the incoming canvas.  For undo/redo, we insert a copy of the whole image as a key frame in the history.

server/src/pages/shot/highlighter-tool.js
```diff
@@ -0,0 +1,115 @@
...
+      compositeContext.globalCompositeOperation = "soft-light";
+    } else {
+      compositeContext.globalCompositeOperation = "multiply";
+    }
+
+    compositeContext.drawImage(this.canvas.current,
+      0, 0, this.canvasWidth, this.canvasHeight);
+    this.drawingContext.drawImage(composite,
+      0, 0, this.canvasWidth, this.canvasHeight);
+  }
```
chenba: This is necessary for the highlighter because it doesn't use the default `globalCompositeOperation`.  A complete image is created here for the editor.  (Alternatively we could pass a `globalCompositeOperation` back to the editor.  I prefer to keep editor simpler if possible.)

Code review note: since the affected area is known, the `drawImage()` calls should use that.  It would require a change in `DraingTool` to adjust the affected area for the `lineWdith` before calling this method.

shared/selection.js
```diff
@@ -0,0 +1,104 @@
+class Selection {
+  constructor(x1, y1, x2, y2) {
+    this.left = x1;
+    this.top = y1;
+    this.right = x2;
+    this.bottom = y2;
+  }
+
+  get top() {
+    return Math.min(this.y1, this.y2);
```
chenba: Code review note: could remove these `min` and `max` calls in the getters since coordinates are always sorted now.

server/src/pages/shot/drawing-tool.js
```diff
@@ -0,0 +1,140 @@
...
+    this.isMousedown = true;
+    this.draw(this.mousedownPosition);
+    this.updateDrawnArea(this.mousedownPosition);
+  }
+
+  onMouseMove(e) {
+    if (!this.isMousedown) {
+      return;
+    }
+    e.preventDefault();
```
chenba: Is this necessary?  Maybe prevent default on dragStart

server/src/pages/shot/pen-tool.js
```diff
@@ -0,0 +1,42 @@
...
+      strokeStyle: props.color || "#000",
+      lineWidth: props.lineWidth || 5,
+      classNames: "pen"
+    };
+  }
+
+  setDrawingProperties() {
+    this.drawingContext.globalCompositeOperation = "source-over";
+    this.drawingContext.strokeStyle = this.state.strokeStyle;
+    this.drawingContext.lineWidth = this.state.lineWidth;
```
chenba: `strokeStyle` and `lineWidth` need not be part of the component state.

ianb:
The synchronous code review was good. There's a few things we talked about that could be changed as a followup, though nothing particularly critical... more like changes that could happen as the code further develops.

Issue #4347, opened 2018-04-19
Add new block types


ianb:
Right now we have only `'dmca'` for a block type. I propose we add `usererror` (when a user can't figure out how to delete something) and `abuse` (illegal or something).

This requires a database migration (there's an enum), and we need to be sure that all blocks show up reasonably. We might have DMCA-specific language on the site.

`usererror` and `abuse` can display as though they are simply deleted, IMHO we don't need to inform users about the underlying reasons.

ianb:
Note that we'll follow up in https://github.com/mozilla-services/screenshots-admin/issues/29 to make use of these block types.

chenba:
Closing this now that #4417 is merged.

PR #4417, opened 2018-05-10
Add watchdog submissions DB table migration SQL.

### LINKED ISSUE #4376: [Watchdog] Add DB Migration Scripts
In order to track the Watchdog submissions, we need to make DB changes.  We should add the migration scripts for the changes before starting work on the submission requests.

chenba:
Fix #4376 and #4347

### DIFF
--- a/server/db-patches/patch-22-23.sql
+++ b/server/db-patches/patch-22-23.sql
@@ -1,0 +1,9 @@
+-- pg-patcher runs migrations in transactions. However, ALTER TYPE ... ADD VALUE
+-- cannot be executed inside a transaction block. We will swap out the enum type
+-- with a new one instead.
+ALTER TYPE shot_block_type RENAME TO old_shot_block_type;
+CREATE TYPE shot_block_type AS ENUM ('none', 'dmca', 'abuse', 'usererror', 'watchdog');
+ALTER TABLE data ALTER COLUMN block_type DROP DEFAULT;
+ALTER TABLE data ALTER COLUMN block_type TYPE shot_block_type USING block_type::text::shot_block_type;
+ALTER TABLE data ALTER COLUMN block_type SET DEFAULT 'none'::shot_block_type;
+DROP TYPE old_shot_block_type;
--- a/server/db-patches/patch-23-22.sql
+++ b/server/db-patches/patch-23-22.sql
@@ -1,0 +1,4 @@
+-- Records using the new enum values from 22-23 must be updated or deleted in
+-- order to update the enum type. Moreover, the entire `data` table must be
+-- updated to use a new enum type (see 22-23 for what's entailed).
+-- For those reasons, we refrain from removing the enum values in here.
--- a/server/schema.sql
+++ b/server/schema.sql
@@ -1,8 +1,10 @@
 CREATE TYPE shot_block_type AS ENUM (
     'none',
-    'dmca'
+    'dmca',
+    'abuse',
+    'usererror',
+    'watchdog'
 );
-ALTER TYPE shot_block_type OWNER TO ianbicking;
 CREATE TABLE accounts (
     id character varying(200) NOT NULL,
     token text,
@@ -58,6 +60,21 @@ CREATE TABLE states (
     state character varying(64) NOT NULL,
     deviceid character varying(200)
 );
+CREATE TABLE watchdog_submissions (
+    id integer NOT NULL,
+    shot_id character varying(270) NOT NULL,
+    request_id character(36) NOT NULL,
+    nonce character(36) NOT NULL,
+    positive_result boolean
+);
+CREATE SEQUENCE watchdog_submissions_id_seq
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
+ALTER SEQUENCE watchdog_submissions_id_seq OWNED BY watchdog_submissions.id;
+ALTER TABLE ONLY watchdog_submissions ALTER COLUMN id SET DEFAULT nextval('watchdog_submissions_id_seq'::regclass);
 ALTER TABLE ONLY accounts
     ADD CONSTRAINT accounts_pkey PRIMARY KEY (id);
 ALTER TABLE ONLY data
@@ -70,6 +87,8 @@ ALTER TABLE ONLY property
     ADD CONSTRAINT property_pkey PRIMARY KEY (key);
 ALTER TABLE ONLY states
     ADD CONSTRAINT states_pkey PRIMARY KEY (state);
+ALTER TABLE ONLY watchdog_submissions
+    ADD CONSTRAINT watchdog_pkey PRIMARY KEY (id);
 CREATE INDEX data_deviceid_idx ON data USING btree (deviceid);
 CREATE INDEX devices_accountid_idx ON devices USING btree (accountid);
 CREATE INDEX images_shotid_idx ON images USING btree (shotid);
@@ -83,4 +102,6 @@ ALTER TABLE ONLY images
     ADD CONSTRAINT images_shotid_fkey FOREIGN KEY (shotid) REFERENCES data(id) ON DELETE CASCADE;
 ALTER TABLE ONLY states
     ADD CONSTRAINT states_deviceid_fkey FOREIGN KEY (deviceid) REFERENCES devices(id) ON DELETE CASCADE;
--- pg-patch version: 20
+ALTER TABLE ONLY watchdog_submissions
+    ADD CONSTRAINT watchdog_shot_id_fkey FOREIGN KEY (shot_id) REFERENCES data(id) ON DELETE CASCADE;
+-- pg-patch version: 24
--- a/server/db-patches/patch-23-24.sql
+++ b/server/db-patches/patch-23-24.sql
@@ -1,0 +1,12 @@
+CREATE TABLE watchdog_submissions (
+  id SERIAL,
+  shot_id CHARACTER VARYING(270) NOT NULL,
+  request_id CHARACTER(36) NOT NULL,
+  nonce CHARACTER(36) NOT NULL,
+  positive_result boolean,
+  CONSTRAINT watchdog_pkey PRIMARY KEY (id),
+  CONSTRAINT watchdog_shot_id_fkey FOREIGN KEY (shot_id)
+      REFERENCES data(id)
+      ON UPDATE NO ACTION
+      ON DELETE CASCADE
+);
--- a/server/db-patches/patch-24-23.sql
+++ b/server/db-patches/patch-24-23.sql
@@ -1,0 +1,1 @@
+DROP TABLE watchdog_submissions;
--- a/server/src/dbschema.js
+++ b/server/src/dbschema.js
@@ -6,7 +6,7 @@ const mozlog = require("./logging").mozlog("dbschema");
 
 // When updating the database, please also run ./bin/dumpschema --record
 // This updates schema.sql with the latest full database schema
-const MAX_DB_LEVEL = exports.MAX_DB_LEVEL = 22;
+const MAX_DB_LEVEL = exports.MAX_DB_LEVEL = 24;
 
 exports.forceDbVersion = function(version) {
   mozlog.info("forcing-db-version", {db: db.constr, version});


server/db-patches/patch-22-23.sql
```diff
@@ -0,0 +1,14 @@
...
+  id SERIAL,
+  shot_id CHARACTER VARYING(270) NOT NULL,
+  request_id CHARACTER(36) NOT NULL,
+  nonce CHARACTER(36) NOT NULL,
+  positive_result boolean,
+  CONSTRAINT watchdog_pkey PRIMARY KEY (id),
+  CONSTRAINT watchdog_shot_id_fkey FOREIGN KEY (shot_id)
+      REFERENCES data(id)
+      ON UPDATE NO ACTION
+      ON DELETE CASCADE
```
6a68: @chenba I'm not sure we should ever delete items from this table, even if the parent shot is deleted. Will the upstream watchdog service preserve everything that would be legally needed in such a case?

server/db-patches/patch-23-22.sql
```diff
@@ -0,0 +1,5 @@
+-- It is not possible to remove a value from an enum type with ALTER.
+-- Moreover, any records that's using the new enum value must be updated
+-- or deleted.  So we are not reverting the shot_blocK_type change here.
```
6a68: It's a little bit involved, but you can remove enum values: https://blog.yo1.dog/updating-enum-values-in-postgresql-the-safe-and-easy-way/

chenba: Yes, I'm aware we could replace the old enum with a new one.  But as stated in the comment, we would still need to delete the shots that's been marked with the watchdog block type (or change to some other value that exist in the new enum) since the new enum wouldn't have that value.

I suppose we try it and let it error out if any shots still using the watchdog block type?

server/db-patches/patch-22-23.sql
chenba: The plan is to not delete the shot until it's been reviewed manually.  As for preserving data for legal reasons, I'll let @clouserw answer.

wresuolc: A watchdog client shouldn't need to hold on to anything if it has confidence in the answer it received.

Since we're just starting this, we want to review the images in case of false positives.  In that case, we should flag and remove from public access, but not delete until we hear back from legal.  In the future, we might just auto-delete, so, let's not make it too complex.  Making it an SQL query we have to run to delete flagged images would be fine.

server/db-patches/patch-22-23.sql
```diff
@@ -0,0 +1,22 @@
+-- pg-patcher runs migrations in transactions. However, ALTER TYPE ... ADD VALUE
+-- cannot be executed inside a transaction block. We will swap out the enum type
+-- with a new one instead.
```
chenba: Turns out we need to swap in a new enum anyway. :worried:

6a68:
Gah! So annoying, my most important comment wasn't included in the review.

@chenba I realized, when looking at the migration script, that changing the column type is going to rewrite the table synchronously, and it's going to need a lock on the whole `data` table while it does that. According to /metrics, we've got ~2M shots, and ~2M deleted, recoverable shots. Linear rewriting of a table with 4M rows, plus indexes, seems likely to cause an outage in production :-(

I looked around for suggestions on what to do here, but haven't found a definitive answer yet. Something like copying the table to a new table in small batches, changing the enum along the way, then swapping out the old and new tables at the end via renaming, seems like it could work, but I'm not sure.

Since this turns out to be a complicated situation, I wonder if it might also be worth replacing that enum with a foreign key to a single-column table containing the values. Might make this specific change easier in the future.

Sorry I didn't catch this on the first pass. I have time tomorrow to look a

server/schema.sql
```diff
@@ -1,8 +1,11 @@
 CREATE TYPE shot_block_type AS ENUM (
     'none',
-    'dmca'
+    'dmca',
+    'abuse',
+    'usererror',
+    'watchdog'
 );
-ALTER TYPE shot_block_type OWNER TO ianbicking;
+ALTER TYPE shot_block_type OWNER TO chenba;
```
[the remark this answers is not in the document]
chenba: :-/ This file is generated.  I don't think we need that line.

chenba:
@6a68 Yeah, at this point any ALTER to `data` or `images` is likely to be painful.  (Although I have no idea how fast the DB server is.)  We could also look into running ALTER TYPE outside of a transaction.  Although that might require forking and patching pg-patcher.

server/db-patches/patch-22-23.sql
```diff
@@ -0,0 +1,22 @@
...
+ALTER TYPE shot_block_type RENAME TO old_shot_block_type;
+CREATE TYPE shot_block_type AS ENUM ('none', 'dmca', 'abuse', 'usererror', 'watchdog');
+ALTER TABLE data ALTER COLUMN block_type DROP DEFAULT;
+ALTER TABLE data ALTER COLUMN block_type TYPE shot_block_type USING block_type::text::shot_block_type;
```
autrilla: This isn't great. Basically it's recreating the entire column by converting the existing value to text and then that text to the new enum type. Because the column is rewritten, the whole row is rewritten, and therefore the whole table. The database has a row size of around 500 bytes, and with 4.2e7 rows, it will take a while. PostgreSQL will take an exclusive lock on the whole table during the migration, so while those 21 GB are being written to disk, that table won't be usable. There are alternatives we could discuss.

Also I might be wrong as I'm certainly not a postgres expert, but testing on my laptop does seem to indicate the full table rewrite is happening.

autrilla: Actually, I don't think we need to discuss this, I found out since PostgreSQL 9.1 there's a strictly better alternative. Since 9.1, you can simply `ALTER TYPE enum ADD VALUE 'foo';`.

Here's a quick benchmark of what I propose compared to what this PR does: https://gist.github.com/autrilla/3837ed08581a30132594e24a7f640a4c

As you can see, `ADD VALUE` is very fast, and this PR indeed rewrites the whole table.

chenba: @autrilla The original patch did use `ALTER TYPE ... ADD VALUE`; see the comment at https://github.com/mozilla-services/screenshots/pull/4417/files#diff-23625b9e1cf673be5214f52029f9531bR1 on why that was changed.  There was also [some discussion in IRC](https://mozilla.logbot.info/screenshots/20180515#c14759281) on potential approaches.

autrilla: My bad, I missed that. Let's discuss it on Monday then!

oremj: We might want to move away from ENUMs if this list is going to change or keep growing.

oremj: We could use a natural foreign key to a table of block types or a check constraint instead.

6a68: Yeah, I agree, though we still need to migrate away from the current implementation without downtime

autrilla: @chenba it's dirty, but what does pg-patcher do if we were to start our migration SQL with

```sql
COMMIT;
ALTER TYPE ... ADD VALUE;
BEGIN;
```

If it's supported, it's probably the simplest way. Other alternatives I can think of:

1. Do what @oremj said, use a table of block types, add a new column that uses that table, and backfill. The application will have to read and write to both columns for the time being.
2. Add the value to the database manually.
3. Patch pg-patcher so we can have migrations that run outside a transaction.

My opinion is that backfilling is going to be very cumbersome.

chenba: @autrilla Postgresql would throw the error: "ALTER TYPE ... ADD cannot be executed from a function or multi-command string"

ianb:
We had some discussion, and came up with a plan for migration:

1. We'll rely on @autrilla manually executing the SQL to update the database (outside of pgpatcher)
2. He will restore a recent snapshot of the production database to test the SQL, and see how long it takes to execute (this restored database won't ever be hooked up to any application, and will be deleted immediately after)
3. The database can be updated at any time before the next server release
4. The database version (`SELECT value FROM property WHERE key = 'patch'`) should be updated at the time of production
5. We still have to confirm that we can update the database version without the application complaining that it is running against an unexpectedly new version of the database

chenba:
I've confirmed that the server will attempt to roll back to the version in `dbschema.js` when the patch level is higher in the db.

In this particular case, the rollback is dropping the watchdog submissions table.  I propose we break up the current migration into two: the special one for adding enum values, which won't have a down script, and one for watchdog submissions.

ianb:
I think the list is good. I'm assuming we use "abuse" for both abusive and illegal content. I proposed just "abuse" because I didn't want the database to include any assertions about the law ;)

Regarding the rollback, can we just make the application never roll back? pgpatcher doesn't seem to want to do that, but maybe we can convince it. We can make the behavior configurable (and turn on rollback for dev and stage), but I think in general that we shouldn't automatically roll back the database in production (backward migrations aren't as well tested, and in practice they are very hard to manage with our current system).

chenba:
pg-patcher doesn't have an option for not running migrations, so I don't think we can convince it. But we could add some code to bypass the `pgpatcher()` call if the patch level in the db is greater than or equal to the one in the code for certain env(s).  That does change the heartbeat endpoint, which check if the two patch levels match.

ianb:
Oh, of course, not running pgpatcher is sensible. Probably that can be a followup to this patch (assuming everything is okay here except the migration concerns). Though we may want to be prepared to backport the conditional pgpatcher invocation, so that we can deploy it before this patch.

autrilla:
On a copy of the current prod instance:

```
pageshot=> \dT+ shot_block_type
                                       List of data types
 Schema |      Name       |  Internal name  | Size | Elements | Access privileges | Description 
--------+-----------------+-----------------+------+----------+-------------------+-------------
 public | shot_block_type | shot_block_type | 4    | none    +|                   | 
        |                 |                 |      | dmca     |                   | 
(1 row)
pageshot=> ALTER TYPE shot_block_type OWNER TO pageshot;
ALTER TYPE
Time: 7.361 ms
pageshot=> \dT+ shot_block_type^C
pageshot=> ALTER TYPE shot_block_type ADD VALUE 'abuse';
ALTER TYPE
Time: 79.705 ms
pageshot=> ALTER TYPE shot_block_type ADD VALUE 'usererror';
ALTER TYPE
Time: 2.997 ms
pageshot=> ALTER TYPE shot_block_type ADD VALUE 'watchdog';
ALTER TYPE
Time: 2.946 ms
pageshot=> \dT+ shot_block_type 
                                       List of data types
 Schema |      Name       |  Internal name  | Size | Elements  | Access privileges | Description 
--------+-----------------+-----------------+------+-----------+-------------------+-------------
 public | shot_block_type | shot_block_type | 4    | none     +|                   | 
        |                 |                 |      | dmca     +|                   | 
        |                 |                 |      | abuse    +|                   | 
        |                 |                 |      | usererror+|                   | 
        |                 |                 |      | watchdog  |                   | 
(1 row)
```

chenba:
I think it'd be better if this is split into two db patches. That way ops can handle the enum changes, and the new table can be a regular db patch.

chenba:
I've split up the enum changes and the new table into two db patches.

autrilla:
I still see the migration SQL creating a new type. Weren't we changing it to use `ALTER TYPE ... ADD VALUE`, or are we making sure the migrations can still be ran by pg-patcher and we'll run a separate migration in production?

chenba:
Production (and I think stage?) won't rely on the enum migration here.  That's gotta be manually done.

@autrilla do you need the SQL for those ALTERs and incrementing the patch level in the property table written down somewhere?  (I haven't been involved with a manual migration liek this either, so I don't know what the exact steps are.)

autrilla:
I can handle adding the enum values, but I have no clue how pg-patcher does things. I can look into it on Monday, unless someone already knows what needs to be done in that regard.

chenba:
pg-patcher is called with a desired patch level when the server starts.  If the level in the argument is different than the one in the property table, then pg-patcher would upgrade or downgrade the db, in one transaction, until the level is reached.  So when the enum values are manually added, the patch level in the property table should be updated (to 23 in this case).  That way pg-patcher would skip patch 23.

6a68:
OK. @chenba @ianb is this good to land? Did we want to add any other items to the enum from the admin app?

ianb:
I think this is good to go, yes

PR #4466, opened 2018-05-16
Fixes #3964 - Download Cancel

### LINKED ISSUE #3964: "Download canceled by the user" error is wrongly displayed when you cancel saving the selection to a specific location
[Notes]:
- This issue is only reproducible if the "Always ask you where to save files" option is checked.
- The issue is not reproducible if the shot is uploaded first.

[Affected versions]:
- Nightly 59.0a1
- Beta 58.0b15

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a profile where the "Always ask you where to save files" option is checked in "about:preferences#general" page.

[Steps to reproduce]:
1. Open Firefox browser with the profile from prerequisites and navigate to any website.
2. Perform a selection on the page.
3. Click the download button.
4. Click the "Cancel" button from the "Save/Save as" window.
5. Observe the behavior.

[Expected result]:
- The download is canceled and no errors are displayed.

[Actual result]:
- "Error: Download canceled by the user" error is displayed.

[Additional Notes]:
- This issue is not applicable for Firefox 57 release, since the [bug 1394851](https://bugzilla.mozilla.org/show_bug.cgi?id=1394851) was marked as "wontfix" for this version.
download canceled by the user

punamdahiya:


punamdahiya:
Thanks @chenba for review. 

@ianb @chenba one quick clarification, with this fix, all downloads error from browser.downloads.download API call will be caught silently and logged. I hope that's the desired behavior and in none of error flows for download API  we would like user to see the 'Firefox Screenshots went haywire..' notification

My findings, there is no easy and robust way to identify Download Canceled error message thrown by browser.downloads.download API.

The properties of error object are [ "fileName", "lineNumber", "columnNumber", "message" ]
with message being "Download canceled by the user". So unless we compare the message strings, it's not possible to separate the download errors thrown

ianb:
We check `error.message` in other places in the code. There have been plans for error codes (which would be implemented as part of WebExtensions) but they haven't and probably won't happen. The error message is somewhat fragile, but... it's all we got!

That said, I think it's fine to catch them all too, though really we don't even want to log cancellations. I personally would have used something like `error && error.message && error.message.includes("canceled")`

PR #4490, opened 2018-05-23
Skip DB downgrades.

### LINKED ISSUE #4484: Conditionally Skip pg-patcher Backward Patches
This is a follow-up to #4376 & #4347 (PR #4417 is where all the relevant comments live).

With this fix we want to, when the patch level in the database is greater than or equal to `MAX_DB_LEVEL` in `dbschema.js`:

- Always skip pgpatcher if `NODE_ENV` is "production"
- Skip pgpatcher if a certain config is set to true

chenba:
Fixes #4484

### DIFF
--- a/bin/build-docker-image
+++ b/bin/build-docker-image
@@ -2,7 +2,8 @@
 
 set -eu
 
-export NODE_ENV=production
+export NODE_ENV=${NODE_ENV:-production}
+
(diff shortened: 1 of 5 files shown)


chenba:
That codecov report is nonsense.

chenba:
Yeah, let's land this first.

PR #4497, opened 2018-05-24
Fixes #4485 - Update copy icon in notification.

### LINKED ISSUE #4485: Replace copy icon in notifications (not Photon style)
The copy icon in the notification is not the correct icon we use in our design guidelines.

Incorrect icon:
image

Correct icon:
image
https://design.firefox.com/icons/viewer/#copy

Also, is it possible to set a size for the icon? Seems awfully big in the notification.

chenba:


6a68:
@sevaan here's the new icon. I find the heavy border jarring, but I guess that's what the design system calls for, so...ship it?

sevaan:
Yeah, ship it. It's bold, but it's certainly eye-catching and hopefully extra communicative!

PR #4499, opened 2018-05-24
Annotation - Add Text Tool

punamdahiya:
It's initial commit. Creating PR for discussion and early feedback

### DIFF
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -185,6 +185,8 @@ annotationUndoButton =
     .title = Undo
 annotationRedoButton =
     .title = Redo
+annotationTextButton =
+    .title = Add text
 # Note: This button reverts all the changes on the image since the start of the editing session.
 annotationClearButton =
     .title = Clear
@@ -217,6 +219,18 @@ annotationColorSeaGreen =
 annotationColorGrey =
     .title = Grey
 
+# Note: annotationTextSize is a title for text size selection dropdown.
+annotationTextSize =
+    .title = Text size
+# Values shown in text size selection dropdown
+textSizeSmall = Small
+textSizeMedium = Medium
+textSizeLarge = Large
+# Confirm and Cancel button title shown when using text tool
+annotationTextConfirmButton =
+    .title = Confirm
+annotationTextCancelButton =
+    .title = Cancel
 
 ## Settings Page
 
--- a/server/src/pages/shot/editor.js
+++ b/server/src/pages/shot/editor.js
@@ -5,6 +5,7 @@ const sendEvent = require("../../browser-send-event.js");
 const { PenTool } = require("./pen-tool");
 const { HighlighterTool } = require("./highlighter-tool");
 const { CropTool } = require("./crop-tool");
+const { TextTool } = require("./text-tool");
 const { ColorPicker } = require("./color-picker");
 const { EditorHistory, RecordType } = require("./editor-history");
 const { Selection } = require("../../../shared/selection");
@@ -17,6 +18,7 @@ exports.Editor = class Editor extends React.Component {
       canvasCssHeight: Math.floor(this.props.clip.image.dimensions.y),
       tool: "",
       color: "",
+      colorName: "",
       lineWidth: "",
       actionsDisabled: true,
       canUndo: false,
@@ -50,6 +52,7 @@ exports.Editor = class Editor extends React.Component {
     this.setState({
       tool: "pen",
       color: "#000",
+      colorName: "black",
       lineWidth: 5,
       actionsDisabled: true
     });
@@ -138,6 +141,18 @@ exports.Editor = class Editor extends React.Component {
           cancelCropHandler={this.onClickCancelCrop.bind(this)}
           confirmCropHandler={this.onCropUpdate.bind(this)}
           toolbarOverrideCallback={this.overrideToolbar.bind(this)} />;
+      case "textTool":
+        return <TextTool
+          ref={this.selectedTool}
+          color={this.state.color}
+          colorName={this.state.colorName}
+          baseCanvas={this.imageCanvas}
+          canvasPixelRatio={this.state.canvasPixelRatio}
+          canvasCssWidth={this.state.canvasCssWidth}
+          canvasCssHeight={this.state.canvasCssHeight}
+          cancelTextHandler={this.onClickCancelText.bind(this)}
+          confirmTextHandler={this.onClickUpdateText.bind(this)}
+          toolbarOverrideCallback={this.overrideToolbar.bind(this)} />;
       default:
         return null;
     }
@@ -181,9 +196,12 @@ exports.Editor = class Editor extends React.Component {
           <Localized id="annotationHighlighterButton">
             <button className={`button transparent highlight-button ${highlighterState}`} id="highlight" onClick={this.onClickHighlight.bind(this)} title="Highlighter"></button>
           </Localized>
+          <Localized id="annotationTextButton">
+            <button className={`button transparent text-button`} id="text" onClick={this.onClickText.bind(this)} title="Text"></button>
+          </Localized>
           <ColorPicker activeTool={this.state.tool}
             setColorCallback={this.setColor.bind(this)}
-            color={this.state.color} />
+            color={this.state.color} colorName = {this.state.colorName}/>
           <span className="annotation-divider"></span>
           <Localized id="annotationUndoButton">
             <button className={`button transparent undo-button ${undoButtonState}`} id="undo"
@@ -210,8 +228,8 @@ exports.Editor = class Editor extends React.Component {
     </div>;
   }
 
-  setColor(color) {
-    this.setState({color});
+  setColor(color, colorName) {
+    this.setState({color, colorName});
   }
 
   renderShotsLoading() {
@@ -359,6 +377,35 @@ exports.Editor = class Editor extends React.Component {
       sendEvent("pen-select", "annotation-toolbar");
     }
   }
+
+  onClickText() {
+    this.previousTool = this.state.tool;
+    this.setState({tool: "textTool"});
+    sendEvent("text-select", "annotation-toolbar");
+  }
+
+  onClickUpdateText(affectedArea, incomingCanvas) {
+    if (!affectedArea || !incomingCanvas) {
+      this.setState({tool: this.previousTool});
+      return;
+    }
+    this.history.pushDiff(this.imageCanvas, affectedArea);
+
+    this.imageContext.globalCompositeOperation = "source-over";
+    this.imageContext.drawImage(incomingCanvas,
+      affectedArea.left * this.state.canvasPixelRatio,
+      affectedArea.top * this.state.canvasPixelRatio,
+      affectedArea.width * this.state.canvasPixelRatio,
+      affectedArea.height * this.state.canvasPixelRatio,
+      affectedArea.left, affectedArea.top, affectedArea.width, affectedArea.height);
+    this.setState({tool: this.previousTool});
+    this.deriveButtonStates();
+  }
+
+  onClickCancelText() {
+    this.setState({tool: this.previousTool});
+  }
+
 
   onMouseUp(e) {
     // This is here so that when the user releases the mouse button outside of
--- a/server/src/pages/shot/text-tool.js
+++ b/server/src/pages/shot/text-tool.js
@@ -1,0 +1,258 @@
+const React = require("react");
+const PropTypes = require("prop-types");
+const { Localized } = require("fluent-react/compat");
+const sendEvent = require("../../browser-send-event.js");
+const { Selection } = require("../../../shared/selection");
+const { ColorPicker } = require("./color-picker");
+
+// Text input height is determined by text size and padding
+// See frame.scss mixin type-box($size)
+const TEXT_INPUT_PADDING = 4;
+
+// Padding used by draggable div
+const DRAG_DIV_PADDING = 10;
+
+// Editor header height in pixel
+const EDITOR_HEADER_HEIGHT = 100;
+
+const FONT_STYLE = "sans-serif";
+const FONT_WEIGHT = 900;
+const INIT_FONT_SIZE = 36;
+
+let previousTextInputWidth;
+let previousInputText = "";
+
+exports.TextTool = class TextTool extends React.Component {
+  constructor(props) {
+    super(props);
+    this.textInput = React.createRef();
+
+    this.canvasCssWidth = props.canvasCssWidth;
+    this.canvasCssHeight = props.canvasCssHeight;
+
+    const VISIBLE_HEIGHT = this.canvasCssHeight > window.innerHeight ? window.innerHeight - EDITOR_HEADER_HEIGHT :
+                                                                       this.canvasCssHeight;
+
+    const INIT_LEFT = (this.canvasCssWidth / 2) - (INIT_FONT_SIZE / 2);
+    const INIT_TOP = window.scrollY + (VISIBLE_HEIGHT / 2) - ((INIT_FONT_SIZE) / 2) - TEXT_INPUT_PADDING;
+
+    this.state = {
+      left: INIT_LEFT,
+      top: INIT_TOP,
+      color: this.props.color,
+      colorName: this.props.colorName,
+      textSize: "large-text"
+    };
+  }
+
+  componentDidMount() {
+    this.textInput.current.focus();
+    previousTextInputWidth = this.textInput.current.clientWidth;
+    this.adjustHeight();
+    if (this.props.toolbarOverrideCallback) {
+      this.props.toolbarOverrideCallback();
+    }
+  }
+
+  componentDidUpdate(oldProps, oldState) {
+    if (oldState.textSize !== this.state.textSize) {
+      this.props.toolbarOverrideCallback();
+      this.adjustHeight();
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.props.toolbarOverrideCallback) {
+      this.props.toolbarOverrideCallback();
+    }
+  }
+
+  adjustHeight() {
+    const styles = window.getComputedStyle(this.textInput.current);
+    const height = parseFloat(styles.lineHeight) + parseFloat(styles.paddingTop) + parseFloat(styles.paddingBottom);
+    this.textInput.current.style.minHeight = `${height}px`;
+  }
+
+  render() {
+    const textToolBox = this.renderTextBox();
+    return <div
+      className={`text-tool-container centered`}
+      onDragOver={this.onDragOver.bind(this)}
+      onDrop={this.onDrop.bind(this)}
+      style={{height: this.props.baseCanvas.style.height, width: this.props.baseCanvas.style.width}}>
+      {textToolBox}
+    </div>;
+  }
+
+  renderTextBox() {
+    const dragDivTopPx = `${this.state.top - DRAG_DIV_PADDING}px`;
+    const dragDivLeftPx = `${this.state.left - DRAG_DIV_PADDING}px`;
+    const dragDivPadding = `${DRAG_DIV_PADDING}px`;
+
+    const dragDivStyles = {
+      position: "absolute",
+      top: dragDivTopPx,
+      left: dragDivLeftPx,
+      padding: dragDivPadding,
+      cursor: "move"
+    };
+
+    return [
+      <div draggable="true" key="drag" onDragStart={this.onDragStart.bind(this)} style={dragDivStyles}>
+        <div id="text-input" ref={this.textInput} contentEditable="true" key="text" onKeyDown={this.onKeyDown.bind(this)}
+          onKeyUp={this.onKeyUp.bind(this)} className={`${this.state.textSize} ${this.state.colorName} text`} onFocus={this.moveCursorToEnd.bind(this)}>
+        </div>
+      </div>
+    ];
+  }
+
+  moveCursorToEnd(e) {
+    if (document.createRange) {
+        const range = document.createRange();
+        range.selectNodeContents(e.target); // Select contents of the element with the range
+        range.collapse(false); // false means collapse to end rather than the start
+        const selection = window.getSelection();
+        selection.removeAllRanges();
+        selection.addRange(range);
+    }
+  }
+
+  renderToolbar() {
+    return <div className="editor-header default-color-scheme"><div className="annotation-tools">
+      <Localized id="annotationTextSize">
+        <select className={`text-select`} title="Text Size" onChange={this.onChangeTextSize.bind(this)} value={this.state.textSize}>
+          <Localized id="textSizeSmall"><option value="small-text">Small</option></Localized>
+          <Localized id="textSizeMedium"><option value="medium-text">Medium</option></Localized>
+          <Localized id="textSizeLarge"><option value="large-text">Large</option></Localized>
+        </select>
+      </Localized>
+      <ColorPicker activeTool={this.state.tool}
+        setColorCallback={this.setColor.bind(this)}
+        color={this.state.color} />
+      <span className="annotation-divider"></span>
+      <Localized id="annotationTextConfirmButton">
+        <button className={`button transparent confirm-text`} id="confirm-text" onClick={this.onClickConfirm.bind(this)} title="Confirm Text">Confirm</button>
+      </Localized>
+      <Localized id="annotationTextCancelButton">
+        <button className={`button transparent cancel-text`} id="cancel-text" onClick={this.onClickCancel.bind(this)} title="Cancel Text">Cancel</button>
+      </Localized>
+    </div></div>;
+  }
+
+  setColor(color, colorName) {
+    this.setState({color, colorName});
+    this.textInput.current.focus();
+  }
+
+  onClickConfirm(e) {
+    // Exit if user doesn't enter any text
+    if (!this.textInput.current.textContent) {
+       if (this.props.cancelTextHandler) {
+          this.props.cancelTextHandler();
+        }
+        sendEvent("cancel-text", "text-toolbar");
+        return;
+    }
+    const styles = window.getComputedStyle(this.textInput.current);
+    const FONT_SIZE = parseInt(styles["font-size"], 10);
+    const x = this.state.left + parseFloat(styles["padding-left"]);
+    const y = this.state.top + TEXT_INPUT_PADDING + parseFloat(styles["line-height"]) / 2;
+
+    const textCanvas = document.createElement("canvas");
+    textCanvas.width = this.props.baseCanvas.width;
+    textCanvas.height = this.props.baseCanvas.height;
+    const drawingContext = textCanvas.getContext("2d");
+
+    drawingContext.scale(this.props.canvasPixelRatio, this.props.canvasPixelRatio);
+    drawingContext.textBaseline = "middle";
+    drawingContext.fillStyle = styles.backgroundColor;
+    drawingContext.fillRect(this.state.left,
+                            this.state.top,
+                            this.textInput.current.clientWidth,
+                            this.textInput.current.clientHeight);
+    drawingContext.fillStyle = styles.color;
+    drawingContext.font = `${FONT_WEIGHT} ${FONT_SIZE}px ${FONT_STYLE}`;
+    drawingContext.fillText(this.textInput.current.textContent, x, y);
+
+    const textSelection = new Selection(this.state.left,
+                                        this.state.top,
+                                        this.state.left + this.textInput.current.clientWidth,
+                                        this.state.top + this.textInput.current.clientHeight);
+
+    if (this.props.confirmTextHandler) {
+      this.props.confirmTextHandler(textSelection, textCanvas);
+    }
+    sendEvent("confirm-text", "text-toolbar");
+  }
+
+  onClickCancel(e) {
+    if (this.props.cancelTextHandler) {
+      this.props.cancelTextHandler();
+    }
+    sendEvent("cancel-text", "text-toolbar");
+  }
+
+  onKeyDown(e) {
+    this.adjustX(e);
+  }
+
+  onKeyUp(e) {
+    // Fix to remove <br> element inserted on press of space bar inside contenteditable div
+    while (this.textInput.current.firstElementChild) {
+      this.textInput.current.removeChild(this.textInput.current.firstElementChild);
+    }
+    this.adjustX(e);
+  }
+
+  adjustX(e) {
+    if (previousInputText === this.textInput.current.textContent) {
+      return;
+    }
+    const rectInput = e.target.getBoundingClientRect();
+    const rectCanvas = this.props.baseCanvas.getBoundingClientRect();
+    const WIDTH_DIFF = this.textInput.current.clientWidth - previousTextInputWidth;
+    this.setState({left: rectInput.left - rectCanvas.left - WIDTH_DIFF / 2});
+    previousTextInputWidth = this.textInput.current.clientWidth;
+    previousInputText = this.textInput.current.textContent;
+  }
+
+  onChangeTextSize(event) {
+    const size = event.target.value;
+    this.setState({textSize: size});
+    this.textInput.current.focus();
+  }
+
+  onDragStart(event) {
+    const style = window.getComputedStyle(event.target);
+    event.dataTransfer.setData("text/plain",
+      (parseFloat(style.getPropertyValue("left")) - event.clientX) +
+       "," + (parseFloat(style.getPropertyValue("top")) - event.clientY));
+  }
+
+  onDragOver(event) {
+    event.preventDefault();
+    return false;
+  }
+
+  onDrop(event) {
+    const offset = event.dataTransfer.getData("text/plain").split(",");
+    const dragLeft = event.clientX + parseFloat(offset[0]);
+    const dragTop = event.clientY + parseFloat(offset[1]);
+
+    this.setState({left: dragLeft + DRAG_DIV_PADDING, top: dragTop + DRAG_DIV_PADDING});
+    event.preventDefault();
+    return false;
+  }
+};
+
+exports.TextTool.propTypes = {
+  baseCanvas: PropTypes.object,
+  color: PropTypes.string,
+  colorName: PropTypes.string,
+  toolbarOverrideCallback: PropTypes.func,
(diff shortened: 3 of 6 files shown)


6a68:
@punamdahiya This looks good to me so far. I'm not sure about the confirm screen for text edits - the UX there seems weird. What if you want to change the font size while you're editing text? How do you edit, save, then reposition / re-start editing?

One of the common idioms here seems to be double-click to re-edit text, for instance, https://pablo.buffer.com. I'm not sure this is possible with the current undo/redo approach, though. @chenba what do you think? Is it possible to have multiple canvases on a 'stage' that can be dragged and dropped / re-edited after a later edit occurs? I'm pretty sure that is not currently possible.

6a68:
Yeah, I'd skip the confirm/cancel screen. Have a look at the notes in the blue box in this part of the spec: https://mozilla.github.io/testpilot-assets/Firefox_Screenshots/Advanced_Annotations/#artboard8

punamdahiya:
@6a68 Thanks for clarifying. With confirm cancel, changing text size while editing is a valid issue. 
I was assuming user can enter multiple text in single save session and was going with confirm cancel  to avoid hitting text tool again before user finishes a single text edit.

I agree , keeping single text edit similar to pablo.buffer.com is much simpler user flow and will resolve changing text size. I assume double click inside the text area or subsequent click on text tool icon in toolbar will highlight and restart editing the text.  Click outside the text area should draw text on canvas. I hope that goes well with undo/redo approach.

chenba:
Sorry for being so late to this...

If we are not going to use a third party editor or library, keep undo/redo, and allow users to re-edit previous annotations (changing text content, stretch or move a drawing, etc.), then we need to change the undo/redo to:

1. Keep a copy of the full image after every edit; or
1. Keep a list of editing commands so we can replay them to reconstruct the state of the image at any point

The first approach is a lot simpler but potentially use a lot of memory.  But we could cap the number of undos.

The second... 😓

punamdahiya:
yes, for now  going with the first approach by implementing all text edit in a separate screen and hit of confirm should save image and push it on history stack to support undo/redo

chenba:
Eh, not sure what I was thinking...probably wasn't: simply saving the image for undo won't work.  Since we need to keep a canvas for each edit, the undo/redo would be performed on the canvas with the edit to undo/redo.

punamdahiya:
familiarizing with the undo/redo flow, I see we are saving canvas for each edits, for RecordType.Frame  used in crop we have special handling of width and height in replay. Since in text edit we aren't changine width and height of Canvas looks like RecordType.DIFF makes more sense

punamdahiya:
Checking in initial commit to text annotation tool, click of text tool icon displays text input at a predefined location on canvas. After user is done entering text, hit of confirm brings user back to main edit screen. Since we are saving Canvas DIFF,  undo/ redo works similar to other drawing tools.

Next Steps:
a) Move cursor in text input and increase text input width as user types in by measuring text width
b) Double click to re-edit text on text edit screen
c) Move text size (and color control ?) to text edit screen toolbar (may need UI spec)
d) Handle Mouse Down + Mouse Move to make text input div draggable
e) Implement handles to change text width

chenba:
@punamdahiya I cannot reproduce the UI shifting bug you mentioned. I've tried with various image sizes.

johngruen:
I swear i added this earlier, but i guess i didn't manage to save it:
demo of all colors/sizes: https://codepen.io/johngruen/full/VdQaKm
demo of default placement: https://codepen.io/johngruen/full/MXQmLg

punamdahiya:
@chenba looks like it was cached code, as I am not able to reproduce it either, will ping with image if I am able to replicate it.

punamdahiya:
@johngruen it was worth the wait, very helpful. Thank you!

johngruen:
@punamdahiya np, there seems to be a weird collision between the `draggable` method and `contenteditable` in my prototype when it comes to text selection and typing.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,236 @@
...
+      ref={this.el}
+      className={`text-tool-container centered`}
+      onMouseDown={this.onMouseDown.bind(this)}
+      style={{height: this.props.baseCanvas.style.height, width: this.props.baseCanvas.style.width}}>
+      {textSelectionBox}
+    </div>;
+  }
+
+  renderTextBox() {
+    if (!this.state.textSelection) {
```
punamdahiya: This is redundant, will remove

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,236 @@
...
+    this.el = React.createRef();
+    this.state = {
+      textSelection: null,
+      textContent: ""
+    };
+    this.canvasWidth = parseInt(props.baseCanvas.style.width, 10);
+    this.canvasHeight = parseInt(props.baseCanvas.style.height, 10);
+  }
+
+  componentDidMount() {
```
chenba: Because `textSelection` is null at this point, the initial `render()` did not render a canvas.  That's why `this.canvas.current` is null here.

punamdahiya: Thanks @chenba - makes sense the life cycle sequence is render -> ComponentDidMount. Moving the initialization of textSelection to ComponentWillMount resolved the issue of this.canvas.current null error

chenba: @punamdahiya `componentWillMount` will be removed, and shouldn't be used; https://reactjs.org/docs/react-component.html#unsafe_componentwillmount

punamdahiya: @chenba Thanks for letting know, as recommended will initialize textSelection state in constructor itself.

punamdahiya:
@chenba Updated PR with latest, few known issues that I would like your input on
a) Handling textSize state. Select dropdown in toolbar is rendered outside TextTool. As of now, value property to display correct selection using textSize state is not working.

b) Texttool uses Selection class to store position of text-input div. Hit of confirm renders text on ‘text-canvas’ using the selection, but with a slight shift.

In addition to above, I am looking into
c) center justify text-input div
d) Select text inside input div using mouse. Thanks!

punamdahiya:
I have handled localization inside TextTool where needed. Please comment if it should work. Thanks

punamdahiya:
Changing textSize after you user start typing flow is acting weird, looking into it.

server/src/pages/shot/editor.js
```diff
@@ -210,8 +232,13 @@ exports.Editor = class Editor extends React.Component {
...
   }
 
-  setColor(color) {
+  setColor(color, colorName) {
     this.setState({color});
+    this.setState({colorName});
+  }
+
+  setTextSize(textSize) {
+    this.setState({textSize});
```
chenba: Is the text size also in the editor because we want it to remember the user's last selection?  It'd be simpler to have only the text size in the text tool state.

punamdahiya: That was the thought but I think it should be ok to have it reset to large for every new text annotation

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,278 @@
...
+                height: this.canvasCssHeight}}>
+      </canvas>,
+
+      <div id="text-input" ref={this.el} contentEditable="true" key="text" onDoubleClick={this.editText.bind(this)} onKeyUp={this.drawText.bind(this)}
+         className={`${this.state.textSize} ${this.state.colorName} text`} style={selectedStyles}>
+      </div>
+    ];
+  }
+
+  renderToolbar() {
```
chenba: The toolbar is rendered by the editor, and the text tool hasn't been updated at that time.  That's why the selected text size is always one behind.  You could try using `componentDidUpdate` to trigger an editor re-render when the text tool's text size has changed.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,278 @@
...
+    const updatedSelection = new Selection(
+      newLeft,
+      newTop,
+      newLeft + this.state.textSelection.width,
+      newTop + this.state.textSelection.height
+    );
+
+    this.setState({textSelection: updatedSelection});
+    return true;
+  }
```
chenba: Is it possible to set `draggable` on `text-input` instead doing this?

server/src/pages/shot/text-tool.js
punamdahiya: chenba: I have updated PR to keep textSize state inside textTool. Need more clarity on how componentDidUpdate works and how to trigger an editor componentDidUpdate from onChange event in TextTool

server/src/pages/shot/text-tool.js
punamdahiya: This worked for use case that's why didn't try Drag and Drop API, I can use and see if it helps to resolve some of the open issues

server/src/pages/shot/text-tool.js
chenba: A react component re-renders when its state changes.  We need the editor to re-render after the text tool has updated, and we can do that by changing the editor's state with a callback.  In the text tool's `componentDidUpdate` (https://reactjs.org/docs/react-component.html#componentdidupdate) you can determine if the `textSize` has changed and call the callback.

This is basically how `overrideToolbar` works; feel free to update it to fit your need.

punamdahiya: perfect makes sense, thanks!

server/src/pages/shot/editor.js
```diff
@@ -181,9 +200,12 @@ exports.Editor = class Editor extends React.Component {
           <Localized id="annotationHighlighterButton">
             <button className={`button transparent highlight-button ${highlighterState}`} id="highlight" onClick={this.onClickHighlight.bind(this)} title="Highlighter"></button>
           </Localized>
+          <Localized id="annotationTextButton">
```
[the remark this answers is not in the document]
punamdahiya: Done

locales/en-US/server.ftl
```diff
@@ -185,6 +185,8 @@ annotationUndoButton =
     .title = Undo
 annotationRedoButton =
     .title = Redo
+annotationTextButton =
+    .title = Text
```
flodolo: Is there a mockup of this feature? I'm having a hard time picturing this tooltip.

Could "Annotate", "Add text" be a better choice?

locales/en-US/server.ftl
```diff
@@ -217,6 +219,18 @@ annotationColorSeaGreen =
...
+    .title = Text Size
+# Values shown in text size selection dropdown
+textSizeSmall = Small
+textSizeMedium = Medium
+textSizeLarge = Large
+# Confirm and Cancel button title shown when using text tool
+annotationTextConfirmButton =
+    .title = Confirm Text
+annotationTextCancelButton =
+    .title = Cancel Text
```
flodolo: These two tooltips seem a bit off. Using simply "Confirm" and "Cancel"?

johngruen: +1

locales/en-US/server.ftl
johngruen: I prefer "Add text" here

chenba:
@punamdahiya I think the initial position of the text input should be the middle of the visible part of the shot or maybe middle of the viewport if that's easier.

chenba:
Found a minor issue where if I type a space character anywhere, then pressing the right arrow key at the end of the input string will _visually_ (but not actually) move the cursor to the beginning of the input string.

chenba:
Hmm once I typed a space character anywhere, pressing the down arrow key also visually (but not actually) move the cursor to the beginning of the input string, and the cursor moves down slightly.

punamdahiya:
@chenba just updated PR with updates to use draggable input, it should resolve the issue seen with arrow keys

punamdahiya:
@chenba latest commit should fix initial position of the text input

locales/en-US/server.ftl
punamdahiya: @flodolo here are links to screenshot of TextTool button in editor and TextTool edit screen seen on click of text button
https://imgur.com/a/SyMIKBv
https://imgur.com/a/n3kWAaO

locales/en-US/server.ftl

chenba:
When I start the text tool, the color is what's selected in the editor.  But if I change the color in the text tool, that selection won't stick when I go back to the editor.  It feels a little inconsistent.  Might be a product/UX question though.

server/src/pages/shot/editor.js
```diff
@@ -51,7 +53,8 @@ exports.Editor = class Editor extends React.Component {
       tool: "pen",
       color: "#000",
       lineWidth: 5,
-      actionsDisabled: true
+      actionsDisabled: true,
+      colorName: "black"
```
chenba: so nit: but perhaps move this line up next to the `color` one?

server/src/pages/shot/editor.js
```diff
@@ -210,8 +228,9 @@ exports.Editor = class Editor extends React.Component {
     </div>;
   }
 
-  setColor(color) {
+  setColor(color, colorName) {
     this.setState({color});
+    this.setState({colorName});
```
chenba: Could combine with line above into `this.setState({color, colorName});`

server/src/pages/shot/editor.js
```diff
@@ -360,6 +379,36 @@ exports.Editor = class Editor extends React.Component {
...
+
+  onClickUpdateText(affectedArea, incomingCanvas, compositeOp) {
+    if (!affectedArea || !incomingCanvas) {
+      this.setState({tool: this.previousTool});
+      return;
+    }
+
+    this.history.pushDiff(this.imageCanvas, affectedArea);
+
+    this.imageContext.globalCompositeOperation = (compositeOp || "source-over");
```
chenba: I think it's always `source-over`.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,218 @@
...
+    this.canvasCssWidth = props.canvasCssWidth;
+    this.canvasCssHeight = props.canvasCssHeight;
+
+    const INIT_LEFT = (this.canvasCssWidth / 2) - INIT_FONT_SIZE * TEXT_INPUT_PADDING_LEFT;
+    const INIT_TOP = (this.canvasCssHeight / 2) - ((INIT_FONT_SIZE * TEXT_INPUT_HEIGHT) / 2);
+
+    this.state = {
+      left: INIT_LEFT,
+      top: INIT_TOP,
+      textContent: "",
```
chenba: I don't think we need to keep the input text content in the component's state.  That text will always be in the element accepting the input.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,218 @@
...
+      padding: dragDivPadding,
+      cursor: "move"
+    };
+
+    return [
+      <canvas ref={this.canvas} id="text-canvas" key="base" width={this.props.baseCanvas.width}
+        height={this.props.baseCanvas.height}
+        style={{width: this.canvasCssWidth,
+                height: this.canvasCssHeight}}>
+      </canvas>,
```
chenba: We don't need this canvas in the DOM since we aren't really using to display anything.  The canvas can be created on confirmation.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,218 @@
...
+      </Localized>
+      <Localized id="annotationTextCancelButton">
+        <button className={`button transparent cancel-text`} id="cancel-text" onClick={this.onClickCancel.bind(this)} title="Cancel Text">Cancel</button>
+      </Localized>
+    </div></div>;
+  }
+
+  setColor(color, colorName) {
+    this.setState({color});
+    this.setState({colorName});
```
chenba: These two `setState()` can be combined into one call.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,218 @@
...
+  drawText(e) {
+    this.setState({textContent: e.target.textContent});
+  }
+
+  onClickConfirm(e) {
+    const styles = window.getComputedStyle(this.textInput.current);
+
+    const BKG_COLOR = styles["background-color"];
+    const FONT_COLOR = styles.color;
+    const FONT_SIZE = parseInt(styles["font-size"].replace(/px/i, ""), 10);
```
chenba: The regex is unnecessary.  `parseInt` will return the correct value.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,218 @@
...
+    this.drawingContext.font = `${FONT_WEIGHT} ${FONT_SIZE}px ${FONT_STYLE}`;
+    this.drawingContext.fillText(this.state.textContent, x, y);
+
+    const textSelection = new Selection(this.state.left,
+                                        this.state.top,
+                                        this.state.left + this.textInput.current.clientWidth,
+                                        this.state.top + this.textInput.current.clientHeight);
+
+    if (this.props.confirmTextHandler) {
+      this.props.confirmTextHandler(textSelection, this.canvas.current, this.globalCompositeOperation || "source-over");
```
chenba: `this.globalCompositeOperation` doesn't exist so that's always "source-over".

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,218 @@
...
+    return false;
+  }
+
+  onDrop(event) {
+    const offset = event.dataTransfer.getData("text/plain").split(",");
+    const dragLeft = event.clientX + parseFloat(offset[0]);
+    const dragTop = event.clientY + parseFloat(offset[1]);
+
+    this.setState({left: dragLeft + DRAG_PADDING});
+    this.setState({top: dragTop + DRAG_PADDING});
```
chenba: These two calls can be combined into one.

static/css/frame.scss
```diff
@@ -605,3 +604,146 @@ body {
   background-repeat: no-repeat;
   background-position: center;
 }
+
+// CSS used for Annotation Text Tool
+.text-select {
+  -moz-appearance: none; //sass-lint:disable-line no-vendor-prefixes
+  appearance: none;
+  background-color: $light-default;
+  background-image: url("/static/img/icon-dropdown.svg");
```
chenba: A relative path would be better here.

chenba:
The main concern is that it's not WYSIWYG when the width of the text is wider than the shot, since the tail of the input is shown as the user types but the head is what drawn onto the shot when confirm is clicked.

server/src/pages/shot/editor.js
punamdahiya: +1

server/src/pages/shot/editor.js
punamdahiya: +1

server/src/pages/shot/editor.js
punamdahiya: Updated to remove compositeOp

server/src/pages/shot/text-tool.js
punamdahiya: Done

server/src/pages/shot/text-tool.js
punamdahiya: It helped to debug by looking at canvas in DOM, but I agree we don't need it , moved it to create inside confirmClick

server/src/pages/shot/text-tool.js
punamdahiya: +1

server/src/pages/shot/text-tool.js
punamdahiya: +1

static/css/frame.scss
punamdahiya: Done

static/img/annotation-text.svg
```diff
@@ -0,0 +1,22 @@
...
+                    </g>
+                    <g id="text" transform="translate(3.375000, 1.687500)" fill="#3E3D40" fill-rule="nonzero">
+                        <polygon id="Shape" points="0 3.375 0 6.75 6.75 6.75 6.75 21.9375 10.125 21.9375 10.125 6.75 16.875 6.75 16.875 3.375"></polygon>
+                        <polygon id="Shape" points="21.9375 1.6875 21.9375 0 16.875 0 16.875 1.6875 18.5625 1.6875 18.5625 21.9375 16.875 21.9375 16.875 23.625 21.9375 23.625 21.9375 21.9375 20.25 21.9375 20.25 1.6875"></polygon>
+                    </g>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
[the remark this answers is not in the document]
punamdahiya: +1 good to know

punamdahiya:
@chenba I have updated PR with your feedback, please review.  I agree when the width of text is wider than the shot that’s a bug and need to be fixed.

@johngruen  can you please help answer question here https://github.com/mozilla-services/screenshots/pull/4499#issuecomment-400850815 Thanks!

locales/en-US/server.ftl
```diff
@@ -185,6 +185,8 @@ annotationUndoButton =
     .title = Undo
 annotationRedoButton =
     .title = Redo
+annotationTextButton =
+    .title = Add Text
```
flodolo: Just noticed that other `title` attributes use sentence case, so this should probably be "Add text"? (likely on @johngruen to answer)

locales/en-US/server.ftl
```diff
@@ -217,6 +219,18 @@ annotationColorSeaGreen =
 annotationColorGrey =
     .title = Grey
 
+# Note: annotationTextSize is a title for text size selection dropdown.
+annotationTextSize =
+    .title = Text Size
```
flodolo: Same question here: "Text size"?

locales/en-US/server.ftl
johngruen: yep

locales/en-US/server.ftl
johngruen: yep

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,205 @@
...
+      padding: dragDivPadding,
+      cursor: "move"
+    };
+
+    return [
+      <div draggable="true" key="drag" onDragStart={this.onDragStart.bind(this)} style={dragDivStyles}>
+        <div id="text-input" ref={this.textInput} contentEditable="true" key="text"
+           className={`${this.state.textSize} ${this.state.colorName} text`} >
+        </div>
+      </div>
```
chenba: We should change the initial position of this.  If the user has a very long shot, and he scrolls to the spot where he wants to add some text, this will cause a jump away from the user's desired position.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,205 @@
...
+
+    this.canvasCssWidth = props.canvasCssWidth;
+    this.canvasCssHeight = props.canvasCssHeight;
+
+    const INIT_LEFT = (this.canvasCssWidth / 2) - INIT_FONT_SIZE * TEXT_INPUT_PADDING_LEFT;
+    const INIT_TOP = (this.canvasCssHeight / 2) - ((INIT_FONT_SIZE * TEXT_INPUT_HEIGHT) / 2);
+
+    this.state = {
+      left: INIT_LEFT,
+      top: INIT_TOP,
```
chenba: These two values do not need to be in the component state, especially since they never change.

static/css/frame.scss
```diff
@@ -605,3 +604,146 @@ body {
...
+$purple: rgb(142, 68, 173);
+$sea-green: rgb(26, 188, 156);
+$grey: rgb(52, 73, 94);
+
+$bg-black: rgba(0, 0, 0, 0.6);
+$bg-white: rgba(255, 255, 255, 0.7);
+
+@mixin type-box($size) {
+  height: $size * 1.3;
+  padding: 4px $size * 0.4 0;
```
chenba: The text and cursor are not middle aligned because of this top padding.  Why do we need it?

server/src/pages/shot/text-tool.js
chenba: Ah sorry, I see where they are updated now.  Please ignore ⬆️

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,205 @@
...
+      </Localized>
+    </div></div>;
+  }
+
+  setColor(color, colorName) {
+    this.setState({color, colorName});
+    this.textInput.current.focus();
+  }
+
+  onClickConfirm(e) {
```
chenba: If the user did not enter any text, we probably shouldn't save anything onto the canvas.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,205 @@
...
+    drawingContext.scale(this.props.canvasPixelRatio, this.props.canvasPixelRatio);
+    drawingContext.textBaseline = "top";
+    drawingContext.fillStyle = `${BKG_COLOR}`;
+    drawingContext.fillRect(this.state.left,
+                            this.state.top,
+                            this.textInput.current.clientWidth,
+                            this.textInput.current.clientHeight);
+    drawingContext.fillStyle = `${FONT_COLOR}`;
+    drawingContext.font = `${FONT_WEIGHT} ${FONT_SIZE}px ${FONT_STYLE}`;
+    drawingContext.fillText(this.textInput.current.textContent, x, y);
```
chenba: Looks like the vertical shift on confirm is from the difference between the height and line-height of the text input div.

locales/en-US/server.ftl

server/src/pages/shot/text-tool.js
punamdahiya: @chenba Is the ask here to change initial position to middle of viewport, cannot replicate the jump away from desired position part when user scrolls on a very long shot

punamdahiya: Updated to change initial position to middle of viewport for a very long shot

static/css/frame.scss
punamdahiya: Used to align text content when rendered in canvas, updated textBaseline to middle and removed this extra padding on top

server/src/pages/shot/text-tool.js
punamdahiya: Seeing this subtle shift when text is rendered on Canvas for large and small text size, could slight difference in y position of fillText be causing this?

punamdahiya:
@chenba Updated PR with feedback on usability changes.  Last few known issues that needs to be figured 
- Subtle vertical shift on confirm for textsize large and small
- To display absolute position text Input as flex to center justify content

server/src/pages/shot/text-tool.js
punamdahiya: Done

locales/en-US/server.ftl
chenba: @punamdahiya ⬆️

punamdahiya:
@chenba Updated PR to handle scroll position, vertical shift seen with large and small fonts and justify-center the text input content.

With these fixes in, PR looks good. Please review. Thanks!

server/src/pages/shot/text-tool.js
punamdahiya: Fixed by explicitly handling vertical shift for respective fonts

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,260 @@
...
+    const y = this.state.top + BUFFER_VERTICAL + parseFloat(styles.height) / 2;
+
+    const textCanvas = document.createElement("canvas");
+    textCanvas.width = this.props.baseCanvas.width;
+    textCanvas.height = this.props.baseCanvas.height;
+    const drawingContext = textCanvas.getContext("2d");
+
+    drawingContext.scale(this.props.canvasPixelRatio, this.props.canvasPixelRatio);
+    drawingContext.textBaseline = "middle";
+    drawingContext.fillStyle = `${BKG_COLOR}`;
```
chenba: Could just use `styles.backgroundColor` (or `styles["background-color"]` if you prefer) here instead assigning it into a variable and doing string interpolation.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,260 @@
...
+    drawingContext.fillRect(this.state.left,
+                            this.state.top,
+                            this.textInput.current.clientWidth,
+                            this.textInput.current.clientHeight);
+    drawingContext.fillStyle = `${FONT_COLOR}`;
```
chenba: Same here.  The interpolation is unnecessary, and so is assigning into `FONT_COLOR`.

chenba:
- If I type some text and then change the font color, the cursor is moved to the beginning of the input.
- The input div shifts to teh left if I type some text and then press the left arrow key.
- The text is still shifted vertically on confirm.
- After entering some text, and then pressing space bar, the cursor end up in the padding of the div. (See first image below)
- At its max width, the div is not center.  (See second image below)

image
image

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,260 @@
...
+
+    // For large and small font increase y position by BUFFER_VERTICAL pixel
+    // to fix vertical shift seen when rendering text on canvas
+    let BUFFER_VERTICAL = 0;
+
+    if (FONT_SIZE === 36) {
+      BUFFER_VERTICAL = 1;
+    } else if (FONT_SIZE === 16) {
+      BUFFER_VERTICAL = 0.5;
+    }
```
chenba: Instead of setting a line-height in the div and then doing these hard coded shifts, leave the line-height at the default value, and then add the difference between the height and line-height to the `y`.

punamdahiya:
Updated PR with below fixes
- Cursor position on color change
- The input div shifts to teh left if I type some text and then press the left arrow key.
- At its max width, the div is not center.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  onKeyUp(e) {
+    // Update left state to left position of text input
+    const rectInput = e.target.getBoundingClientRect();
+    const rectCanvas = this.props.baseCanvas.getBoundingClientRect();
+    const WIDTH_DIFF = this.textInput.current.clientWidth - previousTextInputWidth;
+    this.setState({left: rectInput.left - rectCanvas.left - WIDTH_DIFF / 2});
+    previousTextInputWidth = this.textInput.current.clientWidth;
```
chenba: I think this would be better in onkeydown for the times when a user hold down a key to insert or hold down backspace to delete text.

punamdahiya: this.textInput.current.clientWidth inside keydown doesn't change and  gets updated to clientWidth with new inserted character only onkeyUp

punamdahiya: By this do you mean moving just line 224 in onKeyDown, that could work

chenba:
If the input div is limited by the width of the shot, then maybe we should not allow the user to drag the div partially outside of the shot.  Otherwise, if the user drag the input partially outside of the shot to the left, it'd appear that the input abruptly stop in the shot, before reaching the right edge.

chenba:
I cannot reproduce it consistently, but occasionally the input end up on a second line.  See image below for an example.  In that instance, I typed, dragged the div to move it, and typed some more.

image

chenba:
Found a way to reproduce the line break issue:

1. Type something, include at least one space character
1. Focus another another window (doesn't seem to matter Firefox or some other program)
1. Focus window with annotations editor
1. Type something

punamdahiya:
@chenba updated PR with vertical shift and line break fixes, please review. Thanks!

static/css/frame.scss
```diff
@@ -605,3 +604,147 @@ body {
...
+$sea-green: rgb(26, 188, 156);
+$grey: rgb(52, 73, 94);
+
+$bg-black: rgba(0, 0, 0, 0.6);
+$bg-white: rgba(255, 255, 255, 0.7);
+
+@mixin type-box($size) {
+  padding: 4px $size * 0.5;
+  font-size: $size;
+  height: $size + 8;
```
chenba: Bummer that we need to set this so that the empty div looks ok. (Definitely not an issue if a text input is used instead of a contenteditable div.)  The values here are actually too small; the cursor and text are not middle aligned.  But let's not set it here this way.  Once the input div is rendered, set its height to its line-height plus top and bottom padding.

punamdahiya:
@chenba clarifying what you mean by out of view, code in PR is restricting text content to canvas width and should not ? May be screenshot will help clarify issue here. If this is linux specific, I am trying to set both windows and linux enviroment to test the feature cross platform.

server/src/pages/shot/text-tool.js
```diff
@@ -0,0 +1,257 @@
...
+    if (this.props.cancelTextHandler) {
+      this.props.cancelTextHandler();
+    }
+    sendEvent("cancel-text", "text-toolbar");
+  }
+
+  onKeyDown(e) {
+    // Allow backspace to delete text
+    const key = e.keyCode || e.charCode;
+    if ( key === BACKSPACE_KEYCODE || key === DELETE_KEYCODE ) {
```
chenba: Nit: This works, but according to MDN, we should use `KeyboardEvent.key` if that's available.

chenba:
Here's a gif to illustrate.

hold-to-shift

punamdahiya:
yeah, that's not seen on mac, were you able to replicate this on mac?

chenba:
I haven't tested it on a mac, but I'll take your word for it.  But I must reiterate that the user need to _hold down_ a key to reproduce this bug.

chenba:
When the input div is at its width limit (that is, user cannot enter any more characters), using the left arrow to move the cursor does not work.

chenba:
@punamdahiya Since I'm in an earlier time zone and we are trying to push out the annotations features, so in the name of expediency, I tinkered with some stuff.  You can see the changes at https://github.com/punamdahiya/screenshots/pull/1.  Please review it, merge or cherry-pick whatever works.  The text still shifts on confirm though.

punamdahiya:
@chenba Thanks for fixing hold down key for windows and linux. I have your fixes merged with a minor change to remove rounding  while calculating final x and y position for rendering text to canvas, it helped to tone down shift a little bit.

I see with your PR we are allowing free flow of text, it looks good, will have QA test out more thoroughly.

chenba:
I added the rounding because decimals produce blurry text.  Is this not true on MacOS?

punamdahiya:
Tested on windows and mac and its fine , infact on windows it helped with horizontal shift seen for x position, my unix box not cooperating, haven'tr tried it on unix yet

chenba:
I don't see it now.  Must've something else I was trying at the time.  👍

Issue #4486, opened 2018-05-22
./bin/run-addon doesn't create profile on first run


6a68:
Switched laptops this week, and `./bin/run-addon` doesn't generate the profile. Instead, it prints out the help text for the `open` command. I'm on High Sierra 10.13.4. What am I missing? Nightly is set as the default browser.

I worked around this by manually opening the Profile Manager, creating a new profile named ScreenshotsTestProfile, and saving it in a new /Profile folder inside the repo.

6a68:
The cause of this bug is the naming shift from "FirefoxNightly.app" to "Firefox Nightly.app" on mac.

chenba:
Closing this since #4508 is merged.

PR #4508, opened 2018-05-29
Correctly loop over possible firefoxes

6a68:
@ianb Think I've got it this time. Mind taking a look?

### DIFF
--- a/bin/run-addon
+++ b/bin/run-addon
@@ -10,12 +10,12 @@ nodemon_pid=
 
 binary=
 firefoxes=(
+"$(which firefox || true)"
 "/Applications/FirefoxNightly.app"
 "/Applications/Firefox Nightly.app"
 "/Applications/FirefoxDeveloperEdition.app"
 "/Applications/Firefox Developer Edition.app"
 "/Applications/FirefoxAurora.app"
-"$(which firefox || true)"
 )
 setup_profile=
 server=
--- a/bin/run-addon
+++ b/bin/run-addon
@@ -9,20 +9,20 @@ jpm="$base/node_modules/.bin/jpm"
 nodemon_pid=
 
 binary=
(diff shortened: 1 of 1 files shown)


6a68:
Hmm, but this does not fix #4486

6a68:
Now fixes #4486 too

bin/run-addon
```diff
@@ -9,20 +9,20 @@ jpm="$base/node_modules/.bin/jpm"
...
-$(which firefox || true)
-"
+firefoxes=(
+"/Applications/FirefoxNightly.app"
+"/Applications/Firefox Nightly.app"
+"/Applications/FirefoxDeveloperEdition.app"
+"/Applications/Firefox Developer Edition.app"
+"/Applications/FirefoxAurora.app"
+"$(which firefox || true)"
+)
```
chenba: It probably makes more sense to have `firefox` first?  That way I can change wherever `firefox` is symlink'd to run a different version of Firefox.

6a68:
@chenba updated. r?

bin/run-addon
```diff
@@ -94,7 +94,7 @@ done
 open=$binary # assumes Linux
 unamestr=$(uname)
 if [[ "$unamestr" = "Darwin" ]]; then
-  open="open $binary"
+  open=(open "${binary}")
```
chenba: I don't quite understand why this needs to be an array.  Otherwise LGTM.

6a68: I'm not much of a bash scripter. Suggestions welcome :-)

chenba: The older code of having `open` as a string should work as far as I can tell.

chenba: Oh, the spaces.  :satisfied:

PR #4517, opened 2018-06-02
Fix #3160, remove app startup delay

### LINKED ISSUE #3160: Can we remove the delayed bootstrap startup yet?
Now that OOP webextensions have landed, we might be able to remove the code that waits to init screenshots until the 'sessionstore-windows-restored' event fires.

Let's remove that code, push to Talos, and see if any regressions are introduced.

6a68:
The 'sessionstore-windows-restored' delay is now applied to all
extensions by the addon management code in Firefox. See bug 1457224
for details.

We still invoke `appStartupDone()` to avoid shutting down before startup is complete (see uses of `appStartupPromise`).

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -44,22 +44,6 @@ const prefObserver = {
   }
 };
 
-
-const appStartupObserver = {
-  register() {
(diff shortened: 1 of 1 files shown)


chenba:
That CI failure is from an eslint error.

PR #4529, opened 2018-06-06
Submit shot image to Watchdog on save.

### LINKED ISSUE #4377: [Watchdog] Send Submission Request on Shot Saves
When a shot is saved, send a submission request to Watchdog.

chenba:
Fixes #4377

### DIFF
--- a/server/src/watchdog.js
+++ b/server/src/watchdog.js
@@ -1,0 +1,140 @@
+const config = require("./config").getProperties();
+const mozlog = require("./logging").mozlog("server");
+const { captureRavenException } = require("./ravenclient");
+const db = require("./db");
+const { Shot } = require("./servershot");
+const util = require("util");
+const { URL } = require("url");
+const fetch = require("node-fetch");
+const Hawk = require("hawk");
+const FormData = require("form-data");
+const genUuid = require("nodify-uuid");
+
+const isEnabled = config.watchdog.enable;
+validateConfiguration();
+
+function validateConfiguration() {
+  if (!isEnabled) {
+    return;
+  }
+
+  let isConfigValid = true;
+
+  if (!config.watchdog.id || !config.watchdog.key || !config.watchdog.algorithm) {
+    mozlog.warn("watchdog-invalid-config", {msg: "Watchdog credentials configuration values are missing."});
+    isConfigValid = false;
+  }
+
+  if (!config.watchdog.submissionUrl) {
+    mozlog.warn("watchdog-invalid-config", {msg: "Watchdog submission URL configuration value is missing."});
+    isConfigValid = false;
+  } else {
+    try {
+      new URL(config.watchdog.submissionUrl);
+    } catch (e) {
+      mozlog.warn("watchdog-invalid-config", {msg: "Watchdog submission URL is not a valid URL.", err: e});
+      isConfigValid = false;
+    }
+  }
+
+  if (!isConfigValid) {
+    mozlog.info("watchdog-config-exit", {msg: "Watchdog is enabled but not properly configured."});
+    process.exit(2);
+  }
+}
+
+let positiveEmailList;
+if (config.watchdog.positiveEmail) {
+  positiveEmailList = config.watchdog.positiveEmail.split(";");
+}
+
+let intervalSize = config.watchdog.submissionInterval;
+if (intervalSize <= 1) {
+  intervalSize = 1;
+}
+let odometer = 0;
+function isAtInterval() {
+  if (intervalSize === 1) {
+    return true;
+  }
+
+  const atInterval = odometer === 0;
+  odometer = (odometer + 1) % intervalSize;
+  return atInterval;
+}
+
+function shouldSubmitShot() {
+  return isEnabled && isAtInterval();
+}
+
+function getSubmissionId() {
+  return db.select(
+    "SELECT nextval(pg_get_serial_sequence('watchdog_submissions', 'id')) AS new_id",
+  ).then(rows => {
+    if (rows.length) {
+      return rows[0].new_id;
+    }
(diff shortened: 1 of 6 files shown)


server/src/watchdog.js
```diff
@@ -0,0 +1,141 @@
...
+      filename: imageId,
+      contentType: obj.contentType
+    });
+
+    return getSubmissionId();
+  }).then(submissionId => {
+    const uuid = util.promisify(genUuid.generate);
+    return uuid(genUuid.V_RANDOM).then(nonce => ({submissionId, nonce}));
+  }).then(({submissionId, nonce}) => {
+    const callbackUrl = `${shot.backend}/watchdog/${submissionId}?nonce=${nonce}`;
```
chenba: The match result is in the request body from Watchdog, so we don't need unique callback urls.

server/src/watchdog.js
```diff
@@ -0,0 +1,141 @@
...
+      `INSERT INTO watchdog_submissions (id, shot_id, request_id, nonce)
+       VALUES ($1, $2, $3, $4)`,
+      [submissionId, shot.id, request_id, nonce.toString()]
+    );
+  }).catch(e => {
+    mozlog.error("watchdog-failed-submission", {err: e});
+  });
+}
+
+exports.Watchdog = {
```
ianb: This singleton seems a little odd. Why not just expose `submit` directly? Or, I could imagine taking all these functions and wrapping them up in a class instead of using global variables, and then making an instance of that class given the configuration, and exporting that instance.

server/src/watchdog.js
```diff
@@ -0,0 +1,141 @@
...
+      .then(res => res.json())
+      .then(respJson => ({submissionId, nonce, request_id: respJson.id}));
+  }).then(({submissionId, nonce, request_id}) => {
+    return db.insert(
+      `INSERT INTO watchdog_submissions (id, shot_id, request_id, nonce)
+       VALUES ($1, $2, $3, $4)`,
+      [submissionId, shot.id, request_id, nonce.toString()]
+    );
+  }).catch(e => {
+    mozlog.error("watchdog-failed-submission", {err: e});
```
ianb: We should probably make sure something is logged to Sentry as well.

server/src/watchdog.js
```diff
@@ -0,0 +1,141 @@
...
+function submit(shot) {
+  if (!shouldSubmitShot()) {
+    return;
+  }
+
+  const imageUrl = shot.getClip(shot.clipNames()[0]).image.url;
+  const imageId = imageUrl.split("/").pop();
+  const form = new FormData();
+
+  Shot.getRawBytesForClip(
```
ianb: This would all be a lot nicer with async/await... though I suppose there's coverage concerns. Seems a shame, though.

server/src/watchdog.js
```diff
@@ -0,0 +1,141 @@
...
+  } else {
+    try {
+      new URL(config.watchdog.submissionUrl);
+    } catch (e) {
+      mozlog.warn("watchdog-invalid-config", {msg: "Watchdog submission URL is not a valid URL.", err: e});
+      isConfigValid = false;
+    }
+  }
+
+  return isConfigValid;
```
ianb: If the config seems incomplete, then it seems like that should be a fatal errors. Other setup issues are treated as fatal errors (including using process.exit if necessary), and that seems to work well.

server/src/watchdog.js
```diff
@@ -0,0 +1,141 @@
...
+let positiveEmailList;
+if (config.watchdog.positiveEmail) {
+  positiveEmailList = config.watchdog.positiveEmail.split(";");
+}
+
+let denominator = config.watchdog.throttleDenominator;
+if (denominator <= 1) {
+  denominator = 1;
+}
+let odometer = 0;
```
ianb: It would be nice if these were a little clearer. Maybe denominator and odometer themselves could be a class? Like, after looking I can see that denominator means "every {denominator} shots, actually submit", but it's not obvious to me.

And below `shouldSubmitShot()` has side-effects that one wouldn't expect from its name (in that it's maintaining and incrementing a counter).

server/src/watchdog.js
chenba: Do people look at or find those coverage reports helpful?

6a68: >  Do people look at or find those coverage reports helpful?

I don't, but we could make test coverage a code quality theme for Q3 or Q4, and then it'd be a useful first step.

server/src/watchdog.js
6a68: Is there any reason to keep a counter at all? Seems simpler to just do a coin flip each time:

"if Math.random() * N < 1, send to watchdog"

chenba: A counter allows for a strict rate limit.  We didn't define a rate for this in Screenshots, so either approach is probably ok (with the probability approach being simpler).  It's also meant to be temporary because we do want to send 100% of the shots to Watchdog eventually.

server/src/watchdog.js
ianb: I don't look at them personally. We're kind of catching up with tests, but I don't think we were in a bad place before, which is to say we aren't catching up with *quality*. So... I don't know. But I'm also leaning away from Screenshots, so my opinion on the matter isn't as important as y'all's.

test/server/test_file_management.py
```diff
@@ -142,6 +142,7 @@ def restart_server(**extra_env):
     env['DEBUG_GOOGLE_ANALYTICS'] = ''
     env['NODE_ENV'] = 'production'
     env['LOG_QUERY_LIMIT'] = '0'
+    env['ENABLE_WATCHDOG'] = 'false'
```
chenba: Need to set this because the default is true but the server would halt because other config values do not exist.

Issue #4379, opened 2018-04-25
[Watchdog] Add Endpoint for Handling Positive Responses


chenba:
At the very least:

- make the image inaccessible
- flag the shot appropriately (use a block type from #4347 or create a new one?)

PR #4532, opened 2018-06-07
Handle Watchdog Callback Requests

### LINKED ISSUE #4378: [Watchdog] Add Endpoint for Handling Negative Responses
Record it.  That's it.

chenba:
Fix #4378, #4379 

Note that #4529 should be merged before this.

### DIFF
--- a/server/src/watchdog.js
+++ b/server/src/watchdog.js
@@ -78,6 +78,19 @@ function getSubmissionId() {
   });
 }
 
+function getSubmission(id) {
+  return db.select(
+    "SELECT * FROM watchdog_submissions WHERE id = $1", [id]
+  ).then(rows => {
+    if (rows.length) {
+      return rows[0];
+    }
+    const err =  new Error("Unable to find a Watchdog submission.");
+    err.watchdogSubmissionId = String(id);
+    throw err;
+  });
+}
+
 const credentials = {
   id: config.watchdog.id,
   key: config.watchdog.key,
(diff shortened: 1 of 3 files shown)


server/src/watchdog.js
```diff
@@ -138,3 +149,45 @@ exports.submit = function(shot) {
     mozlog.error("watchdog-failed-submission", {err: e});
   });
 };
+
+exports.handleResult = function(req) {
+  getSubmission(req.params.submissionId).then(record => {
+    if (record.request_id !== req.body.watchdog_id) {
+      throw new Error(`Received mismatching Watchdog ID in callback for submission ${record.id}.`);
```
ianb: Sentry doesn't do well with dynamically-generated error messages.

This pattern works in the add-on, but I'm unsure if it works on the server:

```js
let exc = new Error("Received mismatching Watchdog ID in callback");
exc.recordId = String(record.id); // Note, if you don't turn it into a string then "undefined" gets lost
exc.watchDogId = String(req.body.watchdog_id);
```

But... I'm really not sure if that works.

server/src/watchdog.js
```diff
@@ -138,3 +149,45 @@ exports.submit = function(shot) {
...
+    }
+    if (req.query.nonce !== record.nonce) {
+      throw new Error(`Received mismatching nonce in callback for submission ${record.id}.`);
+    }
+    if (record.positive_result !== null) {
+      throw new Error(`Watchdog submission ${record.id} already has a result.`);
+    }
+
+    if (req.body.positive === false) {
```
ianb: `if (req.body.positive)` ? Unless there's some weird in-between where it's null...?

server/src/watchdog.js
```diff
@@ -138,3 +149,45 @@ exports.submit = function(shot) {
...
+      handleNegative(record);
+    } else {
+      handlePositive(record);
+    }
+  }).catch(e => {
+    mozlog.error("watchdog-failed-callback", {err: e});
```
ianb: This should probably use `responses.sendError(req, res, e)`, or else the catch should happen in `server.js`. Or you should use `ravenclient.captureRavenException`

PR #4535, opened 2018-06-07
Fix #4528, Pass addonData through to embedded WebExtension startup

### LINKED ISSUE #4528: Include upstream PDF-related patch
In [Bug 1466349](https://bugzilla.mozilla.org/show_bug.cgi?id=1466349) there's a patch to Screenshots that we should include in our repo.

ianb:
This includes the mozilla-central commit from https://bugzilla.mozilla.org/show_bug.cgi?id=1466349
Commit is https://hg.mozilla.org/releases/mozilla-beta/rev/4a309778103f

### DIFF
--- a/addon/bootstrap.js
+++ b/addon/bootstrap.js
@@ -99,9 +99,10 @@ const LibraryButton = {
 };
 
 const APP_SHUTDOWN = 2;
-let startupReason;
+let addonData, startupReason;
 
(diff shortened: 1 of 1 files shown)


chenba:
@ianb There are a couple new permission entries in that hg commit.  Are those going to be copied over?  I don't remember seeing a bug for that.

PR #4536, opened 2018-06-07
Fix #4012, removing potentially sensitive information on title update

### LINKED ISSUE #4012: <title> is automatically submitted when uploading screenshot
I originally reported this over irc in #security and was asked to file an issue:

I noticed when taking a screenshot and uploading it to screenshots.firefox.com, the <title> of the page I screenshoted is automatically submitted and used as a title on screenshots.firefox.com. While sensitive information on the screen can be avoided using the cropping tool, it's non obvious that the <title> is going to be submitted as well (in my case, containing the info that I cropped out :)).

I understand why it's built the way it is and not sure if and how it should be fixed, I thought I just leave this here as feedback.

testeaxeax:
Fix #4012

### DIFF
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -807,6 +807,9 @@ app.post("/api/set-title/:id/:domain", function(req, res) {
       return null;
     }
     shot.userTitle = userTitle;
+    if (shot.openGraph && shot.openGraph.title) {
+        shot.openGraph.title = userTitle;
+    }
     return shot.update();
   }).then((updated) => {
     simpleResponse(res, "Updated", 200);


server/src/server.js
```diff
@@ -807,6 +807,8 @@ app.post("/api/set-title/:id/:domain", function(req, res) {
       return null;
     }
     shot.userTitle = userTitle;
+    // Here we remove potentially confidential data
+    shot.openGraph = {};
```
chenba: I think the title should be updated.

If there are concerns regarding the data saved along with the shot, then please file another bug so it can be triaged.

testeaxeax:
Updated commit.

server/src/server.js
```diff
@@ -807,6 +807,10 @@ app.post("/api/set-title/:id/:domain", function(req, res) {
       return null;
     }
     shot.userTitle = userTitle;
+    // Here we remove potentially confidential data
+    if (shot.openGraph) {
+        shot.openGraph.title = null;
+    }
```
chenba: Sorry that I wasn't clear before, but we should update `openGraph.title` to the new title.

testeaxeax: Sorry I misunderstood you. I hope the commit is okay now.

chenba:
@testeaxeax Thanks!

PR #4617, opened 2018-07-12
Put in a package-lock.json file

ianb:
Per [Bug 1475246](https://bugzilla.mozilla.org/show_bug.cgi?id=1475246), this is to keep our transitive dependencies locked


chenba:
CI is failing because NSP is reporting quite a number of new vulnerabilities.

ianb:
Why would it be reporting new vulnerabilities? This should be all the packages we normally install. Does nsp look at package-lock for what it should check? (I didn't think it did, I recall getting errors for packages that we didn't directly require)

6a68:
This doesn't make any sense. I tried manually removing the line from the .npmrc file, re-running npm install locally, then doing `./node_modules/.bin/nsp check -o summary`, with no vulnerabilities found.

chenba:
nsp looks at `package-lock.json` if that exists (in addition to `package.json`).  And there's no option to switch that off.  :-/

6a68:
I’m generating the package-lock locally (via npm install *after* removing the line from the .nsprc) and not seeing this error. Could someone else try?

chenba:
Argh, I tried it again.  First I got 13 vulnerabilities.  (15 was reported in the CircleCI run for those keeping scores.)  Then `(+) No known vulnerabilities found` consistently after. 😕

I'll run the tests again on CircleCI.

6a68:
@punamdahiya Try removing that line from your .npmrc, then re-running `npm install` to generate a package-lock file, and see if the reported vulnerabilities change

punamdahiya:
@6a68 tried deleting the entry in .npmrc and regenerating package-lock.json. It still complains about the above 2 vulnerabilities  because of base64URL version 2.0.0

chenba:
Not only I get a different number of vulnerabilities locally than CircleCI, but even the output format is different this time.

6a68:
Other than those things, this works OK for me (with a gigantic list of `npm audit` errors that we ignore, for now).

chenba:
Are the changes to `ajv`, `addons-linter`, and `svgo` in package.json intentional?   Otherwise it looks ready to be merged.

chenba:
Oh those are the results of npm audit fix.  👍

chenba:
Is this ready to be merged?

ianb:
I feel unclear why this does or doesn't pass checks now, but let's say it's ready to merge and go from there!

PR #4621, opened 2018-07-13
Fixes #4620 - Text tool vertical shift on confirm

### LINKED ISSUE #4620: [Annotations] Text shifts up slightly on confirm
Inside Texttool, slight shift is seen when text is rendered on canvas on hit of confirm. This is because y position used to fill text on canvas varies across platform due to difference in line-height.
For example for large  font-size 36p, line-height in windows - 41.5, unix - 43, mac- 36.5

punamdahiya:


punamdahiya:
@chenba Patch tries to adjust y position of rendering text on Canvas by computing difference between  line-height and font-size and trying to reduce shift by adding one-third of difference to y position.

This definitely tone down the shift in Windows and Mac. Please test with unix and review if with patch we have better UX. Thanks!

server/src/pages/shot/text-tool.js
```diff
@@ -155,8 +155,12 @@ exports.TextTool = class TextTool extends React.Component {
     }
     const styles = window.getComputedStyle(this.textInput.current);
     const FONT_SIZE = parseInt(styles["font-size"], 10);
+
+    // Due to line-height differences of how fonts are rendered across platforms
+    // adjust text y position to one-third of difference of line-height and font-size
+    const ADJUST_VERTICAL_SHIFT = (parseFloat(styles["line-height"]) - FONT_SIZE) / 3;
```
chenba: How did you arrive at `3` here?

PR #4649, opened 2018-07-16
Fixes #4643 - Focus text input on color change

### LINKED ISSUE #4643: [Annotations] The text caret wrongly disappears after the string color is changed
[Affected versions]:
- Nightly 63.0a1
- Dev Edition 62.0b7
- Release 61.0.1 Unbranded

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a clean Firefox profile with the latest Firefox Screenshots v33.0.0dev installed.
- Have a saved shot link.

[Steps to reproduce]:
1. Open the Firefox browser with the profile from prerequisites and navigate to the saved shot link.
2. Click the "Edit this image" button and click the "Add text" button.
3. Click the "Color Picker" button and change the color.
4. Observe the behavior.

[Expected result]:
- The color is changed and the text caret is displayed in the text box.

[Actual result]:
- The color is changed but the text caret disappears from the text box.

[Notes]:
- This issue is not reproducible if you change the text size.
- Here is a screen recording with this issue:
ss - text caret dissapears

punamdahiya:


punamdahiya:
@chenba Adding back the focus on text input on color change, this behavior is consistent with text size change and helps user to identify where to type text especially when text input is empty and color picked has same background color as color on image

punamdahiya:
@chenba Yes it takes default behavior of focus which is right now beginning of the string. Without this fix cursor disappears on color change. My take showing cursor is better than hiding especially if there's no content in text input. 

For user who changes color after typing text, they can always click at desired location to move text caret till we find a better fix.

chenba:
I do like it for the empty input case.  Since the beginning of the input div can be outside of the shot, there's still a case where users can't see where the cursor is.

And I'd think majority of the time, users do not want to resume typing at the beginning of the input.  As you said, users can always click into the desired position.  But if they need to do that anyway, then this doesn't really fix much.

PR #4656, opened 2018-07-17
Wait to fetch image data a few times in Watchdog.

### LINKED ISSUE #4612: [Watchdog] Investigate Why dev Screenshots Server Is Not Sending Requests to dev Watchdog
Follow-up to #4606 
Followed by #4610

chenba:
Fixes #4612

### DIFF
--- a/server/src/watchdog.js
+++ b/server/src/watchdog.js
@@ -116,19 +116,20 @@ exports.submit = function(shot) {
   // image data. This is called for every shot saved, so let's keep it simple
   // and minimal.
 
+  const delay = function(waitInMs) {
+    return new Promise((resolve) => {
+      setTimeout(() => resolve(), waitInMs);
+    });
+  };
+
+  const getRawBytesPromise = function(timeToWait) {
+    return delay(timeToWait).then(() => Shot.getRawBytesForClip(imageId));
+  };
+
   const getImageRawBytes = function() {
     const MAX_ATTEMPTS = 3;
     const DELAY = 1000; // milliseconds
     let attemptCount = 1;
-
-    const delay = function(waitInMs) {
-      return new Promise((resolve) => {
-        setTimeout(() => resolve(), waitInMs);
-      });
-    };
-    const getRawBytesPromise = function(timeToWait) {
-      return delay(timeToWait).then(() => Shot.getRawBytesForClip(imageId));
-    };
 
     return new Promise((resolve, reject) => {
       const tryGetBytes = function() {
--- a/server/src/watchdog.js
+++ b/server/src/watchdog.js
@@ -107,14 +107,53 @@ exports.submit = function(shot) {
   const imageId = imageUrl.split("/").pop();
   const form = new FormData();
 
-  Shot.getRawBytesForClip(
-    imageId
-  ).then((obj) => {
-    if (obj === null) {
-      const err = new Error("Unable to fetch image for Watchdog submission.");
-      err.imageId = String(imageId);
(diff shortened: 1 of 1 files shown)


server/src/watchdog.js
```diff
@@ -107,14 +107,53 @@ exports.submit = function(shot) {
...
+      });
+    };
+    const getRawBytesPromise = function(timeToWait) {
+      return delay(timeToWait).then(() => Shot.getRawBytesForClip(imageId));
+    };
+
+    return new Promise((resolve, reject) => {
+      const tryGetBytes = function() {
+        getRawBytesPromise(DELAY).then(obj => {
+          if (obj === null) {
```
6a68: The beauty of Promise code is that it's not necessary to manually maintain a counter.

What if `getRawBytesPromise` just did something like:

```
// Three attempts with 1000 msec delay
getRawBytesPromise(1000)
  .then(bytes => { return bytes ? bytes : getRawBytesPromise(1000) })
  .then(bytes => { return bytes ? bytes : getRawBytesPromise(1000) })
  .then(bytes => { bytes ? resolve(bytes) : reject(new Error(...)) }
```

I think this approach would be simpler and more idiomatic.

server/src/watchdog.js
```diff
@@ -107,14 +107,53 @@ exports.submit = function(shot) {
...
+    const MAX_ATTEMPTS = 3;
+    const DELAY = 1000; // milliseconds
+    let attemptCount = 1;
+
+    const delay = function(waitInMs) {
+      return new Promise((resolve) => {
+        setTimeout(() => resolve(), waitInMs);
+      });
+    };
+    const getRawBytesPromise = function(timeToWait) {
```
6a68: Probably couldn't hurt to move `getRawBytesPromise` and the `delay` function out to top level, to avoid redeclaring them over and over.

server/src/watchdog.js
6a68: Er, sorry, I meant "what if `getImageRawBytes` did something like..."

chenba: I'd argue in this case the counter is not necessary because we call the function three times.  Promises don't have anything to do with it.  😃  It is easier to read though.

server/src/watchdog.js
chenba: hmmm I did that to minimize the scope of the functions.  I think it wouldn't hurt anything?  I'll move them up.

server/src/watchdog.js
chenba: Currently the code retries on the first two rejections from `Shot.getRawBytesForClip` and rejects with its Error on the third.  I don't think we can switch to the proposed promise chain without losing the original Error from `Shot.getRawBytesForClip`.  In order for `getRawBytesPromise` to return something falsey, it needs to ignore the rejections.

6a68: Sure, you could handle the rejection case any number of ways while still getting rid of the manual iteration counting code. We can chat about this tomorrow after standup :+1:

chenba: Sorry, I wasn't clear: I want to reject if it's `null` but retry two times if it's a rejection.  `Shot.getRawBytesForClip` returns null when a (not {deleted,blocked,expired}) image is not found in the DB, so that's not an AWS/S3 related issue.

chenba:
Pushed additional commits for ease of review.  For the promise chain approach, I added a flag for retrying on `Shot.getRawBytesForClip` rejection.

Issue #4653, opened 2018-07-17
[Annotations][Enhancement] The Text tool should support pasting formatted text or it should discard formatting but still paste the text


Softvision-AndreeaCupsa:
**[Notes]:**
- A lot of users don't know about simple pasting (with "CTRL"/"CMD" + "SHIFT" + "V" keys) and only more technical users know about this method which lead the user to not be able to paste formatted text with "CTRL" + "V" keys or "Paste" option from the context menu in the "Add text" section, this in turn affects the overall user experience. 

- Most websites have their own text content formatted and this leads the user to not being able to paste text with "CTRL" + "V" keys or "Paste" option from the context menu from these websites in the "Add text" section.

**[Affected versions]:**
- Nightly 63.0a1
- Dev Edition 62.0b8
- Release Unbranded

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
 - The "xpinstall.signatures.required" boolean pref is set to "false".
 - The "extensions.legacy.enabled" boolean pref is set to "true".
 - The "extensions.install.requireBuiltInCerts " boolean pref is set to "false"
 - Have the Screenshots 33.1.0 dev version installed from here, on a new clean profile.
 - Have a saved shot link.

**[Steps to reproduce]:**
1. Open the Firefox browser with the profile from the prerequisites and navigate to the saved shot link.
2. Save the following string to the clipboard: **_String_**.
3. Click the "Edit this image" button.
4. Click the "Add text" button.
5. Paste the string with "CTRL" + "V" keys or "Paste" option from the context menu and observe the behavior.

**[Expected result]:**
 - The string is pasted into the text box and discards all previous formatting and takes on the selected style characteristics of the "Add text" section.

**[Actual results]:**
 - The string is not pasted.

**[Additional notes]:**
- Attached a screen recording of the issue:

![rec of the issue - paste](https://user-images.githubusercontent.com/37875464/42817182-fe541210-89d5-11e8-9ef5-8b97a3255eba.gif)

johngruen:
Let's just strip out formatting

chenba:
Perhaps we should consider using a text input instead of a contenteditable div.

punamdahiya:
Yes, we should explore pros and cons of moving from contenteditable div to text input, however at this stage  I believe that's a core change and that refactoring should be taken as separate issue so that the fix can be tested for all flows

chenba:
[...]

PR #4657, opened 2018-07-17
Paste formatted text as plain text

punamdahiya:


punamdahiya:
@chenba opened #4658 for switching to text input that is another approach of fixing this and few more issues logged in QA. Submitting this PR which fixes #4653 with contenteditable div

chenba:
We no longer need this.  Closing.

PR #4659, opened 2018-07-18
Clear only dirty areas of a canvas.

### LINKED ISSUE #4463: [Annotations] clearRect() Only on Dirty Areas
There are a few place where `clearRect()` is called for the entire area of the canvas.  In the tools, we should only `clearRect()` on the dirty area.

chenba:
Fixes #4463

### DIFF
--- a/server/src/pages/shot/drawing-tool.js
+++ b/server/src/pages/shot/drawing-tool.js
@@ -63,8 +63,8 @@ exports.DrawingTool = class DrawingTool extends React.Component {
     }
     this.mousedownPosition = this.captureMousePosition(e);
     this.isMousedown = true;
-    this.draw(this.mousedownPosition);
     this.updateDrawnArea(this.mousedownPosition);
+    this.draw(this.mousedownPosition);
   }
 
   onMouseMove(e) {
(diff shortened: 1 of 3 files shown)


6a68:
OK, start with an image plucked from mozilla.org:

screen shot 2018-07-18 at 2 41 51 pm

Highlight across part of the image:

screen shot 2018-07-18 at 2 42 49 pm

Undo and redo a few times. Note the blurriness and offset of the chunk that was highlighted - this only gets worse as you undo/redo some more, or undo/redo on top of additional drawing areas:

screen shot 2018-07-18 at 2 43 22 pm

As I mentioned in the review, I'm guessing this comes from rounding off subpixel distances, but maybe it also has to do with pixel density issues?

6a68:
Here's how it looks after undo/redo about 20x, ending with the highlight reapplied:

screen shot 2018-07-18 at 2 47 51 pm

chenba:
So strange that it's not reproducible on every shot.  Even with two shots of identical dimensions, I'd get different results. 😕

chenba:
The issue with the blurred area is not from this PR and is now tracked in #4678.

PR #4708, opened 2018-07-25
Fixes #4680 - Text width should not extend beyond 5px inside canvas

### LINKED ISSUE #4680: [Annotations] Text width should not extend beyond 5px inside canvas
After playing with the canvas, i think we need to adjust the max string length so that text input cannot exceed the width of the canvas - 10px (5 on each side)

Now there's an edge case where the user could type longer strings with a small font and then resize, but i think that's OK

punamdahiya:
Fixes #4680

### DIFF
--- a/server/src/pages/shot/text-tool.js
+++ b/server/src/pages/shot/text-tool.js
@@ -58,9 +58,9 @@ exports.TextTool = class TextTool extends React.Component {
     hasFirstInput = false;
     this.textInput.current.nextSibling.textContent = this.textInput.current.placeholder;
     this.textInput.current.focus();
+    this.textInput.current.nextSibling.style.maxWidth = `${this.canvasCssWidth - 2 * TEXT_DRAG_EDGE_LIMIT}px`;
     this.adjustWidth();
     previousTextInputWidth = this.textInput.current.clientWidth;
-    this.center = this.getCenter();
     if (this.props.toolbarOverrideCallback) {
       this.props.toolbarOverrideCallback();
     }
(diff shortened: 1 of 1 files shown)


chenba:
Having a limit based on character count doesn't work very well.  If I enter a bunch of `Q`s [0], delete them, and enter `i`s, the limit is way too low.

[0] In my particular shot, the first `Q` got chopped off: 
image

punamdahiya:
@chenba good catch! updated PR to remove maxlength attribute and use hidden div to enforce text content limit when text input width exceeds canvas. Please review

chenba:
The issue with the first character being clipped in the input still exists:

q-before

(This depends on the combination of shot width and text content; it doesn't happen every time.)

And then the text shifts to the right so the first character is displayed in its entirety on confirm:

q-after

punamdahiya:
@chenba updated PR to account for padding on both sides before enforcing text content limit which for some character sizes results in less character but better experience than clipped. Please review. Thanks!

chenba:
The limit works great.  There's still a small problem with the alignment.  If I reach the limit by entering characters starting at the default position, there's an empty margin on the left but not the right.  This is the yellow line with the stars in the image below.  The width of the empty margin depends on the text entered.

If I drag the text around after reaching the limit, then I would encounter the edge limit.  This is the green line in the image below.

I'll leave it up to you if you want to fix it in this PR or open another issue for it.  It's a pretty small bug.

image

punamdahiya:
Thanks @chenba,  looks like need to debug and adjust left position when text content reaches max limit which I will prefer to triage and fix as separate bug

PR #4713, opened 2018-07-27
Fixes #4652 - Scroll text input into view

### LINKED ISSUE #4652: [Annotations] Text box is not brought into view when it is refocused
**[Affected versions]:**
- Nightly 63.0a1
- Dev Edition 62.0b8
- Release Unbranded

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
 - The "xpinstall.signatures.required" boolean pref is set to "false".
 - The "extensions.legacy.enabled" boolean pref is set to "true".
 - The "extensions.install.requireBuiltInCerts " boolean pref is set to "false"
 - Have the Screenshots 33.1.0 dev version installed from here, on a new clean profile.
 - Have a very long (over 2000px) shot saved.

**[Steps to reproduce]:**
1. Open the Firefox browser with the profile from the prerequisites and navigate to my shots page. 
2. Click on the long shot and click on "Edit this image" button.
3. Click the "Add text" button.
4. Scroll down until the text box is no longer visible.
5. Change the text size and observe the behavior.

**[Expected result]:**
 - The text box is correctly focused and is in view.

**[Actual results]:**
 - The area below the text box is in view, but the text box is not in view.
 
 **[Notes]:**
 - This issue is also reproducible if you change the text color or type something when the text box is focused but is out of view.
 - This issue is not reproducible if the text box is below the visible area.
 - Screen recording with the issue: [link](https://drive.google.com/open?id=1QM0kBNWjZSDRoaTUQEiwo8dbF_OMtueL)

punamdahiya:


punamdahiya:
@chenba PR fixes refocus scrolled textbox on change of color or text size. Please review

chenba:
Any reason you chose `{block: "end"}` over `{block: "center"}`?

chenba:
Having the text input at the very bottom is probably not what people are used to and is less noticeable since people won't look for it there.  It's especially unnatural when scrolling downward.  Can we add some detection so that scrollIntoView is called only when the text input is not visible?

punamdahiya:
true, just noticed  on subsequent scrolled pages , keeping text in center will be a better experience and noticeable than in end even with the nit mentioned in earlier comment.  As for restricting scrollIntoView, will check if there is an optimized way to detect and avoid it.

punamdahiya:
@chenba Updated PR to detect and call scrollInToView only when input is not visible . Thanks

chenba:
Works great.

PR #4726, opened 2018-07-31
Replace shot expiration with favorites for FxA users

### LINKED ISSUE #4688: [FxA] Remove Time selection and add favorites for FxA users
Per the spec: https://mozilla.github.io/testpilot-assets/Firefox_Screenshots/NEW_FxA_Integration/shot

We still need  clear indicator favorited shots on the my shots page

chenba:
Fixes #4688

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -176,7 +176,6 @@ class Body extends React.Component {
     this.state = {
       hidden: false,
       closeBanner: false,
-      isChangingExpire: false,
       imageEditing: false
     };
   }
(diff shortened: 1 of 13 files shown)


server/src/pages/shot/view.js
```diff
@@ -340,19 +339,21 @@ class Body extends React.Component {
...
-    }
 
+    let favoriteShotButton = null;
     let trashOrFlagButton;
     let editButton;
     const highlight = this.state.highlightEditButton ? <div className="edit-highlight" onClick={ this.onClickEdit.bind(this) } onMouseOver={ this.onMouseOverHighlight.bind(this) } onMouseOut={ this.onMouseOutHighlight.bind(this) }></div> : null;
+
+    if (this.props.isFxaAuthenticated) {
+      const activeFavClass = this.props.expireTime ? "" : "is-fav";
+      favoriteShotButton = <div className="fav-wrapper"><button
```
chenba: This `div` is here because brightness filter on the icon image affects the background color, which make the hover and active background inconsistent with the other icons.

server/src/pages/shot/view.js
```diff
@@ -528,6 +515,18 @@ class Body extends React.Component {
...
 
+  onClickFavorite() {
+    if (this.props.expireTime) {
+      sendEvent("set-favorite", "navbar");
+      const INDEFINITE = 0;
+      this.props.controller.changeShotExpiration(this.props.shot, INDEFINITE);
+    } else {
+      sendEvent("unset-favorite", "navbar");
+      const TWO_WEEKS_IN_SECONDS = 1209600;
+      this.props.controller.changeShotExpiration(this.props.shot, Date.now() + TWO_WEEKS_IN_SECONDS);
```
punamdahiya: Date.now() here will be in milliseconds and should be updated to use Date.now() /1000

server/src/pages/shotindex/view.js
```diff
@@ -329,7 +330,11 @@ class Card extends React.Component {
 
     let neverExpireIndicator = null;
     if (!shot.expireTime) {
-      neverExpireIndicator = <Localized id="shotIndexNoExpirationSymbol"><div className="never-expires" title=""></div></Localized>;
+      if (this.props.hasFxa) {
+        neverExpireIndicator = <Localized id="shotIndexFavoriteIcon"><div className="favorite-shot" title=""></div></Localized>;
```
punamdahiya: shotIndexFavoriteIcon needs  to be added in  server.ftl

server/src/pages/shot/view.js
chenba: Good catch.  The value should be in ms not seconds.

punamdahiya: cool, update metrics.md should be updated with new event 'unset-favorite' and 'set-favorite'

punamdahiya: and remove ones which are no longer used such as set-expiration-to-indefinite

punamdahiya:
LGTM!  @6a68  @ianb  This PR adds a dependency on psycopg2 adapter to access postgres db from python for fixing server tests for account id.
https://github.com/mozilla-services/screenshots/pull/4726/commits/25afa8d112d55c81d80df108989ebb1bb86b9881
As you have more context of our build system, keeping in loop about this change. Thanks!

6a68:
We discussed this in the morning meeting today. The alternative to flipping the state in the DB directly would be to manually run the automation through the OAuth dance, which can be kind of buggy, and will increase our false failure rate. This seems like a fine approach; we can always add oauth tests later. Merging

PR #4729, opened 2018-08-01
Fix #4665; Sign dev and stage XPIs with the autograph dev-root

### LINKED ISSUE #4665: Sign dev/stage built XPIs with Autograph
To ensure the `mozillaAddons` permission is not stripped out of our dev and stage builds, we can set up Autograph signing for Circle-CI.

Autograph docs on XPI signing: https://github.com/mozilla-services/autograph/blob/master/signer/xpi/README.rst

6a68:
@g-k Mind taking a look? Not sure how I can verify the built xpi is really signed until this lands in master. Suggestions welcome :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,9 @@ bootstrap_zip: addon
 unsigned_bootstrap_xpi: bootstrap_zip
 	cp build/screenshots-bootstrap.zip build/screenshots.xpi
 
+dev_signed_bootstrap_xpi: bootstrap_zip
+	curl -F "input=@build/screenshots-bootstrap.zip" -o build/screenshots.xpi -H "Authorization: ${AUTOGRAPH_EDGE_TOKEN}" https://autograph-edge.stage.mozaws.net/sign
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -115,6 +115,9 @@ bootstrap_zip: addon
 unsigned_bootstrap_xpi: bootstrap_zip
 	cp build/screenshots-bootstrap.zip build/screenshots.xpi
 
+dev_signed_bootstrap_xpi: bootstrap_zip
+	curl -v -F "input=@build/screenshots-bootstrap.zip" -o build/screenshots.xpi -H "Authorization: ${AUTOGRAPH_EDGE_TOKEN}" https://autograph-edge.stage.mozaws.net/sign
```
g-k: Will `-v` result in the token being logged? Maybe we could pipe through sed or `grep -v` to filter it out and retain verbose output.

6a68: Good catch. Let's see how far we can get without the `-v`

PR #4734, opened 2018-08-02
Suppress logging the actual curl command to avoid leaking the autograph edge token

6a68:
@g-k r?


static/homepage/install-test-local.html
```diff
@@ -38,6 +38,7 @@ <h1>Dev Firefox Screenshots</h1>
             <li><strong><code>xpinstall.signatures.required</code></strong> to <code>false</code></li>
             <li><strong><code>extensions.legacy.enabled</code></strong> to <code>true</code></li>
             <li><strong><code>extensions.install.requireBuiltInCerts</code></strong> to <code>false</code></li>
+            <li><strong><code>xpinstall.signatures.dev-root</code></strong> to <code>true</code>
```
g-k: @6a68  FWIW This dev-root will need to be added as a new boolean property, but settings the others shouldn't be necessary.

PR #4737, opened 2018-08-03
Add special shot hostname config for Watchdog

### LINKED ISSUE #4725: [Watchdog] Set Positive Match Only for Shots of a Special Domain on dev
dev Watchdog randomly returns a positive match about 10% of the time.  This can be annoying and disruptive for QA using dev Screenshots.  Add a configurable special domain that, when set, a positive Watchdog match will take effect.

Note that this means about 10% of the shots from the special domain will be a positive match, not 100%.

Current nominee for special domain: w3schools

chenba:
Fixes #4725 

I will also file a PR in the cloudops repo to set the value in dev.

### DIFF
--- a/server/src/watchdog.js
+++ b/server/src/watchdog.js
@@ -103,6 +103,19 @@ exports.submit = function(shot) {
     return;
   }
 
+  if (config.watchdog.devOnlyMatchHostname) {
+    try {
+      const url = new URL(shot.origin);
+      if (url.hostname !== config.watchdog.devOnlyMatchHostname) {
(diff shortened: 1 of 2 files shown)


server/src/watchdog.js
```diff
@@ -103,6 +103,12 @@ exports.submit = function(shot) {
     return;
   }
 
+  if (config.watchdog.devOnlyMatchHostname
+      && ((new URL(shot.origin)).hostname !== config.watchdog.devOnlyMatchHostname)) {
```
6a68: Wrap the `new URL()` call in a try/catch to prevent errors due to shots from weird URLs like `file:///`, where `siteName` is set to `null` when uploading ([here's](https://screenshots.firefox.com/WLyju7NyRbn6jwGs/null) an example).

PR #4749, opened 2018-08-06
Fixes #4741 - Update color in text tool for localized sites

### LINKED ISSUE #4741: color option in new text tool does not work
The color option in the new text tool does not work at all. The text is always black.

bildschirmfoto 2018-08-04 um 23 24 22

punamdahiya:


punamdahiya:
@ianb Please review PR that fixes color selection inside Text tool for localized sites. Thanks

chenba:
Fixes #4741

server/src/pages/shot/color-picker.js
```diff
@@ -108,9 +108,11 @@ exports.ColorPicker = class ColorPicker extends React.Component {
 
   onClickSwatch(e) {
     const color = e.target.style.backgroundColor;
+    const colorName = e.target.dataset.colorName.toLowerCase().replace(/\s/g, "-");
```
chenba: There is no need to use regex here.

punamdahiya:
@chenba Updated PR with the feedback. Thanks

chenba:
Hmm is there any reason that the data attributes aren't just the value we need?  I don't see why they need to be user friendly strings.

punamdahiya:
for now data values are used to point to correct css class and setting values to classnames can work.

punamdahiya:
Updated to pick colorName directly from data values.

PR #4750, opened 2018-08-06
Fixes #4724 - Add image editor new feature promotion dialog

### LINKED ISSUE #4724: Put in a popup promotion for the editing tools
Would be a simple popup dialog, that would show to people once (twice?) to tell them about this new/improved feature.

We could use the same React component we use to invite people to use Firefox. Or we can use the component used for the share dialog. We can use localStorage to track who has seen the popup before.

@clouserw is going to ask about text/copy. We should get some text in soon so we can get it to translators.

punamdahiya:


punamdahiya:
Screenshot showing new edit feature promotion dialog on shot page
promotion-dialog-screenshot

punamdahiya:
@ianb PR adds promo dialog that can be used to promote new Screenshots features. Thanks!

punamdahiya:
Thanks @flodolo updated PR with new copy from Michelle, there is no change in ids

server/src/pages/shot/controller.js
```diff
@@ -20,6 +20,17 @@ function shouldHighlightEditIcon(model) {
   return !hasSeen;
 }
 
+function shouldShowPromo(model) {
+  if (!model.isOwner) {
+    return false;
+  }
+  const hasSeen = localStorage.hasSeenPromoDialog;
+  if (!hasSeen && model.enableAnnotations) {
+    localStorage.hasSeenPromoDialog = "1";
```
ianb: I don't think we need to be unduly shy, we could show it several times.

server/src/promo-dialog.js
```diff
@@ -0,0 +1,51 @@
...
+const sendEvent = require("./browser-send-event.js");
+
+exports.PromoDialog = class PromoDialog extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    if (this.props.display) {
+      this.props.promoOpen(this.props.promoType);
```
ianb: Given how `promoOpen` is implemented, I think this should only happen on `componentDidMount` – at render time we don't know what's in the document.

server/src/promo-dialog.js
```diff
@@ -0,0 +1,51 @@
...
+      return <div id="promo-dialog-panel" className="promo-panel" >
+        <div className={`${this.props.promoType} promo-box default-color-scheme`}>
+        <a className="box-close" onClick={this.closePanel.bind(this)}></a>
```
ianb: This should probably be a button (though I'm not sure if that matters much), and should have an aria-title or something on it. We don't currently have a string for just "close", but as I think about it, the string should be less ambiguous anyway, like "close notification".

server/src/promo-dialog.js
```diff
@@ -0,0 +1,51 @@
...
+          <span className="title">Take Note!</span>
+        </Localized>
+        <Localized id="promoMessage">
+          <p className="message">
+            Updated editing tools let you crop, highlight, and even add text to your shot.
+          </p>
+        </Localized>
+        <span className="promo-star"></span>
+         <Localized id="promoLink">
+          <span id="promo-link" className="message link" onClick={this.closePanel.bind(this)}>
```
ianb: But this probably should be a link or button, since it is interactable. You could use `aria-role`, but that's not great, while just making it an `<a>` or `<button>` basically solves the accessibility issue.

server/src/pages/shot/controller.js
punamdahiya: @ianb is there a recommended frequency (2-3 times?)

ianb: I will arbitrarily say 3, because 3 is a good number.

punamdahiya: or interval like for next 1-2 days from first seen?

punamdahiya: ok cool will change it to display for 3 shot page loads

server/src/promo-dialog.js
punamdahiya: Updated to <a> and use aria-label as "Try Edit Link"

server/src/promo-dialog.js
punamdahiya: Updated to and use aria-label as "Close Notification"

server/src/promo-dialog.js
punamdahiya: componentDidMount have this.props.display as undefined on first render, moved the check inside componentDidUpdate

punamdahiya:
@6a68 Adding you as reviewer on PR updated with @ianb feedback on aria-labels and increasing frequency of notification display to 3 shot page loads

server/src/pages/shot/controller.js
```diff
@@ -20,6 +20,36 @@ function shouldHighlightEditIcon(model) {
   return !hasSeen;
 }
 
+function shouldShowPromo(model) {
+  if (!model.isOwner || !model.enableAnnotations) {
+    return false;
+  }
+  let show = false;
+  const count = localStorage.hasSeenPromoDialog;
+  const interval = 1; // in minutes
```
punamdahiya: Now that we have increased frequency of promo display to 3, interval here controls when the next notification should be shown from the last seen , it's set to 1 minute for now

server/src/promo-dialog.js
```diff
@@ -0,0 +1,56 @@
...
+            Updated editing tools let you crop, highlight, and even add text to your shot.
+          </p>
+        </Localized>
+        <span className="promo-star"></span>
+         <Localized id="promoLink">
+          <a id="promo-link" aria-label="Try Edit Tool" className="message link" onClick={this.closePanel.bind(this)}>
+            Give it a try
+          </a>
+        </Localized>
+        <span className="promo-star"></span>
```
6a68: Instead of two spans and a link, just make this a regular paragraph with emoji at the start and end. There's no need for the fancy `::before` CSS stuff.

`<p>✨ <Localized id="promoLink">Give it a try</Localized> ✨</p>`

static/css/frame.scss
```diff
@@ -47,6 +47,61 @@
...
+    content: "\02728";
+  }
+
+  .link {
+    cursor: pointer;
+    color: #009ec0;
+  }
+
+  a.box-close{
+    float:right;
```
6a68: The parent card is positioned, so just absolutely position this to the top and right, rather than get into the sketch of float-based layout.

server/src/promo-dialog.js
```diff
@@ -0,0 +1,56 @@
...
+exports.PromoDialog = class PromoDialog extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    if (this.props.display) {
+      return <div id="promo-dialog-panel" className="promo-panel" >
+        <div className={`${this.props.promoType} promo-box default-color-scheme`}>
+        <a className="box-close" aria-label="Close Notification" onClick={this.closePanel.bind(this)}></a>
```
6a68: Indent the children of promo-box

static/css/frame.scss
```diff
@@ -47,6 +47,61 @@
...
 
+.promo-panel {
+  .edit {
+    right: 50px;
+    top: 75px;
+  }
+  .promo-box {
+    border-radius: 3px;
+    box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.3), 0 0 20px rgba(0, 0, 0, 0.2);
+    position: absolute;
```
6a68: The promo card is intended to be overlaid on top of the page, not participate in the page's layout, so it would make more sense to position the outermost element, the .promo-panel itself.

static/css/frame.scss
```diff
@@ -47,6 +47,61 @@
...
+  }
+
+  .title {
+    font-size: 14px;
+    line-height: 14px;
+    font-weight: bold;
+  }
+
+  .message {
+    display: inline-block;
```
6a68: I don't think you need `inline-block` here. (Was there a particular problem you were trying to solve?)

static/css/frame.scss
```diff
@@ -47,6 +47,61 @@
...
+  }
+
+  a.box-close{
+    float:right;
+    cursor:pointer;
+    color: #000;
+    width:15px;
+    height: 15px;
+    line-height: 10px;
+    margin: -8px -3px 0 0;
```
6a68: If you use absolute positioning, I don't think you'll need to resort to negative margin layout tricks

static/css/frame.scss
```diff
@@ -47,6 +47,61 @@
...
+    cursor:pointer;
+    color: #000;
+    width:15px;
+    height: 15px;
+    line-height: 10px;
+    margin: -8px -3px 0 0;  
+  }
+
+  .box-close:before {
+    content: "×";
```
6a68: Probably better to just use the close icon from the photon design system, see the 'Navigation' section at https://design.firefox.com/icons/viewer/

static/css/frame.scss
```diff
@@ -47,6 +47,61 @@
...
+    top: 75px;
+  }
+  .promo-box {
+    border-radius: 3px;
+    box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.3), 0 0 20px rgba(0, 0, 0, 0.2);
+    position: absolute;
+    width: 180px;
+    z-index: 999;
+    padding: 10px 5px;
+    text-align:center;
```
6a68: missing space

static/css/frame.scss
```diff
@@ -47,6 +47,61 @@
   }
 }
 
+.promo-panel {
+  .edit {
+    right: 50px;
+    top: 75px;
```
6a68: Until we're using this panel in multiple places, it doesn't make sense to abstract out positioning for the edit button. This is confusing as written, since the 'edit' and 'promo-box' classes are applied to the same element. Just get rid of this .edit class and keep all the positioning declarations in one spot.

server/src/promo-dialog.js
```diff
@@ -0,0 +1,56 @@
...
+        <Localized id="promoTitle">
+          <span className="title">Take Note!</span>
```
6a68: Just use a header element here, and override its margins in the CSS. `h4` seems like roughly the right size

server/src/pages/shot/view.js
```diff
@@ -452,6 +456,24 @@ class Body extends React.Component {
     </reactruntime.BodyTemplate>);
   }
 
+  promoOpen(promoType) {
+    if (promoType === "edit") {
```
6a68: We don't use the promo card for anything else, so get rid of `promoType` and associated conditionals.

server/src/pages/shot/view.js
```diff
@@ -452,6 +456,24 @@ class Body extends React.Component {
...
+      document.querySelector(".button.transparent.edit").style.backgroundColor = "#ededf0";
```
6a68: We definitely want to avoid altering other components directly like this. It makes it really hard to reason about the behavior of each component in isolation. Instead, whenever this promo card is shown, we want to also make sure the `highlightEditButton` state is set. The parent page is already managing both bits of state in `componentDidMount`:

```js
  componentDidMount() {
    this.setState({highlightEditButton: this.props.highlightEditButton});
    this.setState({promoDialog: this.props.promoDialog});
  }
```

All you need to do is ensure the highlight is set whenever the promo is visible. I think this should work:

```js
  componentDidMount() {
    this.setState({highlightEditButton: this.props.highlightEditButton || this.props.promoDialog});
    this.setState({promoDialog: this.props.promoDialog});
  }
```

6a68: btw you can just apply the hover style to the edit button when `highlightEditButton` is true. That'd make the highlight a little more noticeable, even when the card isn't present.

static/css/frame.scss
punamdahiya: I think it was to support the longer message string and  previous UI of hand wave emoji, tried long string with title as text and it works

punamdahiya:
@6a68 Updated PR with the feedback,  only action item that needs to be thought  through is background-color on highlight button same as hover style.

6a68:
I'd suggest (1) create a 'highlight' class that applies the button hover style (2) use the `classnames` library to conditionally add that class name if `this.state.highlightEditButton` is true.

So, instead of this:

```js
// line 363 in pages/shot/view.js
editButton = <Localized id="shotPageEditButton">
  <button className="button transparent edit" title="Edit this image" onClick={ this.onClickEdit.bind(this) } ref={(edit) => { this.editButton = edit; }}></button>
</Localized>;
```

you'd have something like

```js
const highlightEdit = this.state.highlightEditButton;
editButton = <Localized id="shotPageEditButton">
  <button className={classnames("button", "transparent", "edit", {"highlight": highlightEdit})} title="Edit this image" onClick={ this.onClickEdit.bind(this) } ref={(edit) => { this.editButton = edit; }}></button>
</Localized>;
```

You'll also need to require the classnames library in the shot view. It's used in the homepage view already, in case you want to poke around at some other examples.

6a68:
@punamdahiya Looking good! Things to fix:

- You'll need to have two different top/right positions for the card, based on the media query that's used to collapse the buttons at narrow screen widths. The card looks centered at narrow width, but wrong at full width:

screen shot 2018-08-09 at 3 09 44 pm
screen shot 2018-08-09 at 3 09 49 pm


- The "Give it a try" text isn't supposed to be deleted, it just shouldn't be a link. See my [comment](https://github.com/mozilla-services/screenshots/pull/4750#discussion_r209043906) from earlier.

- The promo-dialog component is only used on the shot page, so it should be in the `server/src/pages/shot` directory, not top-level in the `server/src` directory.

server/src/pages/shot/view.js
```diff
@@ -452,6 +456,11 @@ class Body extends React.Component {
     </reactruntime.BodyTemplate>);
   }
 
+  promoClose() {
+    this.setState({promoDialog: false});
+    localStorage.hasSeenPromoDialog = "3";
```
6a68: Is there some reason to set this to the string `"3"` instead of the integer `3`?

server/src/pages/shot/view.js
```diff
@@ -480,6 +489,12 @@ class Body extends React.Component {
       this.setState({imageEditing: true});
       sendEvent("start-annotations", "navbar");
     }
+    // If user clicked edit after seeing new edit tool promo
+    // set counter to max to stop showing notification again
+    if (this.props.promoDialog) {
+      localStorage.hasSeenPromoDialog = "3";
```
6a68: Same issue here, just use the number 3.

punamdahiya:
@6a68 PR has the final set of changes updated. Thanks!

6a68:
@johngruen Hey, any thoughts on the language here? There's some awkward singular / plural conflict between the two lines. Also, this promo card is effectively part of onboarding, since new users will now always see it the first time they see a shot page; should the wording be tweaked to sound less like upselling new features, and more just advertising the features?

screen shot 2018-08-10 at 4 35 37 pm

6a68:
@punamdahiya I opened https://github.com/punamdahiya/screenshots/pull/2 against your PR to fix a few final things. It can wait until you're back; I think the copy could use another pass before we hand it off to localizers.

clouserw:
* "...text to your shot."  ->  "...text to your shots."
* "Give it a try" -> "Give them a try"

john has more feedback he will write in here

punamdahiya:
@6a68 taking out highlight style in favor of star icon LGTM. Updated PR with final copy and ready for  review. Thanks!

server/src/pages/shot/promo-dialog.js
```diff
@@ -0,0 +1,40 @@
...
+        </Localized>
+        <Localized id="promoTitle">
+          <h4 className="title">Take Note!</h4>
+        </Localized>
+        <Localized id="promoMessage">
+          <p className="message">
+            Updated editing tools let you crop, highlight, and even add text to your shot.
+          </p>
+        </Localized>
+        <p className="message">✨<Localized id="promoLink"><span className="message-text">Give it a try</span></Localized>✨</p>
```
6a68: Give *them* a try

server/src/pages/shot/promo-dialog.js
```diff
@@ -0,0 +1,40 @@
...
+      </div>;
+    }
+    return null;
+  }
+
+  closePanel(event) {
+    this.props.promoClose();
+    sendEvent("promo-closed");
```
6a68: Whenever a new metrics event is added, it should be documented in https://github.com/mozilla-services/screenshots/blob/master/docs/METRICS.md. Sorry I didn't catch this sooner.

static/css/frame.scss
```diff
@@ -47,11 +47,61 @@
...
+  .message-text {
+    margin: 0 2px;
+  }
+
+  a.box-close {
+    position: absolute;
+    top: 2px;
+    right: 2px;
+    cursor: pointer;
+    color: #000;
```
6a68: use `$black` instead

server/src/pages/shot/promo-dialog.js
punamdahiya: thanks, good catch!

server/src/pages/shot/promo-dialog.js
punamdahiya: Done

static/css/frame.scss
punamdahiya: Updated

PR #4759, opened 2018-08-09
Fixes #4683 - FxA SignIn from Homepage

### LINKED ISSUE #4683: [FxA] Implement Sign in From Site Homepage
Spec is available here: https://mozilla.github.io/testpilot-assets/Firefox_Screenshots/NEW_FxA_Integration/home

Header is the important bit as is the ability to be able to sign in via FXA

punamdahiya:


server/src/pages/homepage/model.js
```diff
@@ -4,6 +4,7 @@ exports.createModel = function(req) {
   const model = {
     title: "Firefox Screenshots",
     showMyShots: !!req.deviceId,
+    hasFxa: !!req.accountId,
```
punamdahiya: After successful /api/fxa-oauth/login redirect to homepage , req.accountId stays null. Loading shot  or shot index page sets req.accoutId and displays Setting links on homepage. @6a68 @chenba do you know what else needs to be ported into homepage model that's handled in shot pages to reflect req.accountId

chenba: Not sure what's going on from just reading the code.  The auth middleware should be executing for `/`.

chenba:
Fixes #4683

chenba:
@punamdahiya I updated the base branch of this PR to fxa-integration, which is the new feature branch for the FxA work.

punamdahiya:
@chenba Need inputs on below line items, thanks!
a) On home page , we don’t have authInfo.accountId set after oauth login redirection 
https://github.com/mozilla-services/screenshots/blob/master/server/src/server.js#L231
Do we need to explicitly call sendAuthInfo on oauth post redirection similar to one inside /api/login route handler?

https://github.com/mozilla-services/screenshots/blob/master/server/src/server.js#L613

b) On shot page, after oauth redirection, refresh of page is reflecting isFxaAuthenticated property correct value, this could be how React handles properties, thoughts?

 https://github.com/mozilla-services/screenshots/blob/fxa-integration/server/src/pages/shot/view.js#L572

chenba:
@punamdahiya a) I do see the accountId in the request object.  Do you have STR for me to try?  b) I don't quite understand the question.  What do you think React is doing?

punamdahiya:
@chenba As discussed  it seems SignButton is rendered before wantsauth call at launch in controller, completes setting accountId for shot and shot index pages.

Thanks for pointing to getAuthData wantsauth call below
https://github.com/mozilla-services/screenshots/blob/master/server/src/pages/shot/controller.js#L43

For homepage, similar to shot and shotindex page we should invoke window.wantsauth.getAuthData() to set request parameter with accountId from cookies

punamdahiya:
@chenba Updated homepage to ask for authentication information. Race condition between oauth redirection fxa-oauth/confirm-login in server.js  and getting accountId from cookies is still an issue. 
https://github.com/mozilla-services/screenshots/pull/4759/commits/4f83f0988c126ad196a8b41ac4bf141cd3974f80#diff-6d9c37686372a8dee7210c450a7c50a1R1092

In the PR here, I am reloading page inside controllers to show signin/settings button which will not stick (I think that's the reason of failing tests) and needs feedback on how to refresh model after calls ( /api/login, /api/set-login-cookie) from wantsauth.js that fills accountid cookie.

https://github.com/mozilla-services/screenshots/pull/4759/commits/4f83f0988c126ad196a8b41ac4bf141cd3974f80#diff-ab078f88c3f0d52c16befd93e12aaf53R9

https://github.com/mozilla-services/screenshots/pull/4759/commits/4f83f0988c126ad196a8b41ac4bf141cd3974f80#diff-6db5e3f6a06e986e2aa9ab0f5c0170d1R31


https://github.com/mozilla-services/screenshots/blob/master/static/js/wantsauth.js#L83

chenba:
`wantsauth`'s "login-successful" callback does not expose an account id or an FxA authenticated flag, so you'd need to add that in order for the update-model-then-rerender approach to work.

punamdahiya:
Adding FxA auth state by checking accountId cookie and returning that in savedAuthState should work. 
https://github.com/mozilla-services/screenshots/blob/master/static/js/wantsauth.js#L37

Will add a fetch call from wantsauth.js to check and store FxA state

punamdahiya:
@chenba Updated PR to save accountId in auth state retrieved from getAuthData() to use it inside login-successful event handlers  for homepage and shot. 
Shot index refreshes model with a subsequent /shots call to load shots data from db that help to load hasFxa state from model.js

chenba:
If I sign in from My Shots, the page still display the Sign In link and only shots from my current device after successful auth.  Is that page not covered by this PR?

server/src/pages/homepage/controller.js
```diff
@@ -5,7 +5,24 @@ let model;
 
 exports.launch = function(m) {
   model = m;
-  render();
+  if (window.wantsauth) {
+    if (window.wantsauth.getAuthData()) {
+      Object.assign(model, window.wantsauth.getAuthData());
+      model.hasFxa = !!model.accountId;
```
chenba: We don't really need both `hasFxa` and `accountId` in the model.

server/src/pages/settings/controller.js
```diff
@@ -20,7 +20,8 @@ exports.disconnectDevice = function() {
       window.alert("Error disconnecting: " + req.status + " " + req.statusText);
       window.Raven.captureException(new Error(`Error disconnecting: ${req.status} ${req.statusText}`));
     } else {
-      location.reload();
+      // Redirect to homepage
+      window.location = "/";
```
chenba: Why this change?  And would My Shots make more sense than the home page since we know this is a Screenshots user?

server/src/pages/shot/view.js
```diff
@@ -409,7 +410,10 @@ class Body extends React.Component {
 
     const noText = this.props.abTests && this.props.abTests.downloadText
                    && this.props.abTests.downloadText.value === "no-download-text";
-
+    const signIn = this.props.isOwner ?
+                    <div className="shot-fxA-signin">
+                      <SignInButton isAuthenticated={this.props.isFxaAuthenticated} initiatePage={this.props.shot._id} />
```
chenba: `this.props.shot.id` or `shot.id` to use the getter instead of accessing `_id`.

server/src/server.js
```diff
@@ -659,6 +659,23 @@ app.get("/api/check-login-cookie", function(req, res) {
...
+  }
+  const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip()});
+  const accountId = cookies.get("accountid", {signed: true});
+  if (accountId) {
+    simpleResponse(res, accountId, 200);
+  } else {
+    simpleResponse(res, null, 401);
+  }
+});
+
```
chenba: This endpoint isn't necessary.  `api/login` already look up the account id, so let's make use of that.  `auth.js` could get that for `sitehelper.js`, which then hands it to `wantsauth.js` on "login-successful".

server/src/server.js
```diff
@@ -1088,12 +1111,19 @@ app.get("/api/fxa-oauth/confirm-login", function(req, res, next) {
...
   }
+
   if (!req.query) {
     next(errors.missingParams());
     return;
   }
   const { code, state } = req.query;
-  checkState(req.deviceId, state).then(isValid => {
+  // Retrieve endpoint and 32-bit value used to verify if
+  // the redirect is authetntic. Use endpoint to redirect to page
```
chenba: nit: typo on "authetntic"

server/src/server.js
```diff
@@ -1088,12 +1111,19 @@ app.get("/api/fxa-oauth/confirm-login", function(req, res, next) {
...
+  // where the login request was initiated from.
+  const data = state.split(END_POINT_SEPARATOR);
```
chenba: nit: use the optional limit param for `split`

server/src/server.js
```diff
@@ -1112,7 +1142,9 @@ app.get("/api/fxa-oauth/confirm-login", function(req, res, next) {
               ua: req.headers["user-agent"],
             }).send();
           }
-          res.redirect("/settings");
+          // Redirect to endpoint
+          const pageUri = endpoint ? "/" + endpoint : "/";
```
chenba: Would My Shots make more sense as the default?  I don't know what the new home page is going to have for content, but I think users want to see their shots after logging in.

server/src/signin-button.js
```diff
@@ -0,0 +1,38 @@
...
+
+exports.SignInButton = class SignInButton extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    let signin;
+    if (this.props.isAuthenticated) {
+      signin = <a className="settings-button" href="/settings" aria-label="Settings">
```
chenba: @flodolo A `aria-label` attr should be localized just like any other attr?

server/src/signin-button.js
```diff
@@ -0,0 +1,38 @@
...
+          <span>Sign In</span>
+        </Localized>
+      </a>;
+    }
+    return signin;
+  }
+
+  onClickSignIn(event) {
+    window.location = "/api/fxa-oauth/login/" + this.props.initiatePage;
+    sendEvent("fxa-signin", this.props.initiatePage, {useBeacon: true});
```
chenba: This needs to be added to `metrics.md`.

server/src/signin-button.js
```diff
@@ -0,0 +1,38 @@
...
+        </Localized>
+      </a>;
+    } else {
+      signin = <a className="signin-button" onClick={ this.onClickSignIn.bind(this) } aria-label="Sign In">
+        <Localized id="gSignIn">
+          <span>Sign In</span>
+        </Localized>
+      </a>;
+    }
+    return signin;
```
chenba: Could return without declaring and assigning.  Just a style comment though. 😄

static/css/frame.scss
```diff
@@ -118,6 +118,14 @@
   padding-right: $grid-unit * 1.5;
 }
 
+.shot-fxA-signin {
```
chenba: The uppercase 'A' here seems random, and it's not consistent with the rest of the class names.

server/src/signin-button.js
flodolo: Yes, it should

server/src/server.js
punamdahiya: Redirecting to my shots for now until we get specific UX feedback on the flow

server/src/signin-button.js
punamdahiya: @chenba @flodolo Do we have existing settings aria-label localized?
https://github.com/mozilla-services/screenshots/blob/master/server/src/pages/shotindex/view.js#L209

Trying to find where do we need to add localized strings for aria-label?

chenba: It certainly looks like the only existing `aria-label` is not localized.  Thanks for finding it.  Could you add that as well?

punamdahiya:
@chenba I have updated PR by getting accountId from auth.js /api/login call. For shotindex page, to reliably have accountId we need to a)  update model with accountId returned from login-successful event listener, 
b) pass  model.accountId param to /shots?withdata=true&data=json server call that refreshes model , to pick accountId from model instead of request.accountId
https://github.com/mozilla-services/screenshots/blob/fxa-integration/server/src/pages/shotindex/server.js#L19

If there is an alternate approach , please provide your input. In the meantime updated PR refreshes model with half second delay to validate if it helps retrive accountId in request parameter

docs/METRICS.md
```diff
@@ -311,8 +311,9 @@ These are events that an add-on user can encounter on a shot they own
 33. [x] Visit an image directly, when the image isn't embedded directly in a Screenshots shot page, `web/visit/direct-view-owner`
 34. [x] View an image directly, when the image is being shown as part of a Facebook/Twitter style preview (the og:image or twitter:image), `web/visit/direct-view-embedded-owner`
 35. [x] Close new edit tools promotion dialog, `web/promo-closed`
-1. [x] Add shot to favorites `web/set-favorite/navbar`
-1. [x] Remove shot from favorites `web-unset-favorite/navbar`
+36. [x] Add shot to favorites `web/set-favorite/navbar`
+37. [x] Remove shot from favorites `web-unset-favorite/navbar`
+38. [x] Signin to Firefox Accounts `web/fxa-signin`
```
chenba: Markdown doesn't care about the actual number in the numbered lists btw.

punamdahiya: true, I believe its more for readability to increase number for a items listed in a group

punamdahiya:
@chenba Updated PR with my shots to use accountId  returned on login-successful. I have kept as separate commit for easy review. Thanks!

server/src/pages/homepage/controller.js
```diff
@@ -5,7 +5,22 @@ let model;
...
+        render();
+      });
+    }
+  }
+  try {
+    render();
+  } catch (e) {
+    window.Raven.captureException(e);
+    throw e;
+  }
```
chenba: Why is this `render()` call in a `try-catch` but the one above in the event callback is not?

server/src/pages/shotindex/controller.js
```diff
@@ -11,7 +11,8 @@ let firstRun = true;
 
 const queryParamModelPropertyMap = {
   q: "defaultSearch",
-  p: "pageNumber"
+  p: "pageNumber",
+  a: "accountId"
```
chenba: Why do we need this?

punamdahiya: This is used to pass accountId stored in model to server.js  here https://github.com/mozilla-services/screenshots/pull/4759/commits/bbe67aa5e02520f07df14fc6832b4a6222e0e25e#diff-78f59c6794bdfdf3aa07ed2363aa584dR17

server/src/pages/homepage/controller.js
punamdahiya: Followed pattern from shot page, for homepage it could be overkill

server/src/pages/shotindex/controller.js
chenba: Sure, that's where it is used.  But I don't think it's necessary.  The request to get the shots should have the account id in its cookies.

punamdahiya: got it, updated PR to remove passing accountId from queryParamModelPropertyMap

server/src/pages/shotindex/controller.js
```diff
@@ -25,33 +25,51 @@ exports.launch = function(m) {
...
+    } else {
+      window.wantsauth.addAuthDataListener((data) => {
+        Object.assign(model, data);
+        if (firstRun && m.shots === null) {
+          // The actual shot data hasn't been loaded yet, so we'll immediately request it:
+          refreshModel();
+        }
+        firstRun = false;
+        render();
+      });
```
chenba: By the time `refreshModel()` is called, the cookies should be available.  Otherwise the `reloaded` code would've executed.  So I don't think we need wrap `refreshModel` inside wantsauth.

punamdahiya: @chenba  reloaded code relies on deviceId which can return true if deviceId cookie is set and accountId is not ( before confirm-login redirection from Firefox Account authentication)

Is the ask above to remove refreshModel from addAuthDataListener or completely out of window.wantsauth as it exists right now in master

punamdahiya: Updated code to move refreshModel out of window.wantsauth. The code is still listening to wantsauth data listener and re-rendering code on authinfo retrieval

server/src/pages/homepage/view.js
```diff
@@ -16,6 +17,7 @@ class Head extends React.Component {
       <reactruntime.HeadTemplate {...this.props}>
         <link rel="stylesheet" href={ this.props.staticLink("/static/css/home.css") } />
         <script src={this.props.staticLink("/static/js/UITour-lib.js")} async></script>
+        <script src={ this.props.staticLink("/static/js/wantsauth.js") } />
```
chenba: There's a bug with the way auth currently works: the pages send a request to `/api/login` every time.  Once the client has the cookies, pages should not send login requests.  It looks like that's what My Shots meant to do with its conditional inclusion of `wantsauth`, but it's looking at a undefined property in the condition.

Let's do this:

- If there's a device id, then don't include `wantsauth`.
- If there's a device id but not account id, and the page wants to know if the user has FxA auth, then include `wantsauth`.

That means we need to add something to indicate that the page wants FxA auth after a successful FxA login.  A query param should work.

static/css/partials/_partials.scss
```diff
@@ -10,6 +10,7 @@
 @import "buttons";
 @import "large-icon-block";
 @import "share-panel";
+@import "signin-fxA";
```
chenba: The uppercase `A` here is inconsistent with the rest of the filenames.

server/src/pages/homepage/view.js
punamdahiya: @chenba since we are redirecting to /shots on successful login, we can safely remove wantsauth.js from homepage view and avoid /api/login calls

static/css/partials/_partials.scss
punamdahiya: Renamed to use lowercase 'a'

server/src/pages/homepage/view.js
chenba: We did discuss redirecting users back to wherever they init'd the login so that includes the homepage.  `/shots` was the fallback when there's no redirect uri.  Is that not the case?

And all the pages need to be fixed.  I was thinking doing that in this PR since we are knee deep in auth.  But I could file another bug if that's better.

punamdahiya: @chenba As discussed, will fix auth bug in this PR itself by moving conditional wantsauth.js include at one place in reactrender.js

Will update homepage to redirect user back to homepage after successful FxA login

chenba: hmm actually `reactruntime` would be more appropriate.

punamdahiya: @chenba Updated PR to fix auth bug and load wantsAuth.js inside reactruntime only when needed

server/src/pages/homepage/view.js
```diff
@@ -101,10 +103,8 @@ class Body extends React.Component {
                   <a href="#" className="screenshots-logo"></a>
                 </div>
                 <div className="nav-links">
-                  <Localized id="homePageGetStarted">
-                    <a href="#how-screenshots-works" onClick={ this.onClickGetStarted.bind(this) }>Get Started</a>
-                  </Localized>
```
chenba: If we are removing this anchor, then the handler, content, etc. should be removed as well.

server/src/pages/settings/controller.js
```diff
@@ -20,7 +20,8 @@ exports.disconnectDevice = function() {
       window.alert("Error disconnecting: " + req.status + " " + req.statusText);
       window.Raven.captureException(new Error(`Error disconnecting: ${req.status} ${req.statusText}`));
     } else {
-      location.reload();
+      // Redirect to homepage
```
chenba: This comment no longer matches the line below.

punamdahiya:
Fixes #4819

server/src/pages/shot/controller.js
```diff
@@ -43,10 +43,12 @@ exports.launch = function(data) {
     if (window.wantsauth.getAuthData()) {
       Object.assign(model, window.wantsauth.getAuthData());
       model.isExtInstalled = true;
+      model.isFxaAuthenticated = model.accountId && model.accountId === model.shotAccountId;
     } else {
       window.wantsauth.addAuthDataListener((data) => {
         Object.assign(model, data);
         model.isExtInstalled = true;
+        model.isFxaAuthenticated = model.accountId && model.accountId === model.shotAccountId;
```
chenba: These lines are nearly identical to those in the `if`.  Consider putting them into a function.

server/src/pages/shotindex/controller.js
```diff
@@ -25,31 +25,45 @@ exports.launch = function(m) {
...
+    }
+
+    if (window.wantsauth.getAuthData()) {
+      Object.assign(model, window.wantsauth.getAuthData());
+    } else {
+      window.wantsauth.addAuthDataListener((data) => {
+        Object.assign(model, data);
+        render();
+      });
+    }
```
chenba: This if-else will execute on `refreshModel()`, which is not necessary.

server/src/pages/homepage/view.js
chenba: Oh, the l10n string for `homePageGetStarted` too.

server/src/pages/shotindex/controller.js
punamdahiya: @chenba Updated shotindex to getAuthData only if both req.accountId (value stored in model.hasFxa) and model.accountId is null, ensuring this code is not exceuted on refreshModel

server/src/pages/homepage/view.js
punamdahiya: Done

punamdahiya:
@chenba PR updated with feedback. Please review. Thanks!

server/src/pages/homepage/controller.js
```diff
@@ -5,9 +5,27 @@ let model;
...
+    }
+  }
   render();
 };
 
+function updateModel() {
+  if (!model.hasFxa) {
+    model.hasFxa = !!model.accountId;
+  }
+}
```
chenba: nit: since this function might not be invoked at all, perhaps it's better as a function expression.

server/src/pages/homepage/view.js
chenba: @johngruen In today's meeting we agreed to start merging FxA patches into master, so I would like your take on this change.  It removes the Get Started nav and content from the home page.

chenba: @punamdahiya The content is still here.

punamdahiya: @chenba by content you mean below text on homepage. I believe that should be removed as part of homepage re-design @johngruen wdyt?
screenshot_2018-08-30 firefox screenshots

server/src/pages/shotindex/controller.js
```diff
@@ -25,33 +25,55 @@ exports.launch = function(m) {
...
     }
-    const authTimeout = setTimeout(() => {
-      // eslint-disable-next-line no-global-assign, no-native-reassign
-      location = location.origin + "/#tour";
-    }, FIVE_SECONDS);
-    window.wantsauth.addAuthDataListener((data) => {
-      if (location.search.indexOf("reloaded") > -1) {
-        return;
+
+    if (!model.accountId && !model.hasFxa) {
```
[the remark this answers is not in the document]
punamdahiya: Check here ensures we don't call if-else block inside it on refreshModel. We do need this if-else block for first-run on redirection from FxA login page.
1) On redirection first-run /shots call , model.hasFxa is false ( req.accountId doesn't exist) and model.accountId is still undefined and will be set by authdata returns from wantsauth API inside if else block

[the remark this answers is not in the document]
punamdahiya: Will add comment above this line to help explain this check

server/src/pages/shotindex/controller.js
```diff
@@ -25,33 +25,55 @@ exports.launch = function(m) {
...
+      if (window.wantsauth.getAuthData()) {
+        Object.assign(model, window.wantsauth.getAuthData());
+        updateModel();
+      } else {
+        window.wantsauth.addAuthDataListener((data) => {
+          Object.assign(model, data);
+          updateModel();
+          render();
```
chenba: Let's replace this `render` with `refreshModel`...

server/src/pages/shotindex/controller.js
```diff
@@ -25,33 +25,55 @@ exports.launch = function(m) {
...
+        });
       }
-      clearTimeout(authTimeout);
-      location.search = "reloaded";
-    });
+    }
+  }
+
+  if (firstRun && m.shots === null) {
```
chenba: ...and make this an `else-if` of https://github.com/mozilla-services/screenshots/pull/4759/files#diff-6db5e3f6a06e986e2aa9ab0f5c0170d1R30

server/src/pages/shotindex/controller.js
punamdahiya: Not sure if I get it, putting refreshModel inside window.wantsauth listener means we need to load wantsauth API everytime which we  don't want

server/src/pages/homepage/view.js
chenba: @punamdahiya hmm yeah, that'd be better.  Let's leave the page content alone then.

server/src/reactruntime.js
```diff
@@ -37,6 +37,7 @@ exports.HeadTemplate = class HeadTemplate extends React.Component {
       localeScripts.push(<script key={`l10n-${locale}`} src={this.props.staticLink(`/static/locales/${locale}.js`)} />);
     }
 
+    const wantsAuth = (!this.props.authenticated) || (this.props.authenticated && !this.props.hasFxa && this.props.authFxa);
```
chenba: nit: could use a newline here.  Having that long line atop of the return looks a little weird.

server/src/server.js
```diff
@@ -1112,7 +1126,12 @@ app.get("/api/fxa-oauth/confirm-login", function(req, res, next) {
               ua: req.headers["user-agent"],
             }).send();
           }
-          res.redirect("/settings");
+          // Redirect to endpoint with auth param indicating successful Fxa auth flow.
+          // 'auth' param is used in reactrender and reactruntime to load wantsauth.js
+          // and display Fxa Sign button state when request doesn't have accountId
```
chenba: "In" is missing after "Sign" here.

server/src/reactruntime.js
punamdahiya: Done

server/src/pages/homepage/controller.js
```diff
@@ -5,9 +5,26 @@ let model;
...
+        render();
+      });
+    }
+  }
   render();
 };
 
+function updateModel(authData) {
+  Object.assign(model, authData);
+  if (!model.hasFxa) {
```
chenba: Since this function is only called on `wantsauth`, I don't think we need this `if`.  We want to set or update `hasFxa` upon receiving auth data.

punamdahiya: +1

server/src/pages/shotindex/controller.js
```diff
@@ -25,33 +25,56 @@ exports.launch = function(m) {
...
-    window.wantsauth.addAuthDataListener((data) => {
-      if (location.search.indexOf("reloaded") > -1) {
-        return;
+    // On redirection after FxA login, check if model.hasFxa set in reactrender and
+    // model.accountId is unavailable. If yes, update model with savedAuthData returned
+    // from wantsauth API
+    if (!model.accountId && !model.hasFxa) {
+      if (window.wantsauth.getAuthData()) {
+        updateModel(window.wantsauth.getAuthData());
+        firstRun = false;
```
chenba: Since line 68 below always execute, we don't need this here or the line in the else below.

punamdahiya:
@chenba Updated PR with last few feedbacks and squashed commit. Thanks!

server/src/pages/shotindex/view.js
```diff
@@ -42,7 +41,7 @@ class Body extends React.Component {
         <div className="column-space full-height default-color-scheme" id="shot-index-page">
           <div id="shot-index-header" className="header">
             <h1><a href="/shots">Firefox <strong>Screenshots</strong> <sup>Beta</sup></a></h1>
-            {this.props.enableUserSettings ? this.renderSettingsPage() : null}
+            {this.props.enableUserSettings && this.props.hasDeviceId ? this.renderFxASignIn() : null}
```
chenba: After signing in from My Shots, I get teh following mark up for the Setting link:

image

Note that's "Settings" in the Sign In anchor.  And it was not clickable.  I think `SignInButton` probably need to (re)render based on its state, not props.

punamdahiya: @chenba Updated PR with fix of this issue by checking isAuthenticated prop inside click handler.  
As suggested, using state to force re-render doesn't seem to help here even after state is assigned to updated prop value inside getDerivedStateFromProps

server/src/signin-button.js
```diff
@@ -11,25 +11,29 @@ exports.SignInButton = class SignInButton extends React.Component {
   render() {
     if (this.props.isAuthenticated) {
       return <Localized id="settingsButton">
-        <a className="settings-button" href="/settings" aria-label="Settings">
+        <a className="settings-button" onClick={ this.clickHandler.bind(this) } aria-label="Settings">
```
chenba: The markup is still incorrect with this change.  Try changing the `if` condition above to base on a state property that derives from `this.props.isAuthenticated`.

punamdahiya: Updated to use state though it doesn't help with markup. Is there any way to force re-render

server/src/signin-button.js
```diff
@@ -6,10 +6,20 @@ const sendEvent = require("./browser-send-event.js");
 exports.SignInButton = class SignInButton extends React.Component {
   constructor(props) {
     super(props);
+    this.state = {
+      displaySettings: this.props.isAuthenticated
```
chenba: Just `props` here, not `this.props`.

server/src/signin-button.js
```diff
@@ -6,10 +6,20 @@ const sendEvent = require("./browser-send-event.js");
...
+    };
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
+    const newState = {
```
chenba: nit: don't really need to declare and assign a var.

server/src/signin-button.js
```diff
@@ -0,0 +1,50 @@
...
+        <Localized id="gSignIn">
+          <span>Sign In</span>
+        </Localized>
+      </a>
+    </Localized>;
+  }
+
+  clickHandler(event) {
+    if (this.props.isAuthenticated) {
+      window.location = "/settings";
```
chenba: We don't really need when the correct markup is rendered for Settings.

punamdahiya: Updated with squashed commits

server/src/pages/shot/view.js
```diff
@@ -347,14 +346,13 @@ class Body extends React.Component {
...
-        className={`button favorite ${activeFavClass}`}
-        title="Favorite this shot"
-        onClick={ this.onClickFavorite.bind(this) }
-      /></div>;
-    }
+    const activeFavClass = this.props.expireTime ? "" : "is-fav";
+    const shouldShow = this.props.isFxaAuthenticated ? "" : "hidden";
+    favoriteShotButton = <div className="fav-wrapper"><button
+      className={`button favorite ${shouldShow} ${activeFavClass}`}
+      title="Favorite this shot"
```
chenba: Do you mind adding the l10n for this title attribute that was missed in the original patch?

server/src/signin-button.js
```diff
@@ -0,0 +1,46 @@
...
+      <a className="signin-button" onClick={ this.clickHandler.bind(this) } aria-label="Sign In">
+        <Localized id="gSignIn">
+          <span>Sign In</span>
+        </Localized>
+      </a>
+    </Localized>;
+  }
+
+  clickHandler(event) {
+    window.location = "/api/fxa-oauth/login/" + this.props.initiatePage;
```
chenba: This can simply be a href.  Also, the order here _might_ be an issue: the page is navigated away before the event is sent.  Although no browser and server are that quick I guess.

static/css/frame.scss
```diff
@@ -118,6 +118,14 @@
...
 }
 
+.shot-fxa-signin {
+  a.signin-button,
+  a.settings-button {
+    position: relative;
+    padding: 10px;
+  }
+}
+
```
chenba: Move these into `_signin-fxa.scss`?

punamdahiya: Done

PR #4805, opened 2018-08-22
Add logs to Watchdog callbacks

### LINKED ISSUE #4768: [Watchdog] Add More Logging on Callbacks
@6a68 suggested in https://github.com/mozilla-services/screenshots/issues/4721#issuecomment-409027065 that we should log the results of Watchdog callbacks.

If negative match: debug level message.
If positive match: warning level message.

chenba:
Fixes #4768

### DIFF
--- a/server/src/watchdog.js
+++ b/server/src/watchdog.js
@@ -240,10 +240,12 @@ exports.handleResult = function(req) {
 };
 
 function handleNegative(record) {
+  mozlog.debug("watchdog-negative-match", {msg: `Watchdog submission ${record.id} is a negative match.`});
   return db.update("UPDATE watchdog_submissions SET positive_result = FALSE WHERE id = $1", [record.id]);
 }
 
 function handlePositive(record) {
+  mozlog.info("watchdog-positive-match", {msg: `Watchdog submission ${record.id} is a positive match.`});
   return db.transaction(client => {
     return db.queryWithClient(
       client,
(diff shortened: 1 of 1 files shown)


server/src/watchdog.js
```diff
@@ -240,10 +240,12 @@ exports.handleResult = function(req) {
 };
 
 function handleNegative(record) {
+  mozlog.debug("watchdog-negative-match", {msg: `Watchdog submission ${record.id} is a negative match.`});
   return db.update("UPDATE watchdog_submissions SET positive_result = FALSE WHERE id = $1", [record.id]);
 }
 
 function handlePositive(record) {
+  mozlog.warn("watchdog-positive-match", {msg: `Watchdog submission ${record.id} is a positive match.`});
```
6a68: Although a positive match is important, it's not actually an error state, so use `info` instead of `warn` here

server/src/watchdog.js
```diff
@@ -240,10 +240,12 @@ exports.handleResult = function(req) {
...
   return db.transaction(client => {
     return db.queryWithClient(
       client,
```
6a68: Suggestion: since we're in here anyway, it might be useful to log the outcome of the transaction.

Maybe something like:
```js
return db.transaction(...).then(() => { 
  mozlog.info(`Watchdog positive match ${record.id} successfully committed to the database.`)
}).catch(err => {
  mozlog.error(`Watchdog positive match ${record.id} not committed to the database: ${err}.`);
  throw err;
})
```

PR #4808, opened 2018-08-22
Use Time Difference Amount Not Timestamp when Saving Expiration on Unfavorite

### LINKED ISSUE #4807: [FxA] Incorrect Default Expiration When a Shot is Unfavorited
I introduced a bug in #4688 because I can't unit.

chenba:
Fixes #4807

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -539,8 +539,7 @@ class Body extends React.Component {
       this.props.controller.changeShotExpiration(this.props.shot, INDEFINITE);
     } else {
       sendEvent("unset-favorite", "navbar");
-      const TWO_WEEKS_IN_MS = 1209600000;
-      this.props.controller.changeShotExpiration(this.props.shot, Date.now() + TWO_WEEKS_IN_MS);
+      this.props.controller.changeShotExpiration(this.props.shot, this.props.defaultExpiration);
     }
   }
 


server/src/pages/shot/view.js
```diff
@@ -539,8 +539,8 @@ class Body extends React.Component {
       this.props.controller.changeShotExpiration(this.props.shot, INDEFINITE);
     } else {
       sendEvent("unset-favorite", "navbar");
-      const TWO_WEEKS_IN_MS = 1209600000;
-      this.props.controller.changeShotExpiration(this.props.shot, Date.now() + TWO_WEEKS_IN_MS);
+      const TWO_WEEKS_IN_SECONDS = 1209600;
+      this.props.controller.changeShotExpiration(this.props.shot, Date.now() / 1000 + TWO_WEEKS_IN_SECONDS);
```
punamdahiya: @chenba Are you sure it shoule be in seconds looking at code in controller, expire time is in milliseconds 
https://github.com/mozilla-services/screenshots/blob/fxa-integration/server/src/pages/shot/controller.js#L106

https://github.com/mozilla-services/screenshots/blob/fxa-integration/server/src/pages/shot/model.js#L37

chenba: Right, which was why I changed  it to ms when you pointed out the unit mismatch in the review of the PR for #4688.  The real bug is that the value should be the number of ms from now, not a timestamp in ms.  I'll update this PR.  Thanks!

server/src/pages/shot/view.js
```diff
@@ -540,7 +540,7 @@ class Body extends React.Component {
     } else {
       sendEvent("unset-favorite", "navbar");
       const TWO_WEEKS_IN_MS = 1209600000;
```
punamdahiya: nit: It will be more readable if we assign 14 days value to this const as 60 x60x 24 x 14 x 1000 similar to here https://github.com/mozilla-services/screenshots/blob/master/server/src/config.js#L202

chenba: Updated the patch to use the default.  Thanks.

punamdahiya: LGTM!

PR #4809, opened 2018-08-22
Move the ESR check from bootstrap into webextension API

### LINKED ISSUE #4802: Move ESR check to webextension API
Related to #4800 

Rather than migrate off of bootstrap _and_ enable ESR uploads for Firefox 64, we could instead just keep ESR uploads disabled, and move the 'is ESR?' check to the same API experiment we'll use to check the disabled prefs.

6a68:
Fixes #4802.

* Add a webextension experiment API to Screenshots that exposes app update channel
* Move ESR check from bootstrap to webextension API

Unfortunately, ESR 60 doesn't allow legacy extensions, so there's no way to test this works in that case. But you can set the debugger and see that it returns the correct channel if you test from nightly :-\

### DIFF
--- a/addon/webextension/background/selectorLoader.js
+++ b/addon/webextension/background/selectorLoader.js
@@ -94,14 +94,16 @@ this.selectorLoader = (function() {
   function downloadOnlyCheck(tabId) {
     return communication.sendToBootstrap("isHistoryEnabled").then((historyEnabled) => {
       return communication.sendToBootstrap("isUploadDisabled").then((uploadDisabled) => {
-        return browser.tabs.get(tabId).then(tab => {
-          const downloadOnly = !historyEnabled || uploadDisabled || tab.incognito;
-          return browser.tabs.executeScript(tabId, {
-            // Note: `window` here refers to a global accessible to content
-            // scripts, but not the scripts in the underlying page. For more
-            // details, see https://mdn.io/WebExtensions/Content_scripts#Content_script_environment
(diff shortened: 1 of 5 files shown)


addon/webextension/background/selectorLoader.js
```diff
@@ -94,14 +94,16 @@ this.selectorLoader = (function() {
...
-          const downloadOnly = !historyEnabled || uploadDisabled || tab.incognito;
-          return browser.tabs.executeScript(tabId, {
-            // Note: `window` here refers to a global accessible to content
-            // scripts, but not the scripts in the underlying page. For more
-            // details, see https://mdn.io/WebExtensions/Content_scripts#Content_script_environment
-            code: `window.downloadOnly = ${downloadOnly}`,
-            runAt: "document_start"
+        return browser.experiments.screenshots.getUpdateChannel().then((channel) => {
+          return browser.tabs.get(tabId).then(tab => {
+            const downloadOnly = !historyEnabled || uploadDisabled || channel === 'esr' || tab.incognito;
```
chenba: Linter is complaining about those single quotes around `esr`.

addon/webextension/experiments/screenshots/api.js
```diff
@@ -0,0 +1,34 @@
+/* globals AppConstants */
```
chenba: I think we need `ExtensionAPI` in the globals as well.

PR #4810, opened 2018-08-23
Check some prefs from the API, not from bootstrap

### LINKED ISSUE #4531: Replace history.enabled pref with privacy API
Related to #3875 

We can replace the history.enabled pref check with this API:

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/privacy/websites

6a68:
Move the history enabled and upload-disabled checks from bootstrap into the API.

It might make sense to just check all the prefs in a single API call, but this works for now.

This commit fixes #4531, and depends on the two commits from PR #4809.

### DIFF
--- a/addon/webextension/experiments/screenshots/api.js
+++ b/addon/webextension/experiments/screenshots/api.js
@@ -4,6 +4,8 @@
 
 ChromeUtils.defineModuleGetter(this, "AppConstants",
                                "resource://gre/modules/AppConstants.jsm");
+ChromeUtils.defineModuleGetter(this, "Services",
+                               "resource://gre/modules/Services.jsm");
 
 this.screenshots = class extends ExtensionAPI {
   getAPI() {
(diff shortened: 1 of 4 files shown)


addon/webextension/experiments/screenshots/api.js
```diff
@@ -0,0 +1,42 @@
...
+          // Standard builds:
+          //   'esr' - ESR channel
+          //   'release' - release channel
+          //   'beta' - beta channel
+          //   'nightly' - nightly channel
+          // Non-standard / deprecated builds:
+          //   'aurora' - deprecated aurora channel (still observed in dxr)
+          //   'default' - local builds from source
+          //   'nightly-try' - nightly Try builds (QA may occasionally need to test with these)
+          async getUpdateChannel() {
```
ianb: Do these all need to be async?

6a68: Nope, I copied [the examples](https://webextensions-experiments.readthedocs.io/en/latest/basics.html#how-do-they-work) a little too closely. Since experimental APIs extend the same base `ExtensionAPI` class as built-in APIs, and the `async` bit is definitely [not required there](https://searchfox.org/mozilla-central/source/toolkit/components/extensions/parent/ext-telemetry.js#30), must not be required here, either.

PR #4818, opened 2018-08-27
property is a white space separated list

da2x:
The `property` attribute is [a white space separated list of values](https://www.w3.org/TR/2008/REC-rdfa-syntax-20081014/#A-property) and not a single value. CSS /querySelectors can handle this using [~=](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors).

Some examples and [more context](https://www.ctrl.blog/entry/rdfa-socialmedia-metadata):

    <meta property="dc:publisher og:site_name" content="Allizom Media">
    <meta property="article:modified_time dc:modified og:updated_time" content="2018-08-28">

### DIFF
--- a/addon/webextension/selector/documentMetadata.js
+++ b/addon/webextension/selector/documentMetadata.js
@@ -3,12 +3,12 @@
 this.documentMetadata = (function() {
 
   function findSiteName() {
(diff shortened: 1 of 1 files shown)


chenba:
@da2x Thanks for the patch!

PR #4821, opened 2018-08-29
Fixes #4685 - [FxA] Common Header Component

### LINKED ISSUE #4685: [FxA] Implement Common Header Component for All Site UI
Looking at the [new spec](https://mozilla.github.io/testpilot-assets/Firefox_Screenshots/NEW_FxA_Integration/), there's a common component used as at site header/nav. We should implement while migrating to FxA

Refer to Shot View, My Shots and Home Page specs

punamdahiya:


punamdahiya:
Updated PR with common header for homepage and My Shots. Looking into complexity of bringing in shots header into common component

server/src/header.js
```diff
@@ -0,0 +1,63 @@
...
+      myShots = <button className="nav-button" onClick={ this.onClickMyShots.bind(this) } tabIndex="0">
+        <div className="icon icon-shots"></div>
+        <Localized id="gMyShots">
+          <span>My Shots</span>
+        </Localized>
+      </button>;
+    }
+
+    const signin = this.renderFxASignIn();
+    const search = this.props.searchForm && this.props.initiatePage === "shots" ? this.props.searchForm : null;
```
chenba: Just want to +1 on Jared's concern regarding how and where the content of the header is inserted.  We're already special casing for My Shots here; it doesn't feel right.

punamdahiya: I agree it does seem a bit off to have searchForm moved to Header component when rest of its styles and event handlers are in shotIndex view. Will update to move SearchForm rendering outside of Header component

server/src/pages/homepage/view.js
```diff
@@ -79,46 +74,24 @@ class Body extends React.Component {
...
-              <div className="banner-container">
-                <div className="banner-content">
-                  <h1>Firefox Screenshots</h1>
-                  <Localized id="gScreenshotsDescription">
-                    <p>Screenshots made simple. Take, save, and share screenshots without leaving Firefox.</p>
-                  </Localized>
-                  { this.renderGetFirefox() }
-                </div>
-                <div className="banner-image-front" />
+          <Header hasDeviceId={this.props.showMyShots} isFirefox={this.props.isFirefox} hasFxa={this.props.hasFxa} initiatePage="" />
```
chenba: nit: why not just `hasDeviceId` or `showMyShots` all the way down?

server/src/pages/shot/view.js
```diff
@@ -346,13 +346,15 @@ class Body extends React.Component {
     let editButton;
     const highlight = this.state.highlightEditButton ? <div className="edit-highlight" onClick={ this.onClickEdit.bind(this) } onMouseOver={ this.onMouseOverHighlight.bind(this) } onMouseOut={ this.onMouseOutHighlight.bind(this) }></div> : null;
 
-    const activeFavClass = this.props.expireTime ? "" : "is-fav";
+    const activeFavClass = this.props.expireTime ? "un-fav" : "is-fav";
```
chenba: `un-fav` only affects the span and `is-fav` only affects the div.  It's confusing when class names are unnecessarily added to elements.  Why not have the styles of `un-fav` be the default and then `is-fav` to toggle?  Then we don't even need to change this line.

server/src/pages/shot/view.js
```diff
@@ -426,25 +442,12 @@ class Body extends React.Component {
...
-                { !noText &&
-                    <Localized id="shotPageDownload"><span className="download-text">Download</span></Localized> }
-              </a>
-            </Localized>
-            { signIn }
-          </div>
-        </div>
+          { favoriteShotButton }
+          { trashOrFlagButton }
+          { this.props.enableAnnotations ? editButton : null }
```
chenba: Perhaps move this check up to the assignment of `editButton`.

6a68:
Just to capture what we discussed in IRC:

Seems like the Header is currently doing a lot. It renders header-related stuff for all of the pages, and also contains page-specific code to decide which parts to render.

One way to simplify things might be to separate a generic Header, that renders any sub-components (passed as props) into a region in the DOM, from page-specific headers that know how to create and render page-specific components into the generic Header. For instance, a ShotPageHeader might render itself by creating components for the shot info and shot page buttons, then passing all those components into the Header, actually returning the Header.

Any components/UI found on multiple pages, like the blue upsell bar, or the login/settings button, could either be part of the generic Header, or be sub-components on their own. UI that only occurs on a single page (like the favorite button) could just be elements created in the specific page's PageHeader, or you could go further and make each control its own component (might be a lot of work for a first pass, though).

I'm not sure this is the best way to break up the complexity, but I think it's one way that could work. There are some nice examples of composing a generic Dialog class with specific Dialog subtypes in the [React docs](https://reactjs.org/docs/composition-vs-inheritance.html). Happy to chat further, here or on IRC/Slack :+1:

punamdahiya:
@chenba @6a68 Thanks for feedback. Updated PR with review comments and below items are in progress
- Download button
- Shot page main-actions style
-  Update buttons and localize strings ( ensuring big strings e.g. in German doesn’t break layout)
- Review My Shots and Shot page responsive flow
- Refactor code to use specific header component ShotPageHeader, MyShotsHeader, HomePageHeader to render generic Header

chenba:
@punamdahiya CI failure is caused by linting errors.

server/src/header.js
```diff
@@ -0,0 +1,58 @@
...
+    super(props);
+  }
+
+  onClickMyShots() {
+    sendEvent("goto-myshots", "homepage", {useBeacon: true});
+    window.location = "/shots";
+  }
+
+  renderFxASignIn() {
+    if (!this.props.initialPage) {
```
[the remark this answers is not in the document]
punamdahiya: It's checking for both and in either case signIn button on auth redirection will redirect back to homepage which I believe should be ok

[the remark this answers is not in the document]
punamdahiya: With specific headers, this check is no more needed

chenba:
With all the auth related info being passed as props in multiple places, I think this patch is a good candidate for React's [Context](https://reactjs.org/docs/context.html).

punamdahiya:
@chenba PR is updated with items listed in https://github.com/mozilla-services/screenshots/pull/4821#issuecomment-419334658. Please review. Thanks!

@6a68 @Ian Including you in PR for feedback/inputs. Thanks

server/src/pages/shot/shotpage-header.js
```diff
@@ -0,0 +1,187 @@
...
+    const signin = this.renderFxASignIn();
+    const shotInfo = this.renderShotInfo();
+    return (
+      <Header>
+        { myShotsText }
+        { shotInfo }
+        <div className="shot-alt-actions">
+        { this.props.children }
+        { signin }
+        </div>
```
punamdahiya: Shot page specific header is rendering AllShots, Shot title and SignIn button. This has moved ~120 lines of code from Shot page view to ShotPageHeader. To keep this PR simple I have kept  Favorite, Download, Delete, Edit button with their new UI in view page. Ideally these should be moved in their own components and invoked from ShotPageHeader which can be done in subsequent iterations

static/css/shot-index.scss
```diff
@@ -23,50 +21,31 @@
...
-  }
 
-  @include respond-to("medium") {
-    @include flex-container(row, center, center);
-  }
+#search-form {
+  @include flex-container(row, flex-start, stretch);
+  position: absolute;
+  right: 150px;
+  top: 40px;
```
punamdahiya: Only position of search form has changed, rest of its styles stays as is

punamdahiya:
> With all the auth related info being passed as props in multiple places, I think this patch is a good candidate for React's [Context](https://reactjs.org/docs/context.html).

Creating an auth-context should be useful.  auth-context Provider will be filled in respective view files (homepage, myshots, shot) with FxA state to be consumed in SignIn button as shown in [this example](https://reactjs.org/docs/context.html#dynamic-context) and marketing blue header . 

My understanding benefit to switching to context in this PR is avoid passing auth state as props to signIn button. I can update that as separate commit so that easier to weigh in complexity of this change.

server/src/header.js
```diff
@@ -0,0 +1,33 @@
...
+exports.Header = class Header extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    if (!this.props.hasLogo) {
+      return <header className="header-panel default-color-scheme">
+        {this.props.children}
+      </header>;
```
chenba: I'm usually a fan of early returns, but in this case `const logo = maybe ? <blah> :  null;` then `{ logo }` reads better.  It's also unclear if no logo means no `alt-actions` wrap, or that's just missing.

server/src/pages/homepage/homepage-header.js
```diff
@@ -0,0 +1,51 @@
...
+  renderFxASignIn() {
+      return (
+        this.props.isFirefox && this.props.hasDeviceId ?
+          <SignInButton isAuthenticated={this.props.hasFxa} initialPage="" /> : null
+      );
+  }
+
+  render() {
+    let myShots;
+    if (this.props.hasDeviceId) {
```
chenba: This should be an isOwner check.

server/src/pages/homepage/homepage-header.js
```diff
@@ -0,0 +1,51 @@
...
+      myShots = <Localized id="shotIndexPageMyShotsButton">
+        <button className="nav-button icon-shots" title= "My Shots" onClick={ this.onClickMyShots.bind(this) } tabIndex="0">
```
chenba: There's a space after `title=`.

server/src/pages/shot/shotpage-header.js
```diff
@@ -0,0 +1,187 @@
...
+        <button className="nav-button" title="Homepage" onClick={this.onClickMyShots.bind(this)}>
+          <div className="back-to-home">
+            <h2>
+              Firefox
+            </h2>
+            <h3>
+              Screenshots
+            </h3>
+          </div>
+        </button>
```
chenba: These links should use `<a>` and `hrefs`.  Buttons shouldn't have divs and  headings.

server/src/pages/shot/shotpage-header.js
```diff
@@ -0,0 +1,187 @@
...
+
+  onClickMyShots() {
+    if (this.props.isOwner) {
+      sendEvent("goto-myshots", "navbar", {useBeacon: true});
+      location.href = "/shots";
+    } else {
+      sendEvent("goto-homepage", "navbar", {useBeacon: true});
+      location.href = "/";
+    }
+    // Note: we allow the default action to continue
```
chenba: This comment does not make sense as  there is no default action.

server/src/pages/shot/view.js
```diff
@@ -337,57 +336,58 @@ class Body extends React.Component {
...
     const shouldShow = this.props.isFxaAuthenticated ? "" : "hidden";
-    favoriteShotButton = <div className="fav-wrapper"><Localized id="shotPagefavoriteButton"><button
-      className={`button favorite ${shouldShow} ${activeFavClass}`}
-      title="Favorite this shot"
-      onClick={ this.onClickFavorite.bind(this) }
-    /></Localized></div>;
+
+    favoriteShotButton = <Localized id="shotPageFavoriteButton">
+      <button className={`nav-button ${shouldShow} `} onClick={this.onClickFavorite.bind(this)}>
+        <div className={`icon-favorite favorite ${activeFavClass}`} ></div>
```
chenba: Is it not possible to achieve the style we want by wrapping the button in the div, not vice-versa?

server/src/pages/shotindex/myshots-header.js
```diff
@@ -0,0 +1,32 @@
...
+
+  render() {
+    const signin = this.renderFxASignIn();
+    return (
+      <Header hasLogo={true}>
+        { signin }
+      </Header>
+    );
+  }
+};
```
chenba: This component can be a stateless component as well.

server/src/signin-button.js
```diff
@@ -18,30 +18,35 @@ exports.SignInButton = class SignInButton extends React.Component {
   render() {
     if (this.state.displaySettings) {
       return <Localized id="settingsButton">
-        <a className="settings-button" href="/settings" aria-label="Settings">
+        <button className="nav-button icon-settings" tabIndex="0" title="Settings" onClick={this.clickHandler.bind(this)}>
```
chenba: This should remain a link; it's navigation.

server/src/signin-button.js
```diff
@@ -18,30 +18,35 @@ exports.SignInButton = class SignInButton extends React.Component {
...
-          <span>Sign In</span>
-        </Localized>
-      </a>
-    </Localized>;
+        <button className="nav-button icon-settings" tabIndex="0" title="SignIn" onClick={this.clickHandler.bind(this)}>
+          <Localized id="gSignIn">
+            <span>Sign In</span>
+          </Localized>
+        </button>
+      </Localized>;
```
chenba: It should be a link here as well.

static/css/partials/_header.scss
```diff
@@ -0,0 +1,217 @@
...
+    height: 100%;
+  }
+
+  @include respond-to("large") {
+    @include flex-container(row, center, center);
+  }
+
+  display: flex;
+  height: 96px;
+  background-color: #f9f9fa;
```
chenba: `$light-default` can be used here instead of the hex value.

server/src/signin-button.js
punamdahiya: Updated to button for its advantages around built-in keyboard accessibility that lets  tab between and activate using Return/Enter

chenba: Using links navigate the web is fundamental and I don't agree we should do something else.  Also, I can tab among links and navigate by pressing Enter.  Is that not the case on macOS Firefox?

punamdahiya: It works once we provide href, will update to use navigate link. Thanks

server/src/header.js
```diff
@@ -0,0 +1,21 @@
...
+
+  return <header className="header-panel default-color-scheme">
+      {logo}
+      {props.children}
+    </header>;
+};
+
+exports.Header.propTypes = {
+  hasLogo: PropTypes.bool,
+  children: PropTypes.node.isRequired
```
chenba: Now that #4822 is a thing, this line needs a comma.

server/src/pages/homepage/homepage-header.js
```diff
@@ -0,0 +1,53 @@
...
+const { Header } = require("../../header.js");
+
+exports.HomePageHeader = class HomePageHeader extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  onClickMyShots() {
+    sendEvent("goto-myshots", "homepage", {useBeacon: true});
+    window.location = "/shots";
```
chenba: Please use a link for navigation.

server/src/pages/homepage/homepage-header.js
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+      </Header>
+    );
+  }
+};
+
+exports.HomePageHeader.propTypes = {
+  hasFxa: PropTypes.bool,
+  isOwner: PropTypes.bool,
+  isFirefox: PropTypes.bool
```
chenba: Need dangling comma here.

server/src/pages/homepage/homepage-header.js
punamdahiya: +1, updated and rebased

punamdahiya:
@flodolo Can you please review localization changes in the PR. Thanks!

locales/en-US/server.ftl
```diff
@@ -11,10 +11,13 @@ gSettings = Settings
 gSignIn = Sign In
 
 ## Header
-signInButton =
-    .aria-label = Sign In
 settingsButton =
```
flodolo: If you change attributes in an existing string, you need to change the ID of the string.

locales/en-US/server.ftl
```diff
@@ -123,9 +126,17 @@ shotPageDownloadShot =
     .title = Download
 shotPageEditButton =
     .title = Edit this image
-shotPagefavoriteButton =
+shotPageFavoriteButton =
```
flodolo: This would require localizers to retranslate the string. If it's just for cosmetic reasons, I would avoid changing the ID.

locales/en-US/server.ftl
```diff
@@ -123,9 +126,17 @@ shotPageDownloadShot =
...
     .title = Favorite this shot
+shotPageBackToHomeButton =
+    .title = Homepage
+shotPageAllShotsButton =
+    .title = All Shots
+shotPageAllShots = All Shots
 shotPageDownload = Download
+shotPageDraw = Draw
```
flodolo: Is this a verb or a noun? I assume a verb (action), but please add a comment.

locales/en-US/server.ftl
```diff
@@ -123,9 +126,17 @@ shotPageDownloadShot =
...
+shotPageFavorite = Favorite
```
flodolo: Same question: verb or noun?

locales/en-US/server.ftl
```diff
@@ -11,10 +11,13 @@ gSettings = Settings
 gSignIn = Sign In
 
 ## Header
-signInButton =
-    .aria-label = Sign In
 settingsButton =
-    .aria-label = Settings
+    .title = Settings
```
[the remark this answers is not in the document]
punamdahiya: It's to keep buttons consistent. title should provide ARIA support and shows the text of title as a tooltip

locales/en-US/server.ftl

locales/en-US/server.ftl
punamdahiya: It's a verb, added comment to highlight that

punamdahiya:
@flodolo Thanks for review, updated PR with suggested changes

locales/en-US/server.ftl
```diff
@@ -125,7 +128,16 @@ shotPageEditButton =
     .title = Edit this image
 shotPagefavoriteButton =
     .title = Favorite this shot
+shotPageBackToHomeButton =
+    .title = Homepage
+shotPageAllShotsButton =
+    .title = All Shots
+shotPageAllShots = All Shots
 shotPageDownload = Download
+# Note: Draw, Favorite and Delete text below is used on shot page header buttons as a verb (action)
```
flodolo: Please use one comment per string (not really needed for 'delete', since there's no space for confusion). Localizers don't see the file, see one string at a time, and the associated comment.

The alternative is to use section comments, but I'd avoid that in this case.

punamdahiya: @flodolo Good to know, Done. Thanks!

Issue #4748, opened 2018-08-06
Track channel and user agent in logins


ianb:
We want to distinguish pre-release users (i.e., Nightly and Beta users) from release Firefox users. Right now we're not saving enough in the device table to know the difference.

If we can get the channel information and add it to the login/register call that should be sufficient. Getting the User-Agent wouldn't be as helpful, as you have to match up dates to determine the channel based on the User-Agent. But we could consider User-Agent, as it could be done server-only.

We might want to consider copying the channel information to the shot (data) table on shot creation, so we track which channel the user was using at the moment they uploaded.

chenba:
Depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1386076

chenba:
If we can access prefs with an experiment, which is currently discussed in https://github.com/mozilla-services/screenshots/issues/4800, then we can grab the channel from the prefs as well.

ianb:
It might be easiest to just use release JSON files on the server: https://wiki.mozilla.org/Release_Management/Product_details

Specifically [this file](https://product-details.mozilla.org/1.0/firefox_history_major_releases.json) will identify when we should start treating a version as a production release, and [this file](https://product-details.mozilla.org/1.0/firefox_versions.json) gives some additional information. My only concern here is that if the file isn't updated very promptly (or ahead of a release) we might be unaware of a version update.

clouserw:
Getting channel from the add-on could be a stretch/enhancement.  Grabbing the product details seems easiest and most likely to get done this quarter.  We need to get the channel stuff hooked up to finish the Watchdog integration which we're on the hook for this quarter.

chenba:
`AppConstants.MOZ_UPDATE_CHANNEL` is available through an experiment API (https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/experiments/screenshots/api.js#L29).  Are not we not using that because we want to do everything server side?

6a68:
That API will be available as of Firefox 64, which hits release in December 2018.

PR #4836, opened 2018-09-04
Start #4748, add Firefox major version to the shot table

ianb:
Given the date and version number we can later deduce the Firefox channel

(I don't think we're splitting the database migration from the code currently, but I can't exactly remember?)

### DIFF
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -703,6 +703,10 @@ app.put("/data/:id/:domain",
       return;
     }
     const shot = new Shot(req.deviceId, req.backend, shotId, bodyObj);
+    const match = /Firefox\/(\d+)\./.exec(req.headers["user-agent"] || "");
+    if (match) {
+      shot.firefoxMajorVersion = parseInt(match[1], 10);
+    }
     let responseDelay = Promise.resolve();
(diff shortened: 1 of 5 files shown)


chenba:
We probably need a composite index on `firefox_major_version` and `created`.

ianb:
Sure, I'll add that. I'll also add a channel column while I'm at it, just so we get that database migration out of the way.

server/src/server.js
```diff
@@ -703,6 +703,10 @@ app.put("/data/:id/:domain",
       return;
     }
     const shot = new Shot(req.deviceId, req.backend, shotId, bodyObj);
+    const match = /Firefox\/(\d+)\./.exec(req.headers["user-agent"] || "");
```
chenba: nit, kind of?: I don't know if V8 automatically cache compiled regex literals in this context.  Since it's used on every upload, we should cached it.

ianb: According to the science of StackOverflow, [V8 will cache regexes like these](https://stackoverflow.com/questions/14352100/does-v8-cache-compiled-regular-expressions-automatically)

PR #4839, opened 2018-09-04
Remove favicon from UI and shot model

### LINKED ISSUE #4796: Remove source url favicon code from shot and shotindex page
Favicons against url from where shots are taken, are not functional and hidden on shot and shotindex page. 
With issue #4740 marked won't fix, we should remove related code from shot model and shotindex view and model and screenshots/server/src/proxy-url.js

chenba:
Fix #4796

### DIFF
--- a/shared/shot.js
+++ b/shared/shot.js
@@ -191,7 +191,6 @@ class AbstractShot {
     this.docTitle = attrs.docTitle || null;
     this.userTitle = attrs.userTitle || null;
     this.createdDate = attrs.createdDate || Date.now();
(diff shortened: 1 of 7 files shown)


shared/shot.js
```diff
@@ -458,15 +458,6 @@ class AbstractShot {
     this._createdDate = val;
   }
```
6a68: Should `this.favicon = attrs.favicon || null` (line 194) be removed also?

chenba: Oh yeah, good catch.

PR #4841, opened 2018-09-05
Exclude shots from Watchdog based on Firefox version/channel

### LINKED ISSUE #4770: [Watchdog] Add Release Channels Configuration
Add a configuration that would allow Screenshots to submit shots based on the shots' release channels.  This issue depends on #4748.

chenba:
Fix #4770

### DIFF
--- a/server/src/watchdog.js
+++ b/server/src/watchdog.js
@@ -13,6 +13,18 @@ const genUuid = require("nodify-uuid");
 const isEnabled = config.watchdog.enable;
 validateConfiguration();
 
+// The following is based on
+// https://product-details.mozilla.org/1.0/firefox_history_major_releases.json
+// and the release calendar. We will temporarily use it to determine if the shot
+// is from a browser that's in the release or esr channel, i.e. a major version
+// that is older than or equal to the current release version.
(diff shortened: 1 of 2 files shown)


server/src/watchdog.js
```diff
@@ -13,6 +13,18 @@ const genUuid = require("nodify-uuid");
...
+// The following is based on
+// https://product-details.mozilla.org/1.0/firefox_history_major_releases.json
+// and the release calendar. We will temporarily use it to determine if the shot
+// is from a browser that's in the release or esr channel, i.e. a major version
+// that is older than or equal to the current release version.
+const currentMajorVersion = 62;
+const firefoxMajorReleases = {
+  63: new Date("2018-10-23"),
+  64: new Date("2018-12-11"),
+  65: new Date("2019-01-29"),
```
chenba: @clouserw with the patch as it is, it'll work until Firefox 66 is released.  There's also a flag to control whether we check the channel/version at all.

clouserw: That sounds great.  If we get past Firefox 66 without figuring out what we're doing here let's make sure the failure mode is to not submit images.  Thanks!

PR #4864, opened 2018-09-11
WebExtension migration

6a68:
Having some trouble getting addons-linter to install, and had to manually delete and re-add package-lock.json (am I doing this wrong?), but this otherwise seems to work for me locally.

Things to note, from the actual commit:

-----

Migrate Screenshots to pure WebExtension 

* The addons system, not the Screenshots addon, will be responsible for
detecting changes to the 'extensions.screenshots.disabled' pref, and
starting up or shutting down the Screenshots webextension when changes
occur (bug 1488971).

* The LibraryButton code has been moved from bootstrap.js into the
embedded API. We'll be able to remove this code as soon as Library
menu items can be declared from the WebExtension manifest (bug 1366389).

* Move addon/webextension to webextension, update all the refs.

* Update makefile and build tools.

* Remove the script that deletes duplicate locales (en_CA) at build
time, because this breaks running with web-ext, and there is code to
handle duplicate locales inside export_mc.py already.

* Still TODO: README files and docs directory not yet updated


6a68:
Looks like Circle-CI had problems building addons-linter, and I had problems with that, too. I'm not sure if I broke something with the build tooling changes.

chenba:
The `addons-linter` issue is caused by some kind of (inter)dependency problem.  But https://github.com/mozilla/addons-linter/issues/1617 is not helping either.

If we reinstall `jpm` for this branch, then `addons-linter` would start working again.  With some digging, it seems like it's `sign-addon` from `jpm` that's needed.  But since `addons-linter` doesn't appear to use `sign-addon` directly, maybe it's one of `sign-addon`'s dependencies?  Madness.

6a68:
Nice sleuthing! I'm fine to re-add `jpm` as a dev dependency. Might even be able to get rid of those awful package-lock.json commits

chenba:
`Cannot find module 'regenerator-runtime'` is the actual error.  And it's a webpack dependency.

webextension/experiments/screenshots/schema.json
```diff
@@ -23,6 +23,20 @@
...
+      },
+      {
+        "name": "initLibraryButton",
+        "type": "function",
+        "description": "Initializes the Photon-based Library menu item for Screenshots",
+        "parameters": [],
+        "async": true
+      },
+      {
+        "name": "uninitLibraryButton",
```
aswan: this doesn't appear to actually be implemented?

PR #4865, opened 2018-09-11
Mark a shot as OpenGraph "image" type

notriddle:
When the URL is copy and pasted onto its own line in a [discourse.org](https://discourse.org) forum, the preview looks like this:

[](https://screenshots.firefox.com/p2rtWjzO6zbC7JNf/forum.bors.tech)

If the OpenGraph type is set to "image", it'll actually embed the image directly without any of the superflouous title stuff, the same way it does for other image hosts like imgur.

### DIFF
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -119,7 +119,7 @@ class Head extends React.Component {
       return null;
     }
     const og = [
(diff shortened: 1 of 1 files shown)


chenba:
@johngruen What do you think of this change?

johngruen:
Let's do it!

chenba:
@notriddle Thanks!

PR #4881, opened 2018-09-12
Fix #4650, ignore Renovate commits in generate-commit-log script

### LINKED ISSUE #4650: generate-commit-log script should ignore Renovate commits
The Renovate commit messages have consistent patterns, and we should ignore them when generating commit logs.

testeaxeax:
Fix #4650

### DIFF
--- a/bin/generate-commit-log
+++ b/bin/generate-commit-log
@@ -52,6 +52,8 @@ for line in log.splitlines():
 parts = []
 
 for message in messages:
+    if message.get("from") == "Renovate Bot <bot@renovateapp.com>":
+        continue
     sub = message.get("subject", "")
     if sub.startswith("[PATCH] "):
         sub = sub[len("[PATCH] "):]


bin/generate-commit-log
```diff
@@ -11,6 +11,8 @@ IGNORE_COMMITS = [
     "^chore",
     "^pontoon",
     "UTF.*Pontoon",
+    "^Update .* to .*",
```
chenba: This regex is too permissive.  But there's a better way.  When the script go through the log messages, it builds a dictionary for each.  So `message.get("from")` will give you the author of the commit, and you could match that against renovate.

testeaxeax:
Thanks for the help

bin/generate-commit-log
```diff
@@ -60,6 +60,8 @@ for message in messages:
         if re.search(pat, sub, re.I):
             should_skip = True
             break
+    if message.get("from") == "Renovate Bot <bot@renovateapp.com>":
+        should_skip = True
```
chenba: This works great.  But--and it's nit-ish--let's move this check up to be the first thing in the loop, and `continue` when true.  That way we'll just skip subject regex matching.

testeaxeax:
@chenba I updated the commit.

chenba:
@testeaxeax Thanks!

PR #4883, opened 2018-09-12
Fix #4882, use different secrets for different signing purposes

### LINKED ISSUE #4882: Scope Keygrip usage
Per [Bug 1389708](https://bugzilla.mozilla.org/show_bug.cgi?id=1389708), we shouldn't let you sign something for one purpose, then use the same thing to another purpose. To do this we'll use different Keygrip instances, and different secret keys.

This is instead of #4840

ianb:
This adds signing 'scopes', so that if you get something signed for one scope, you can't use it for another scope. E.g., you can't get a download URL and use that for an authentication key.

This keeps the 'legacy' scope, which is the current single key. This can be used now to make sure everything works when people upgrade, but removed later as people have used to the new specific scopes. But nothing new will be signed with the legacy scope once this is deployed.

This also updates some functions to use async/await.

### DIFF
--- a/server/src/dbschema.js
+++ b/server/src/dbschema.js
@@ -7,7 +7,9 @@ const mozlog = require("./logging").mozlog("dbschema");
 
 // When updating the database, please also run ./bin/dumpschema --record
 // This updates schema.sql with the latest full database schema
-const MAX_DB_LEVEL = exports.MAX_DB_LEVEL = 25;
+const MAX_DB_LEVEL = exports.MAX_DB_LEVEL = 26;
+// This is a list of all the Keygrip scopes we allow (and we make sure these exist):
+const KEYGRIP_SCOPES = ["auth", "proxy-url", "download-url", "ga-user-nonce"];
 
 exports.forceDbVersion = function(version) {
   mozlog.info("forcing-db-version", {db: db.constr, version});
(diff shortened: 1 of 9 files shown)


server/src/dbschema.js
```diff
@@ -93,57 +95,82 @@ exports.createTables = function() {
...
-    textKeys = fetchedTextKeys;
-    keys = new Keygrip(textKeys);
-  }).catch((err) => {
+exports.createKeygrip = async function() {
+  try {
+    const fetchedTextKeysByScope = await loadKeys();
+    textKeysByScope = fetchedTextKeysByScope;
+    keysByScope = {};
+    for (const scope in textKeysByScope) {
+      keysByScope[scope] = new Keygrip(textKeysByScope[scope]);
```
g-k: We should also bump the HMAC algorithm:

```js
new Keygrip(textKeysByScope[scope], 'sha256');
````

**Comment by g-k** [comment]
> The intention is for that to be allowed, with #4884 to remove those keys, since that's the migration strategy we have here.

Ah OK, I misunderstood the migration plan.

> That's also going to break migration, unless we actually instantiate two Keygrips and do our own fallback checking, and/or wrap Keygrip.

OK that can be a separate issue for later.

**Comment by ianb** [comment]
This should be ready, except for the errors in CircleCI. The server tests pass for me locally. I haven't been able to see what the error is; it's a 500 error in the server, but I'm not seeing the server logs in Artifacts.

File: server/src/proxy-url.js [hunk 2]
```diff
@@ -13,7 +13,8 @@ exports.createProxyUrl = function(req, url, hash) {
 };
 
 exports.createDownloadUrl = function(url, filename) {
-  const sig = dbschema.getKeygrip().sign(new Buffer(filename, "utf8"));
+  const path = (new URL(url)).pathname;
```
chenba: This needs `const { URL } = require("url");`

chenba:
Just a couple flake8 lint errors now.

ianb:
Green at last! Will squash and merge...

PR #4919, opened 2018-09-19
Catch Watchdog upstream error

### LINKED ISSUE #4871: [Watchdog] Handle error in Watchdog Response
After https://github.com/mozilla/watchdog-proxy/pull/185, Watchdog will include an `error` in its response to indicate whether PhotoDNA reported an error.  We need to use this property to prevent marking a shot as a false negative.

chenba:
Fixes #4871

### DIFF
--- a/server/src/watchdog.js
+++ b/server/src/watchdog.js
@@ -274,6 +274,15 @@ exports.handleResult = function(req) {
       throw err;
     }
 
+    // Check if there was an upstream error at PhotoDNA.
+    // We leave the result as null.
+    if (req.body.error) {
+      const err = new Error(`Watchdog submission encountered an error at PhotoDNA.`);
+      err.watchdogSubmissionId = String(record.id);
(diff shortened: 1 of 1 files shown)


chenba:
I'll file an issue in Watchdog to document its responses.

As for (intergration) testing, there's no simple way.  We could a) mock a Watchdog service to give us the responses we want to test, b) configure a Watchdog instance to give us the responses we want to test, or c) mock a PhotoDNA service to throw the errors we want.  😩

Could use mocks to unit test to see if the correct errors are thrown.

chenba:
Oh for manual verification, one could try uploading a shot that's >4MB.  But since the result is really a no-op, we'd need to check Sentry for the error.

6a68:
@chenba Cool. Do you think it'd be easy to mock out Watchdog for local testing?

6a68:
Doh, missed your earlier comment :-P

chenba:
Filed #4931 for writing tests for Watchdog integration.

PR #4928, opened 2018-09-20
Fix #4748, add channel column to shot uploads

### LINKED ISSUE #4748: Track channel and user agent in logins

ianb:


webextension/background/takeshot.js
```diff
@@ -55,7 +55,12 @@ this.takeshot = (function() {
       } else {
         shot.thumbnail = thumbnailImage;
       }
-    }).then(() => {
+      if (browser.experiments && browser.experiments.screenshots && browser.experiments.screenshots.getUpdateChannel) {
+        return browser.experiments.screenshots.getUpdateChannel();
+      }
+      return Promise.resolve(null);
```
6a68: You don't need the object property guard stuff or the null return option. The API has already landed in master, and it's already used [elsewhere in the addon](https://github.com/mozilla-services/screenshots/blob/981d57403b1cc5c7f54fb6c890999a6cdd219e9d/webextension/background/selectorLoader.js#L95). You can assume it exists, call it, and return the value

ianb: I've occasionally seen problems loading the experiment in Side View, so I wanted to be careful not to break anyone's experience if that happened. But... this is fine too.

PR #4932, opened 2018-09-20
Add logs only mode for Watchdog integration

### LINKED ISSUE #4926: [watchdog] Disable any actions from watchdog
For our initial test, I'd like to make watchdog logging only, meaning, don't flag an image as positive even if watchdog says it is.  That way we can review what it's flagging and fine tune without affecting users.

For this issue, just no-op the response from Watchdog.  That will give us time to figure out the legal flows.

/cc @chenba

chenba:
Fixes #4926


chenba:
Comment for followup: https://github.com/mozilla-services/screenshots/pull/4932/files#diff-480d08f875b12769794c80d3c6cccb9aR421

Issue #4323, opened 2018-04-10
Upgrade fluent and fluent-react


chenba:
This is a follow-up to #4280.  Upgrading these modules cause runtime errors.  Code changes are most likely necessary.

wresuolc:
Moving into stretch until the fix above is done

chenba:
Fluent 0.8.0 has this workaround: https://github.com/projectfluent/fluent.js/wiki/React-Overlays#custom-markup-parsers

chenba:
Fixed in #4935

PR #4935, opened 2018-09-21
Upgrade to fluent 0.8.0, fluent-react 0.8.1

stasm:


chenba:
Works great. 🍻  Fixes #4323.

server/src/pages/shot/shotpage-header.js
```diff
@@ -31,7 +31,7 @@ exports.ShotPageHeader = class ShotPageHeader extends React.Component {
       );
     }
     return (
-      <Localized id="shotPageAllShotsButton">
+      <Localized id="shotPageAllShotsButton" attrs={{title: true}}>
         <a className="nav-button icon-shots" href="/shots" tabIndex="0" title="All Shots">
           <Localized id="shotPageAllShots">
             <span>All Shots</span>
```
6a68: Missed one spot here: see `expirationSubtitle` in the `renderShotInfo()` method at line 52 of this file. Currently, instead of rendering the shot expiration component, I just see the string 'timediff':

image

Should look something like this:

image

stasm: Thanks for catching it. I was able to run the addon locally which allowed me to fix this.

stasm:
Thanks for the reviews, @flodolo and @6a68! I updated the PR and rebased it on top of the current master.

In order to fix the `timediff` issue I had to make changes to a few strings. The reason why it didn't display properly was that the `$timediff` variable was an instance of a React element. Passing React elements as arguments to translations used to work in `fluent-react` 0.4. Starting with `fluent-react` 0.6.0 this feature [has been removed](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/CHANGELOG.md#dont-pass-elements-as-arguments) in favor of so-called [overlays](https://github.com/projectfluent/fluent.js/wiki/React-Overlays). In short, `<Localized>` can take other React elements as props. If an HTML element of the same name is then found in the translation, the element passed as prop is inserted into the translation (keeping its props) and its children are set to the text content of the element found in the translation.

In the updated version of the PR, I take advantage of this new feature to fix the `timediff` issue as well as to improve the localization of the `<p>` found in `AdBanner`. This follows a good practice of leaving all concatenation and interpolation to the localizers.

```properties
# Fluent code
bannerUpsell = {gScreenshotsDescription} <a>Get Firefox now</a>
```

```js
// React code
const upsellLink = <a className="upsellLink" href="…" onClick={…}></a>;
bannerContent = <Localized id="bannerUpsell" a={upsellLink}><p></p></Localized>;
```

Ideally, we'd use the same pattern in the expiration messages and inline the `timeDiff*Ago` and `timeDiffFuture*` strings into them. [Some redundancy is good for l10n](https://github.com/projectfluent/fluent/wiki/Good-Practices-for-Developers#prefer-redundancy-over-being-smart). However, this approach would require the expiration strings to have nested branches: one level for units (minutes, hours, days) and then another one for the plural category based on the number. While this is well supported by Fluent, Pontoon doesn't have a good UI for such strings yet. I consulted with @flodolo and we agreed that it would be best to keep the current approach for now. That is, continue to translate `<TimeDiff>` components independently and then pass them into other strings as props for interpolation.

stasm:
@flodolo LGTM, thanks. I pushed a tiny update to the PR to include the closing `</timediff>` tag in FTL comments. This makes the en-US files consistent with the output of your script.

6a68:
When rebasing, note that #4947 just added a new string

flodolo:
@stasm 
Do you have a chance to rebase this today? I'll be on PTO Thursday to Sunday, and I'd like to get this out today, given the amount of moving pieces (and pending PRs with strings).

stasm:
@flodolo I'm on it. I'd like to land this today as well.

stasm:
I rebased the PR. It's now ready to merge.

stasm:
Note the developers: Now that Screenshots uses `fluent-react` 0.8.1, please remember to [use the `attrs` prop](https://github.com/projectfluent/fluent.js/wiki/Localized) in `Localized` components which need to translate a prop, like `title`, `alt` and `placeholder`.

chenba:
Thanks, guys!

Issue #4804, opened 2018-08-22
Upgrade to Node.js 10.x


chenba:
Node.js 10 will enter into LTS in October 2018.  We should consider upgrading then.  It would also give us NPM 6.x.

chenba:
[This line](https://github.com/mozilla-services/screenshots/pull/4948/files#diff-1d37e48f9ceff6d8030570cd36286a61R66) can be removed once the CircleCI config is updated to use Node 10.

ianb:
This issue is being closed (without being fixed) because [we are discontinuing the Screenshots server](https://blog.mozilla.org/futurereleases/2019/01/24/clarifying-the-future-of-firefox-screenshots/).

PR #4948, opened 2018-09-24
Fix npm audit fail issue 4803

### LINKED ISSUE #4803: Make npm audit fail the tests
In #4617 (specifically a5935e1ea126007c22c68fb9bf7908f62111c8ab) we added npm audit and removed nsp (per #4704). But npm audit is run like `npm audit || true`, as we have failures we can't reasonably fix. We should reenable it as something that can make the CI tests fail.

I personally would propose we use `npm audit --json | ./bin/some_ignoring_script` where write `some_ignoring_script` to parse the JSON and ignore some packages we don't care to fix. E.g., `jpm` has a dependency (`open`) with a "critical error" that I feel confident we can ignore (because it is only used in our build system, never with untrusted input).

g-k:
Assuming no new insecure deps, this should fix #4803 (for CI at least).


6a68:
> Sorry about all the spam in #screenshots on irc.

I'll take free help + irc spam any time you have spare cycles ^_^

chenba:
The npm version required is a problem locally until #4804 is fixed.  (Or we could document that as a requirement for the project).

g-k:
OK made some changes:

* added mac support in https://github.com/mozilla-services/audit-filter/issues/3
* stripped the binaries; makes them almost 10x smaller
* moved the shell script into git (in `example/bin`) https://github.com/mozilla-services/audit-filter/#install and pulled in the rustup arch detection

Haven't tested this PR on OSX (have tested audit-filer), but try running: `CI=1 npm run-script lint:deps`

PR #4963, opened 2018-09-26
Change 'save' button to 'upload' button

6a68:
I think we were wrong to use 'Save' to mean 'Upload', and I don't recall why we went with it in the first place. It could seem like we're desperate to trick people into uploading shots. Let's not give that impression.

screen shot 2018-09-26 at 1 10 48 pm

### DIFF
--- a/locales/en-US/webextension.properties
+++ b/locales/en-US/webextension.properties
@@ -6,6 +6,7 @@ contextMenuLabel = Take a Screenshot
 myShotsLink = My Shots
 screenshotInstructions = Drag or click on the page to select a region. Press ESC to cancel.
 saveScreenshotSelectedArea = Save
+uploadScreenshotSelectedArea = Upload
 saveScreenshotVisibleArea = Save visible
 saveScreenshotFullPage = Save full page
 cancelScreenshot = Cancel
--- a/webextension/selector/ui.js
+++ b/webextension/selector/ui.js
@@ -420,9 +420,9 @@ this.ui = (function() { // eslint-disable-line no-unused-vars
                           data-l10n-title="downloadScreenshot"><img
                           src="${browser.extension.getURL("icons/download.svg")}" /></button>
                          <button class="preview-button-save"
-                          data-l10n-title="saveScreenshotSelectedArea"><img
+                          data-l10n-title="uploadScreenshotSelectedArea"><img
                           src="${browser.extension.getURL("icons/cloud.svg")}"
-                          />${browser.i18n.getMessage("saveScreenshotSelectedArea")}</button>`
+                          />${browser.i18n.getMessage("uploadScreenshotSelectedArea")}</button>`
                       }
                     </div>
                   </div>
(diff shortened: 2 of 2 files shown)


johngruen:
WFM

6a68:
> Why don't we ditch the string altogether and just use a title like on the other buttons?

SGTM :+1:

johngruen:
@6a68 i made a PR into your branch to remove the string from the upload button: 

https://github.com/6a68/screenshots/pull/8

evilpie:
I am not a big fan of removing the text, have we thought about going to opposite way, i.e adding "Save" etc. ? Even the titles on hover could be more descriptive, like "Save file to disk", "Upload to screenshots.firefox.com"

johngruen:
@evilpie like the idea of having longer tooltips. I'm less enthusiastic about adding more text to buttons b/c localization will make the widths weird. I'll check in with our content strategy person and update strings.

6a68:
@johngruen Hmm. I fear that removing the text will take us from 'misleading button label' to 'inscrutable button label':

screen shot 2018-10-10 at 1 28 31 pm

johngruen:
@6a68 not an hill i want to die on. We can keep the Upload text for a release and look at metrics to determine effects... R+, but can you change two colors in the sass variables file.

```
$link-blue: #0a84ff;
$active-blue: #45a1ff;
```

schuetzm:
Please also change the download button to "save", with an appropriate icon. Nothing is being downloaded here, it's a purely local operation (or should be, at least).

6a68:
@johngruen Hmm. We could A/B test this label change, since it will change the core flow of the product. Might actually be a good exercise to get us prepared for the growth focus in Q1 of next year. How's that grab you?

6a68:
@schuetzm Thanks for the feedback, but I think it's actually much worse to change a button label to a different button, than it is to change the label to a new word. If you are used to the Screenshots convention that 'save' means 'upload', you are going to be unpleasantly surprised to discover 'save' suddenly means 'download'.

Deledrius:
> If you are used to the Screenshots convention that 'save' means 'upload', you are going to be unpleasantly surprised to discover 'save' suddenly means 'download'.

That's a reasonable concern, and I generally agree, but not fixing the mistake because the mistake was made is a circular justification.  Both of these buttons are misleading (and factually incorrect).  Leaving one wrong because it's been wrong just prolongs the problem.  Will it be acceptable to fix this later, after it's been mislabeled even longer?  If anything, user confusion is precisely the reason this should be rectified as soon as possible (in fact, it would have avoided this new concern altogether had it been fixed when the privacy issue was first reported *over a year ago*).

More directly: leaving the save button as "Download" generates confusion over the initial disposition of the image.  If I can both Upload and Download the image, where is it?  Is it being produced on Mozilla's server?  Is it produced locally?  A user can't be sure, because the language implies both and neither simultaneously.

There is a standard convention for this sort of interface: local data may be **saved** from memory onto disk, or they may be **uploaded** to a remote location.  The improved UI, even with this commit which addresses the privacy violation implications resulting from the mislabeled button, still doesn't adhere to this convention.

I suppose that's a problem for a different PR, however.  It's just a bit frustrating to see part of this poor UX persisting.

muxator:
Just for info: I am on Firefox 63 (Ubuntu 18.10), but this PR did not make it.
Maybe in 6 weeks for Firefox 64 we'll have it!
screenshot at 2018-10-21 16-32-38

muxator:
Quick update: at least the official Windows x64 **Firefox 66** package **seems to miss this update**.
The _Upload_ button still reads _Save_.

Cattura

PR #4964, opened 2018-09-26
Fix copy shot button race condition

### LINKED ISSUE #4957: The "Copy" button disappears/appears after the Shots page is refreshed
[Notes]:
- In certain cases the "Copy" button is not displayed on the Shots page after a shot is displayed. The button appears after the page is refreshed multiple times.
- If you edit the save shot, the "Copy" button disappears and is not displayed if you normally refresh the page. You need to refresh the page using the "Ctrl+F5/Cmd+F5" shortkey for the button to reappears.
- The "Copy" button does not disappear if the page is refresh using the "Ctrl+F5/Cmd+F5" shortkey.

[Affected versions]:
- Nightly 64.0a1
- Screenshots 34.0.0 dev

[Affected Platforms]:
- All Windows
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" boolean pref is set to "false".
- The "extensions.legacy.enabled" boolean pref is set tot "true".
- Have the Screenshots 34.0.0 dev version installed from [here](https://screenshots.dev.mozaws.net/homepage/install-test-local.html), on a new clean profile. 

[Steps to reproduce]:
1. Open the Firefox browser with the profile from prerequisites.
2. Navigate to any website (eg:https://www.wikipedia.org/) and save a selection on the page.
3. Observe the "Copy" button.
4. Refresh the page and observe the "Copy" button.

[Expected result]:
Step 3 and Step 4: The "Copy" button is correctly displayed,

[Actual results]:
Step 3: The "Copy" button is correctly displayed.
Step 4. The "copy" button is no longer displayed. 

[Additional Notes]:
- Here is a screen recording of the issue:
copy button not displayedd

chenba:
Fixes #4957

### DIFF
--- a/server/src/pages/shot/controller.js
+++ b/server/src/pages/shot/controller.js
@@ -78,7 +78,6 @@ exports.launch = function(data) {
   }
   model.highlightEditButton = shouldHighlightEditIcon(model);
   model.promoDialog = shouldShowPromo(model);
-  document.dispatchEvent(new CustomEvent("request-addon-present"));
 
   if (firstSet) {
     refreshHash();
@@ -106,6 +105,8 @@ exports.launch = function(data) {
     } else {
       sendEvent("visit", "non-owner");
     }
+
+  document.dispatchEvent(new CustomEvent("request-addon-present"));
 };
 
 exports.changeShotExpiration = function(shot, expiration) {
--- a/server/src/pages/shot/view.js
+++ b/server/src/pages/shot/view.js
@@ -198,8 +198,15 @@ class Body extends React.Component {
     this.setState({
       highlightEditButton: this.props.highlightEditButton || this.props.promoDialog,
       promoDialog: this.props.promoDialog,
-      canCopy: !!this.props.canCopy,
+      isClient: true,
     });
+  }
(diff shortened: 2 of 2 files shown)


server/src/pages/shot/controller.js
```diff
@@ -280,15 +282,28 @@ exports.setTitle = function(title) {
...
+      // rendered content, it's performed after the initial rendering.
+      const delayedRender = () => {
+        if (initialRender) {
+          model.canCopy = true;
+          render();
+          return;
+        }
+        setTimeout(delayedRender, 50);
+      };
+      delayedRender();
```
6a68: Hmm. It seems funny to use a setTimeout here, but I'm not sure of the best workaround.

What if we set `suppressHydrationWarning`, to signal that React should ignore the gap between server + client, or maybe set `state.isClient = true` inside `componentDidMount`, and checked the `canCopy` value from the model if `isClient`?

chenba: `isClient` works great.  Thanks!  Also moved the event dispatch to a more sensible place.

Issue #4777, opened 2018-08-17
Clean Up Old States from the States Table Periodically


chenba:
A state is created whenever someone initiates an FxA auth request.  Then it's deleted when the auth is successful.  The rest of the time, we've got states lingering about.  We should delete those old records.  We can be conservative: delete states that are 1+ week old?

punamdahiya:
Either we or ops should setup a cron job to cleanup the states

chenba:
We can add this to `jobs.js`.

chenba:
This requires a migration to add a `created` timestamp column.

ianb:
This issue is being closed (without being fixed) because [we are discontinuing the Screenshots server](https://blog.mozilla.org/futurereleases/2019/01/24/clarifying-the-future-of-firefox-screenshots/).

PR #4966, opened 2018-09-27
Fixes #4959 - Login flow should work from a non-Firefox browser (with no extension)

### LINKED ISSUE #4959: [FxA] Login flow should work from a non-Firefox browser
Something we should confirm and make sure works:

- Using a browser besides Firefox desktop
- Go to https://screenshots.firefox.com (or localhost:10080)
- See a "sign in" button
- That starts an FxA sign-in flow
- After finishing, you should be able to see My Shots with any shots associated with your FxA account
- You should be able to delete and favorite shots

Note that we don't need to support creating new shots.

We *do not* need any persistent association between the browser, any deviceId, and the FxA id (accountId). Ideally we'd simply set the authentication cookie with an accountId, a null deviceId, and any appropriate queries (query on My Shots, permission check on delete/favorite) would work with a null deviceId and an existent req.accountId.

ianb:
**do not merge**

Needs better abstractions and a close look at all the references to deviceId.
Needs to be tested across all the use cases we care about.
Should fix the confirm-login cookie issue, so both cases (browser and extension) log in similarly


server/src/server.js
```diff
@@ -1111,33 +1118,52 @@ app.get("/api/fxa-oauth/confirm-login", function(req, res, next) {
...
+              }).send();
+            }
+            // Redirect to endpoint with auth param indicating successful Fxa auth flow.
+            // 'auth' param is used in reactrender and reactruntime to load wantsauth.js
+            // and display Fxa SignIn button state when request doesn't have accountId
+            // right after fxa-ouath/confirm-login redirection.
+            const pageUri = endpoint ? "/" + endpoint : "/";
+            // FIXME:
+            // We should be setting the accountid cookie in both paths, like this:
+            const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("auth")});
```
ianb: Arguably we should use sendAuthInfo in both cases

server/src/server.js
```diff
@@ -1095,10 +1102,10 @@ app.get("/api/fxa-oauth/login/*", function(req, res, next) {
 });
 
 app.get("/api/fxa-oauth/confirm-login", function(req, res, next) {
```
ianb: We should convert this to an async/await function

server/src/reactruntime.js
```diff
@@ -37,6 +37,7 @@ exports.HeadTemplate = class HeadTemplate extends React.Component {
       localeScripts.push(<script key={`l10n-${locale}`} src={this.props.staticLink(`/static/locales/${locale}.js`)} />);
     }
 
+    console.log("include wantsauth?", {authenticatied: this.props.authenticated, hasFxa: this.props.hasFxa, authFxa: this.props.authFxa});
     const wantsAuth = (!this.props.authenticated) ||
```
ianb: I don't think we need authFxa here, if we fix the `/confirm-login` endpoint

server/src/pages/shot/model.js
```diff
@@ -27,11 +27,11 @@ exports.createModel = function(req) {
     id: req.shot.id,
     productName: req.config.productName,
     isExtInstalled: !!req.deviceId,
-    isOwner: req.deviceId === req.shot.ownerId || isFxaAuthenticated,
+    isOwner: req.shot.isOwner(req.deviceId, req.accountId),
     isFxaAuthenticated,
     gaId: req.config.gaId,
     deviceId: req.deviceId,
-    authenticated: !!req.deviceId,
+    authenticated: !!(req.deviceId || req.accountId),
```
ianb: I'm unclear why this is set here, and also in reactrender. Probably for no good reason ;)

server/src/pages/homepage/view.js
```diff
@@ -79,6 +79,8 @@ class Body extends React.Component {
       <reactruntime.BodyTemplate {...this.props}>
         <HomePageHeader isOwner={this.props.showMyShots} isFirefox={this.props.isFirefox}
           hasFxa={this.props.hasFxa} />
+        {this.props.showMyShots ? null : <a href="/api/fxa-oauth/login/">Login</a>}
```
ianb: Obviously this is the hackiest of all links, and probably the logic (`this.props.showMyShots`) is wrong too, and is *definitely* named incorrectly.

server/src/pages/shot/server.js
```diff
@@ -12,7 +12,7 @@ app.get("/:id/:domain", function(req, res) {
   const shotId = `${req.params.id}/${req.params.domain}`;
   Shot.get(req.backend, shotId).then((shot) => {
     const noSuchShot = !shot;
-    const nonOwnerAndBlocked = shot && shot.blockType !== "none" && req.deviceId !== shot.ownerId;
+    const nonOwnerAndBlocked = shot && shot.blockType !== "none" && !shot.isOwner(req.deviceId, req.accountId);
```
ianb: We should include something about this in the QA plan (blocked shots and FxA login)

server/src/reactruntime.js
punamdahiya: Link to PR where we added wantsauth check to conditionally include wantsauth.js on all pages https://github.com/mozilla-services/screenshots/pull/4759/files#diff-79019fb192622531c0f432b6833b083dR50

Here we are relying on auth flag from confirm-login redirect to load wantsauth.js and make explict calls to getAuthData that sets accountId cookie and update accountId request parameter used to fill hasFxA model property used in all pages to determine if a user has firefox account

punamdahiya:
Getting below error while retrieving images on myshots page
screenshots-server.server.ERROR: generic-error-response {"msg":"Error getting image from database:\nerror: invalid reference to FROM-clause entry for table \"images\"\n\nerror: invalid reference to FROM-clause entry for table \"images\"\n

server/src/servershot.js
```diff
@@ -326,12 +330,13 @@ class Shot extends AbstractShot {
...
   return db.select(
-    `SELECT images.url, images.contenttype, data.deviceid
-     FROM images
+    `SELECT images.url, images.contenttype, data.deviceid, devices.accountid
+     FROM images, devices
       JOIN data ON images.shotid = data.id
      WHERE images.id = $1
       AND (data.expire_time IS NULL OR data.expire_time > NOW())
       AND data.block_type = 'none'
+      AND data.deviceid = devices.id
```
punamdahiya: Query should be fixed to include JOIN on devices something like 
SELECT images.url, images.contenttype, data.deviceid, devices.accountid FROM images 
JOIN data ON images.shotid = data.id
JOIN devices ON data.deviceid = devices.id
WHERE images.id = $1
AND (data.expire_time IS NULL OR data.expire_time > NOW()) 
AND data.block_type = 'none' AND NOT data.deleted

chenba: Since the condition `data.deviceid = devices.id` is in the WHERE clause, you'll get the same result.  If you swap the placement of `images` and `devices`, that fixes the error.

chenba: Sorry, my previous comment not specific enough.  Swap `images` and `devices` on line 334 to fix the error.

punamdahiya:
Saving images after edit flow is broken. On Mobile Edit tools crop, pen and highlight are not functional and needs to be updated to support mobile

chenba:
We should handle mobile UI/UX in another bug.  Let's keep the scope to FxA auth without the addon and device id.

server/src/server.js
punamdahiya: sendAuthInfo checks for deviceId before settings cookies, if we are going to use that flow we need to update that check to include accountId

punamdahiya:
I will resolve conflict and update branch

server/src/servershot.js
```diff
@@ -165,7 +165,7 @@ class Shot extends AbstractShot {
     this.ownerId = ownerId;
   }
 
-  isOwner(deviceId, accountId) {
+  checkOwner(deviceId, accountId) {
```
punamdahiya: This is redundant and can be removed

server/src/server.js
```diff
@@ -1070,13 +1074,20 @@ const END_POINT_SEPARATOR = "|";
...
     const state = stateBytes.toString("hex");
-    return setState(req.deviceId, state).then(inserted => {
+    // FIXME: should be a different name than "auth":
+    const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("auth")});
+    const stateId = req.deviceId || String(Math.random());
+    if (!req.deviceId) {
+      cookies.set("fxaState", stateId, {signed: true});
+    }
+    console.log("Using oauth state", stateId);
+    return setState(stateId, state).then(inserted => {
```
punamdahiya: Version 27 db support: Logic here assumes a random string as deviceId. If the authentication fails and the data is not deleted in checkstate inside confirm-login route handler, reverting db to 27 will fail to apply stateid_deviceid fk constraint

punamdahiya: @ianb thoughts on how to fix this issue of fail to revert to db 27 and earlier because of numeric state value that violate fk constraint

punamdahiya: #4777 could be one fix but in the fix proposed we are allowing a week window

server/src/servershot.js
ianb: I agree, we shouldn't have different kinds of checks or checking functions around, we should have One Right Way

punamdahiya:
Tests were failing earlier with internal server error because of /api/register call failing due to isOwner not been set. https://pastebin.com/AhvknQaP

With #5001 fix, tests are moving ahead and failing as expected with lint errors

punamdahiya:
Updated with latest master branch

punamdahiya:
Fixes #4959

server/src/pages/homepage/homepage-header.js
```diff
@@ -16,9 +16,8 @@ exports.HomePageHeader = class HomePageHeader extends React.Component {
 
   renderFxASignIn() {
       return (
-        this.props.isFirefox && this.props.isOwner ?
-          <SignInButton isAuthenticated={this.props.hasFxa} initialPage=""
-                        staticLink={this.props.staticLink} /> : null
+          <SignInButton isFxaAuthenticated={this.props.hasFxa} initialPage=""
```
punamdahiya: Renaming isAuthenticated SignInButton property to isFxAuthenticated to indicate its specific to Firefox Account and indicates accountId cookie exists after successful Fxa login

server/src/pages/homepage/view.js
```diff
@@ -167,7 +167,7 @@ Body.propTypes = {
   hasFxa: PropTypes.bool,
   firefoxVersion: PropTypes.string,
   isFirefox: PropTypes.bool,
-  showMyShots: PropTypes.bool,
+  authenticated: PropTypes.bool,
```
punamdahiya: Removing showMyShots that was previously checking for deviceId in favor of authenticated set inside ReactRender.js when a deviceId or accountId auth is found

punamdahiya:
@ianb @6a68 Updated PR to
a) Hide SignIn from shot page for !owned shots in non-Firefox browsers
b) Hide Upsell banner in all browsers when authenticated
c) Remove all data from footer work in all browsers once authenticated

With above changes screenshots flow should work in all browsers as shown in design doc. Thanks

test/server/clientlib.py
```diff
@@ -38,6 +38,12 @@ def login(self):
         resp.raise_for_status()
         return resp
 
+    def fxa_login(self, account_id):
+        device_id = get_device_id(account_id)
+        if device_id:
+          account_info = get_account_cookie_detail(account_id)
+          self.session.cookies.update(account_info)
```
ianb: A couple issues here:

- `fxa_login()` doesn't make it clear what's being logged in. Are you logging the session in? Or is the device being associated with the accountId, as happens when you sign in with the add-on installed?
- There should a method to log the device into an account. (This method would have to call out to pglib, then call `self.login()`)
- For separate reasons we want a session to be able to get an accountId, without getting a deviceId. I don't think this method should do that.
- Testing device_id, but not testing that it's the same as `self.device_id` isn't good.
- You have two spaces instead of four ;)

test/server/pglib.py
```diff
@@ -21,7 +21,7 @@
 # into a cookie jar.
 def attach_device(device_id, account_id):
     dbname = pg_vars["PGDATABASE"] or pg_vars["PGUSER"]
-    conn = psycopg2.connect(dbname=dbname, user=pg_vars["PGUSER"], host=pg_vars["PGHOST"], port=pg_vars["PGPORT"])
+    conn = psycopg2.connect(dbname=dbname, user=pg_vars["PGUSER"], host=pg_vars["PGHOST"], port=pg_vars["PGPORT"], password="postgres")
```
ianb: I'm guessing this is a local configuration issue, but it shouldn't be necessary. It might break tests for other installations, though I'm not sure.

test/server/pglib.py
```diff
@@ -34,6 +34,29 @@ def attach_device(device_id, account_id):
...
+def get_device_id(account_id):
+    dbname = pg_vars["PGDATABASE"] or pg_vars["PGUSER"]
+    conn = psycopg2.connect(dbname=dbname, user=pg_vars["PGUSER"], host=pg_vars["PGHOST"], port=pg_vars["PGPORT"], password="postgres")
+    conn.autocommit = True
+    cur = conn.cursor()
+    cur.execute("SELECT id FROM devices WHERE accountid = '" + account_id + "'")
+    key_row = cur.fetchone()
+    cur.close()
+    conn.close()
+    return {"deviceid":  key_row[0]}
```
ianb: Should be more like `return [row[0] for row in cur.fetchall()]`

test/server/test_fxa_auth.py
```diff
@@ -0,0 +1,78 @@
...
+    first_user = ScreenshotsClient()
+    second_user = ScreenshotsClient()
+
+    first_user.login();
+    shot_url = first_user.create_shot(docTitle="A_TEST_SITE_1")
+    shot_id = urlparse.urlsplit(shot_url).path.strip("/")
+    shot_page = first_user.read_shot(shot_url)
+    print(first_user.read_shot(shot_url)["clip_url"], shot_page["clip_url"])
+    assert first_user.read_shot(shot_url)["clip_content"] == shot_page["clip_content"]
+    assert "A_TEST_SITE_1" in shot_page["page"]
```
ianb: You don't really need to test that the shot is uploaded correctly, we have other tests for that.

test/server/test_fxa_auth.py
```diff
@@ -0,0 +1,78 @@
...
+def get_url(url, expect=None):
+    result = urllib.urlopen(url)
+    if expect and result.getcode() != expect:
+        raise Exception("GET {} returned status {}".format(url, result.getcode()))
+    return result.read()
+
+# test we can change expiration and delete shots made previously
+def test_owned_shot_with_fxa_auth():
+    first_user = ScreenshotsClient()
+    second_user = ScreenshotsClient()
```
ianb: You might name `second_user` as `browser_only_user` or something to make it clear.

test/server/test_fxa_auth.py
```diff
@@ -0,0 +1,78 @@
...
+    shot_url = first_user.create_shot(docTitle="A_TEST_SITE_1")
+    shot_id = urlparse.urlsplit(shot_url).path.strip("/")
+    shot_page = first_user.read_shot(shot_url)
+    print(first_user.read_shot(shot_url)["clip_url"], shot_page["clip_url"])
+    assert first_user.read_shot(shot_url)["clip_content"] == shot_page["clip_content"]
+    assert "A_TEST_SITE_1" in shot_page["page"]
+
+
+    # Use first_user accountId to login to a second session
+    second_user.fxa_login(first_user.accountId)
```
ianb: I guess because this never calls `second_user.login()`, therefore this gets an accountId cookie but no deviceId, but it's not obvious. Maybe it would be fine if after instantiation we just set `second_user.device_id = None` to be double-sure it's not half-logged-in.

From this perspective, maybe it should be something like `second_user.browser_only_login(accountId)`, and that method could ensure that the cookies were cleared, deviceId was unset, and the appropriate accountId cookie was added to the session.

test/server/test_fxa_auth.py
```diff
@@ -0,0 +1,78 @@
...
+    # Read !owned shot created in first_user session
+    page = second_user.read_shot(shot_url)
+
+    # set expiration on the shot should throw 404
+    try:
+        try:
+            second_user.set_expiration(shot_url, 1)
+        except HTTPError, e:
+            if e.response.status_code != 404:
+                raise
```
ianb: This won't raise an error if it's a 200 response, only if it's >=300 and not 404.

You could just add `assert False, "set_expiration should fail"` after the second_user.set_expiration() call, or you could use the pytest pattern here: https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

test/server/clientlib.py
punamdahiya: > A couple issues here:
> 
>     * `fxa_login()` doesn't make it clear what's being logged in. Are you logging the session in? Or is the device being associated with the accountId, as happens when you sign in with the add-on installed?

Updated to move this method in ScreenshotsBase class to simulate browser only session with user logged in with accountId.

>     * There should a method to log the device into an account. (This method would have to call out to pglib, then call `self.login()`)
> 
My understanding ScreenshotsClient class login method logs the device into an account by calling attach_device which can be used to setup an account. 

>     * For separate reasons we want a session to be able to get an accountId, without getting a deviceId. I don't think this method should do that.
> 
Using ScreenshotsBase class from clientlib the above scenario should suffice.

>     * Testing device_id, but not testing that it's the same as `self.device_id` isn't good.
> 
Updated tests with check
>     * You have two spaces instead of four ;)
done

test/server/pglib.py
punamdahiya: that's correct, updated PR to remove explicitly adding password

test/server/test_fxa_auth.py
punamdahiya: Removed shot upload checks related assert

test/server/test_fxa_auth.py
punamdahiya: Done

test/server/test_fxa_auth.py
punamdahiya: With browser only user created as instance of ScreenshotsBase class we don't have deviceId or accountId assigned at initialization.  Should we check for any previous existing accountId cookie?

test/server/clientlib.py
```diff
@@ -7,58 +7,34 @@
...
-            shot_data,
-            json=shot_json,
-        )
-        resp.raise_for_status()
-        return shot_url
+    def _get_id_from_url(self, url):
+        assert url.startswith(self.backend)
+        id = url[len(self.backend):]
+        id = id.strip('/')
+        return id
```
punamdahiya: _get_id_from_url, read_shot, delete_shot, set_expiration are moved from ScreenshotsClient to base class ScreenshotsBase without any change

punamdahiya:
@ianb PR is updated with automated tests listed in #5031. Please review. Thanks!

punamdahiya:
@ianb With this change https://github.com/mozilla-services/screenshots/pull/4726/files#diff-e79f0cf7ae229b6047499adeeea99ae6R36, we have our tests in master running with both deviceId and accountId. 
#4726 added accountId in clientlib  login flow to make set expiration work for few tests in test_csrf and test_file_management. Thinking more we should have updated tests in csrf and file management to login with accountId to test set expiration flow. Should we create an issue to address this? Thoughts?

punamdahiya:
Fixes #4959 , #5031

punamdahiya:
> @ianb With this change https://github.com/mozilla-services/screenshots/pull/4726/files#diff-e79f0cf7ae229b6047499adeeea99ae6R36, we have our tests in master running with both deviceId and accountId.
> #4726 added accountId in clientlib login flow to make set expiration work for few tests in test_csrf and test_file_management. Thinking more we should have updated tests in csrf and file management to login with accountId to test set expiration flow. Should we create an issue to address this? Thoughts?

Created https://github.com/mozilla-services/screenshots/issues/5050 to track above issue

server/src/pages/homepage/view.js
```diff
@@ -105,8 +105,7 @@ class Body extends React.Component {
     return (
       <reactruntime.BodyTemplate {...this.props}>
         <HomePageHeader
-          isOwner={this.props.showMyShots}
-          isFirefox={this.props.isFirefox}
+          isOwner={this.props.authenticated}
```
ianb: Is `isOwner` a misnomer now? (Or maybe it always was)

server/src/pages/shot/shotpage-header.js
```diff
@@ -88,13 +88,15 @@ exports.ShotPageHeader = class ShotPageHeader extends React.Component {
   }
 
   renderFxASignIn() {
-    return (
-      this.props.isOwner ?
+    if (this.props.isOwner) {
```
ianb: Are we choosing not to show the login button to Firefox users that aren't logged in, and are viewing someone else's page?

server/src/pages/shotindex/view.js
```diff
@@ -62,12 +62,12 @@ class Body extends React.Component {
     const children = [];
     if (this.props.shots && this.props.shots.length) {
       for (const shot of this.props.shots) {
-        children.push(<Card shot={shot} downloadUrl={this.props.downloadUrls[shot.id]} abTests={this.props.abTests} clipUrl={shot.urlDisplay} isOwner={this.props.isOwner} staticLink={this.props.staticLink} isExtInstalled={this.props.isExtInstalled} hasFxa={this.props.hasFxa} key={shot.id} />);
+        children.push(<Card shot={shot} downloadUrl={this.props.downloadUrls[shot.id]} abTests={this.props.abTests} clipUrl={shot.urlDisplay} isOwner={this.props.authenticated} staticLink={this.props.staticLink} isExtInstalled={this.props.isExtInstalled} hasFxa={this.props.hasFxa} key={shot.id} />);
```
ianb: We probably shouldn't have changed this to multiple lines a while ago

server/src/server.js
```diff
@@ -1065,15 +1073,16 @@ const END_POINT_SEPARATOR = "|";
...
-  if (!req.deviceId) {
-    next(errors.missingSession());
-    return;
-  }
-
   try {
     const stateBytes = await randomBytes(32);
     let state = stateBytes.toString("hex");
-    const inserted =  await setState(req.deviceId, state);
+    const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("auth")});
```
ianb: We should create a new scope here instead of using `"auth"`. That just requires a different string for `getKeygrip()` and then listing that string in dbschema.js

server/src/server.js
```diff
@@ -1065,15 +1073,16 @@ const END_POINT_SEPARATOR = "|";
...
+    const stateId = req.deviceId || String(Math.random());
```
ianb: We should generate the stateId with a UUID instead of Math.random() (I think this was left over from my extra hacky beginning work)

server/src/server.js
```diff
@@ -1109,7 +1113,9 @@ app.get("/api/fxa-oauth/confirm-login", async function(req, res, next) {
   const data = state.split(END_POINT_SEPARATOR, 2);
   const endpoint = data[1];
 
-  const isValid = await checkState(req.deviceId, data[0]);
+  const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("auth")});
```
ianb: and then the keygrip scope needs to be updated here too

server/src/servershot.js
```diff
@@ -387,45 +389,30 @@ class ServerClip extends AbstractShot.prototype.Clip {
...
-    return shot;
-  });
-};
-
-// FIXME: What is the difference between Shot.getFullShot and Shot.get?
-Shot.getFullShot = function(backend, id) {
+Shot.get = async function(backend, id, deviceId, accountId) {
   if (!id) {
-    throw new Error("Empty id: " + id);
+    return null;
```
ianb: Why this change?

test/server/pglib.py
```diff
@@ -35,6 +35,31 @@ def attach_device(device_id, account_id):
...
+def get_device_id(account_id):
+    dbname = pg_vars["PGDATABASE"] or pg_vars["PGUSER"]
+    conn = psycopg2.connect(dbname=dbname, user=pg_vars["PGUSER"], host=pg_vars["PGHOST"], port=pg_vars["PGPORT"])
+    conn.autocommit = True
+    cur = conn.cursor()
+    cur.execute("SELECT id FROM devices WHERE accountid = '" + account_id + "'")
+    data = cur.fetchall()
+    cur.close()
+    conn.close()
+    return {"deviceid": row[0] for row in data}
```
ianb: Shouldn't this be `[row[0] for row in data]` ? I don't think this should return a dictionary.

test/server/test_account_auth.py
```diff
@@ -0,0 +1,175 @@
...
+def test_not_owned_shot_with_fxa_auth():
+    setup_session_1 = setup_shot_on_device()
+    shot_url = setup_session_1['shot_url']
+
+    setup_session_2 = setup_shot_on_device()
+    browser_only_user = ScreenshotsBase()
+
+    # Use setup_session_2 accountId to login to a browser only session
+    result = browser_only_user.login(setup_session_2['user'].accountId)
+    assert setup_session_2['user'].deviceId == result['deviceid'], "Assosciated deviceId should match"
```
ianb: Assosciated should be Associated

test/server/test_account_auth.py
```diff
@@ -0,0 +1,175 @@
...
+
+# test we can change expiration and delete shots made previously
+def test_owned_shot_ui_with_fxa_auth():
+    setup_session = setup_shot_on_device()
+    shot_url = setup_session['shot_url']
+    browser_only_user = ScreenshotsBase()
+
+    # Use setup user accountId to login to a browser only session
+    result = browser_only_user.login(setup_session['user'].accountId)
+    assert setup_session['user'].deviceId == result['deviceid'], "Assosciated deviceId should match"
```
ianb: *Associated

server/src/pages/shot/shotpage-header.js
punamdahiya: That's correct, we are showing login button on owned shots for Firefox users, as documented in [design doc]([url](https://docs.google.com/document/d/1wxWNqHlQLUoeDrKqI5mUOSrhW-p2SN4_xwB2hRPk02k/edit#))

server/src/pages/shotindex/view.js
punamdahiya: Updated to make it readable again

server/src/server.js
punamdahiya: Updated to use 'fxa-oauth' as new scope

server/src/server.js
punamdahiya: done

server/src/servershot.js
punamdahiya: we are handling null return in calling functions, reverted back as we can keep throw error here as it does no harm.

test/server/pglib.py
punamdahiya: row[0] for row in data is returning string here

punamdahiya:
@ianb Updated PR with review feedback and have squashed commits. 

@g-k With this PR, we are allowing screenshots server site to be [accessible from non-firefox browser](https://docs.google.com/document/d/1wxWNqHlQLUoeDrKqI5mUOSrhW-p2SN4_xwB2hRPk02k/edit#). We are removing dependency on screenshots extension to authenticate user (deviceId).  User can login to non-Firefox browser ( both mobile and desktop) using firefox account  and is considered authenticated and can access their shots and respective shot detail pages. Including  for helping with security review of this PR. Thanks!

test/server/test_account_auth.py
punamdahiya: Got 2/3 :), will update before landing

g-k:
Hey, I was on PTO last Friday. Planning to get to this later today.

test/server/clientlib.py
```diff
@@ -7,58 +7,34 @@
...
-            urljoin(self.backend, "/api/login"),
-            data=dict(deviceId=self.deviceId, secret=self.secret, deviceInfo=json.dumps(self.deviceInfo)))
-        if resp.status_code == 404:
-            resp = self.session.post(
-                urljoin(self.backend, "/api/register"),
-                data=dict(deviceId=self.deviceId, secret=self.secret, deviceInfo=json.dumps(self.deviceInfo)))
-            account_info = attach_device(self.deviceId, self.accountId)
+    def login(self, account_id):
+        # Logs the session in by first checking if a device is
+        # assosciated with account_id and updating session with account cookie
```
g-k: nit: typo in associated

server/src/pages/settings/server.js
```diff
@@ -6,7 +6,7 @@ const app = express();
 exports.app = app;
 
 app.get("/", function(req, res) {
-  if (!req.deviceId) {
+  if (!req.deviceId && !req.accountId) {
```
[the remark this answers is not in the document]
punamdahiya: yes, it's possible on Firefox browser after a user login successfully with Firefox account , a request can have both device and account IDs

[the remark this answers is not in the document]
ianb: Note that these particular checks are just validation checks, not permission/authorization checks. If we erased the condition entirely it wouldn't (well, shouldn't) be a security error. This check just lets us fail slightly earlier with a clearer response code.

server/src/server.js
```diff
@@ -1109,23 +1124,23 @@ app.get("/api/fxa-oauth/confirm-login", async function(req, res, next) {
   const data = state.split(END_POINT_SEPARATOR, 2);
   const endpoint = data[1];
 
-  const isValid = await checkState(req.deviceId, data[0]);
-  if (!isValid) {
-    throw errors.badState();
-  }
+  const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("fxa-oauth")});
+  const stateId = cookies.get("fxaState") || req.deviceId;
```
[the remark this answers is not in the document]
ianb: We don't validate them (though we do validate a signature). That's probably a reasonable defense-in-depth. This is the middleware where we would add that check: https://github.com/mozilla-services/screenshots/blob/master/server/src/server.js#L189-L236

server/src/server.js
```diff
@@ -1065,15 +1078,22 @@ const END_POINT_SEPARATOR = "|";
...
-
   try {
     const stateBytes = await randomBytes(32);
     let state = stateBytes.toString("hex");
-    const inserted =  await setState(req.deviceId, state);
+    const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("fxa-oauth")});
+    let stateId = null;
+
+    if (req.deviceId) {
+      stateId = req.deviceId;
```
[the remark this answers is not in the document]
ianb: Per the next comment, adding the check to the authentication middleware would solve this

server/src/server.js
```diff
@@ -221,11 +221,19 @@ app.use(function(req, res, next) {
     }
     authInfo.abTests = abTests;
   }
+
   if (authInfo.deviceId) {
+    if (!isValidDeviceId(authInfo.deviceId)) {
+      next(new Error("Bad deviceId in login"));
```
ianb: This doesn't actually abort the request. I think just `throw new Error(...)` would work fine here? Otherwise this should be followed by a return.

server/src/server.js
```diff
@@ -221,11 +221,19 @@ app.use(function(req, res, next) {
...
+    }
     req.deviceId = authInfo.deviceId;
     req.userAnalytics = ua(config.gaId, req.deviceId, {strictCidFormat: false});
   }
+
   if (authInfo.accountId) {
+    if (authInfo.accountId.search(/^[a-zA-Z0-9]{1,255}$/) === -1) {
+      next(new Error("Bad accountId in login"));
```
ianb: ditto here

server/src/server.js
punamdahiya: Updated by following with a return

server/src/server.js
punamdahiya: done

server/src/server.js
```diff
@@ -1125,7 +1136,13 @@ app.get("/api/fxa-oauth/confirm-login", async function(req, res, next) {
   const endpoint = data[1];
 
   const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("fxa-oauth")});
-  const stateId = cookies.get("fxaState") || req.deviceId;
+
+  let stateId = cookies.get("fxaState");
+  if (stateId && (stateId.search(/^[a-zA-Z0-9_-]{1,255}$/) === -1)) {
+    const err = new Error("Bad stateId in confirm-login");
+    next(err);
```
punamdahiya: should add return here as well, will update

punamdahiya: done

punamdahiya:
@g-k @ianb PR is updated with the feedback, if it looks good will appreciate if you can give your review plus so that we can merge and have QA start testing. Thanks!

server/src/pages/settings/server.js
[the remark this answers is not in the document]
ianb: We debated this in a standup, and decided that all the cures were worse than the disease, and it's clearer to leave in this more direct approach for now.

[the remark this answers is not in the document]
punamdahiya: Updated to use easier to read !(req.deviceId || req.accountId)

punamdahiya:
Final squashed commit before landing on master, waiting for test run to finish

PR #4983, opened 2018-10-02
Fixes login, show images and API calls 401 errors

punamdahiya:


punamdahiya:
@ianb Made changes on top of  WIP patch that  fixes
a) 401 error seen in
/api/set-expiration ( while setting favorite on shot page)
/api/save-edit ( Save shots after edit)
/api/set-title/ (Edit shot title)

b) display images on myshots and shot page

c) SignIn in header and Signout from settings from non-Firefox browser

server/src/server.js
```diff
@@ -778,15 +778,15 @@ app.post("/api/delete-shot", function(req, res) {
...
+  if (!req.deviceId && !req.accountId) {
     sendRavenMessage(req, "Attempt to disconnect without login");
     simpleResponse(res, "Not logged in", 401);
     return;
   }
   disconnectDevice(req.deviceId).then((result) => {
     const keygrip = dbschema.getKeygrip("auth");
     const cookies = new Cookies(req, res, {keys: keygrip});
-    if (result) {
+    if (result || req.accountId) {
```
ianb: We could remove this if statement entirely and simply always clear the account cookie. If the statement isn't true then no response is sent, so the logic wasn't right to start with.

server/src/servershot.js
```diff
@@ -331,7 +331,7 @@ class Shot extends AbstractShot {
 Shot.getRawBytesForClip = function(uid) {
   return db.select(
     `SELECT images.url, images.contenttype, data.deviceid, devices.accountid
-     FROM images, devices
+     FROM devices, images
```
ianb: I don't think it matters either way, but any reason these got switched?

chenba: Otherwise it's `devices JOIN data ON images.shotid = data.id`

punamdahiya:
Updated with feedback, merging with the sketch branch

PR #4984, opened 2018-10-02
Enable favorite shots on My Shots

### LINKED ISSUE #4934: [FxA] Should be able to favorite shots from MyShots page
Scope of this issue is to make favicon clickable when user has signed in to FxA as discussed here https://github.com/mozilla-services/screenshots/issues/4791#issuecomment-423024784

chenba:
Fixes #4934

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -328,6 +328,8 @@ These are events that an add-on user can encounter on a shot they own
 6. [x] Right-click (or get the context menu) anywhere on the page `contextmenu/background`, `contextmenu/shot-tile`, `contextmenu/search`, or `contextmenu/header` depending on where the user clicks.
 7. [x] Click download from tile: `web/download/myshots-tile`
 8. [x] Clear search with button: `web/clear-search/button`
+8. [x] Click to favorite: `web/favorite/myshots-tile`
+8. [x] Click to un-favorite: `web/unfavorite/myshots-tile`
 
 #### Settings (My Shots)
 
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -315,6 +315,7 @@ shotIndexFavoriteIcon =
     .title = This is a favorite shot and it does not expire
 shotIndexSyncedShot =
     .title = Shot taken on another device
+shotIndexAlertErrorFavoriteShot = Error updating favorite shot status
 
 ## Delete Confirmation Dialog
 
--- a/server/src/pages/shotindex/controller.js
+++ b/server/src/pages/shotindex/controller.js
@@ -20,6 +20,7 @@ exports.launch = function(m) {
       m.shots = m.shots.map((shot) => {
         const s = new AbstractShot(m.backend, shot.id, shot.json);
         s.expireTime = shot.expireTime;
+        s.isFavorite = !shot.expireTime;
         s.isSynced = shot.isSynced;
         return s;
       });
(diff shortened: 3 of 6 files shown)


locales/en-US/server.ftl
```diff
@@ -315,6 +315,7 @@ shotIndexFavoriteIcon =
     .title = This is a favorite shot and it does not expire
 shotIndexSyncedShot =
     .title = Shot taken on another device
+shotIndexAlertErrorFavoriteShot = Error updating favorite shot status
```
chenba: Not sure about this message.  Perhaps @johngruen can advise.  It's similar to [the one on the shot page](https://github.com/mozilla-services/screenshots/blob/master/locales/en-US/server.ftl#L93).  (That one probably should be updated?)  It's used for favorite _and_ un-favorite actions.

server/src/pages/shotindex/controller.js
```diff
@@ -173,6 +173,46 @@ exports.deleteShot = function(shot) {
...
+  try {
+    const resp = await fetch(url, {
+      method: "POST",
+      body,
+    });
+
+    if (!resp.ok) {
+      window.Raven.captureException(
+        new Error(`Error calling /api/set-expiration: ${resp.status} ${resp.statusText}`));
+      throw new Error();
```
ianb: Usually I do: `let exc = new Error(...); window.Raven.captureException(exc); throw exc;`

server/src/pages/shotindex/view.js
```diff
@@ -148,6 +148,9 @@ class Body extends React.Component {
   renderErrorMessages() {
     return (
       <div>
+        <Localized id="">
```
ianb: Why do the other Localized's have ids and this one doesn't?

server/src/pages/shotindex/controller.js
```diff
@@ -173,6 +173,46 @@ exports.deleteShot = function(shot) {
...
+      shot.expireTime = Date.now() + newExpiration;
+    } else {
+      shot.expireTime = null;
+    }
+
+    render();
+  } catch (err) {
+    // The error could be one from the server or a network level error from
+    // fetch; we display a generic message either way as it's (probably?)
+    // friendlier than percolating a fetch error message up to the user.
```
ianb: Probably good to do `console.warn("Error updating expiration:", err);` so it isn't entirely swallowed

server/src/pages/shotindex/view.js
chenba: D'oh!  I deleted the id of some pasted code but didn't insert the new one.

docs/METRICS.md
```diff
@@ -328,6 +328,8 @@ These are events that an add-on user can encounter on a shot they own
 6. [x] Right-click (or get the context menu) anywhere on the page `contextmenu/background`, `contextmenu/shot-tile`, `contextmenu/search`, or `contextmenu/header` depending on where the user clicks.
 7. [x] Click download from tile: `web/download/myshots-tile`
 8. [x] Clear search with button: `web/clear-search/button`
+8. [x] Click to favorite: `web/favorite/myshots-tile`
+8. [x] Click to un-favorite: `web/unfavorite/myshots-tile`
```
chenba: This is new since the previous review.  Now it track seting and unsetting a shot as a favorite with separate events.

server/src/pages/shotindex/view.js
```diff
@@ -463,6 +468,16 @@ class Card extends React.Component {
     this.downloadButton.blur();
     sendEvent("download", "myshots-tile");
   }
+
+  onClickFavorite(shot) {
+    if (shot.isFavorite) {
```
chenba: `isFavorite` is set on each shot after it's instantiated from the JSON data and updated in `toggleFavoriteShot`.  Could consider changing it to a function in shot that returns a bool based on expiration, similar to what was done for `isOwner`.

PR #4994, opened 2018-10-03
Set useIsolating option for Fluent bundles

### LINKED ISSUE #4863: Page title has an extra [FSI] Unicode character "Screenshot: [FSI]..." in Windows taskbar name
When viewing a Screenshot page (such as https://screenshots.firefox.com/t286AJZDpqHVxol2/localhost) on Windows 10 (in Firefox, Chrome, or Edge), the Windows 10 taskbar shows the page title with an extra `FSI` Unicode character (U+2068 FIRST STRONG ISOLATE?) like "Screenshot: [FSI]..."

screenshot_fsi

chenba:
Fixes #4863

### DIFF
--- a/server/src/reactruntime.js
+++ b/server/src/reactruntime.js
@@ -8,13 +8,15 @@ require("fluent-intl-polyfill/compat");
 const { FluentBundle } = require("fluent/compat");
 const { LocalizationProvider } = require("fluent-react/compat");
 const { getLocaleMessages } = require("./locale-messages");
+const bundles = [];
 
 function generateBundles(messages, locales) {
-  const bundles = [];
   for (const locale of locales) {
-    const bundle = new FluentBundle(locale);
-    bundle.addMessages(messages[locale]);
-    bundles.push(bundle);
+    if (!messages[locale]) {
+      const bundle = new FluentBundle(locale);
+      bundle.addMessages(messages[locale]);
+      bundles.push(bundle);
+    }
   }
   return bundles;
 }
--- a/server/src/l10n.js
+++ b/server/src/l10n.js
@@ -46,12 +46,19 @@ exports.init = function(localeStringMap) {
   return initPromise;
 };
 
-exports.getText = function(locales) {
+exports.getText = function(locales, userAgent) {
   const bundles = {};
   const availableLocales = exports.getUserLocales(locales);
 
+  // This is a temporary fix for https://github.com/mozilla-services/screenshots/issues/4863
+  // to be reverted in https://github.com/mozilla-services/screenshots/issues/4993
+  let useIsolating = true;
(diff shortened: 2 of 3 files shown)


server/src/reactruntime.js
```diff
@@ -8,13 +8,15 @@ require("fluent-intl-polyfill/compat");
...
-  const bundles = [];
   for (const locale of locales) {
-    const bundle = new FluentBundle(locale);
-    bundle.addMessages(messages[locale]);
-    bundles.push(bundle);
+    if (!messages[locale]) {
+      const bundle = new FluentBundle(locale);
+      bundle.addMessages(messages[locale]);
+      bundles.push(bundle);
+    }
```
chenba: This is not part of the fix for #4863.  Both `HeadTemplate` and `BodyTemplate` call `generateBundles` and there's no reason to new up the bundles twice.

punamdahiya:
@chenba @Ian Reverting this fix as it has caused regression see #5022, #5020, #5013, #5012 

@stasm @flodolo Including for inputs on why using useIsolating option caused localization to fail

flodolo:
> @stasm @flodolo Including for inputs on why using useIsolating option caused localization to fail

Hopefully Stas will be able to help.

server/src/reactruntime.js
stasm: This change caused the regressions which led to the backing out of this PR in #5025. `messages[locale]` is always truthy because it contains the translations prepared by the server. In consequence, the runtime code never built any `FluentBundles` at all and the dynamic l10n was broken across the app. It only showed up in those places where the server-side rendering didn't already localize the content.

I think this code should look like this instead:

```js
const bundles = new Map();

function generateBundles(messages, locales) {
  for (const locale of locales) {
    if (!bundles.has(locale)) {
      const bundle = new FluentBundle(locale);
      bundle.addMessages(messages[locale]);
      bundles.set(locale, bundle);
    }
  }
  return bundles.values();
}
```

I tested this and it worked fine. I guess the code should also set `useIsolating: false` here, if needed?

However, I'm not sure if caching `bundles` like this is a good idea. Are there scenarios in which `HeadTemplate` and `BodyTemplate` can be re-rendered on clientside? One such scenario would be the user changing their language preferences, but I don't think that's supported by the app right now anyways.

In the future, I think a better solution would be to group translations into two FTL files, one for `HeadTemplate` and the other for `BodyTemplate`. Or maybe use a single `LocalizationProviders` higher up in the render tree, so that both the head and the body can use it. Or at least call `generateBundles` higher up and pass it down to `LocalizationProviders` in the head and in the body via props.

punamdahiya: Thanks @stasm for debugging and helping clarify. Will create a PR to use the fix.

PR #4998, opened 2018-10-04
Fixes #4974 - Remove wantsauth dependency for setting FxA cookie

### LINKED ISSUE #4974: [FxA] Do not rely on wantsauth for FxA cookie-setting
Per some of the comments in #4966, we should be setting the accountid cookie in the `confirm-login` endpoint. We should also remove the special `?auth=1` handling, which is only necessary because it's triggering wantsauth.

punamdahiya:


server/src/reactruntime.js
```diff
@@ -36,9 +36,8 @@ exports.HeadTemplate = class HeadTemplate extends React.Component {
     for (const locale of this.props.userLocales) {
       localeScripts.push(<script key={`l10n-${locale}`} src={this.props.staticLink(`/static/locales/${locale}.js`)} />);
     }
+    const wantsAuth = !this.props.authenticated;
```
punamdahiya: @chenba Including you this updated wantsAuth include criteria. With this change we will be loading wantsAuth on server only when deviceId doesn't exist

punamdahiya:
Rest of the changes in PR are
a) Updating login and confirm login route handlers to use async -await
b)  create accountId cookie on successful firefox account authentication

I am keeping isOwner changes separate from this PR which need a new attribute added to shot and believe should be a separate PR.

server/src/reactruntime.js
chenba: We also don't want `wantsAuth` when the user is already authenticated through FxA.

server/src/server.js
```diff
@@ -605,6 +611,7 @@ app.post("/api/login", function(req, res) {
         return sendParams.accountId = accountId;
       }).then((accountId) => {
         if (vars.ownershipCheck) {
+
```
chenba: This feels unnecessary.  😃

server/src/server.js
```diff
@@ -1091,10 +1098,13 @@ app.get("/api/fxa-oauth/login/*", function(req, res, next) {
...
-  }).catch(next);
+  } catch (err) {
+    mozlog.warn("fxa-oauth-login-failed", {err});
+    next(err);
+  }
 });
 
-app.get("/api/fxa-oauth/confirm-login", function(req, res, next) {
+app.get("/api/fxa-oauth/confirm-login", async function(req, res, next) {
   if (!req.deviceId) {
```
chenba: Wouldn't this `if` stop what we're  trying to do right away?

chenba: Oh, maybe I'm confusing this particular PR with the overall goal of having FxA work without device id?

punamdahiya: Yes,  this check will change to include accountId with PR that allows shots access and authentications on non-Firefox browser

PR #5000, opened 2018-10-05
Define button color

guoyunhe:
In **Linux desktop with dark theme**, default button text color is white or light gray. If button background is set to a light color, button text are not visible. So button text color and background color must both be defined.

Before:

screenshot_20181005_120700

After:

screenshot_20181005_120727


6a68:
@guoyunhe Thanks for the pull request. What linux distro and window manager are you using?

guoyunhe:
@6a68 openSUSE Tumbleweed + KDE 5.14.0 with Breeze Dark theme.

Issue #4989, opened 2018-10-03
[FxA] Update headers to use abstracted isOwner check


punamdahiya:
Remove passing accountId  explicitly to headers.  Headers isOwner property should have check for both deviceId or accountId making hasFxA property redundant.

chenba:
@punamdahiya I don't see `accountId` being passed in the header code.  Am I missing something?

ianb:
Fixed as part of #4966

PR #5003, opened 2018-10-05
Fixes #5001 - Shot page not rendering on non-firefox once account is …

### LINKED ISSUE #5001: [FxA] Page not found on non-Firefox for !owned shot
With #4966 allowing shots access on non-Firefox browser, user can open shot page in both owned and !owned state. 

If a user disconnect on Firefox, shot page link on non-Firefox authenticated state goes in a bad state and shows page not found. Deleting accountId cookie on non-firefox fixes page not found error

STR
1) Take a shot in nightly, and login with Fxa
2) Login with Fxa in !Firefox browser to access shots
3) Open the shot page in owned state
4) Disconnect on nightly
5) Refresh shot page in !Firefox gives page not found

punamdahiya:
…disconnected from device

### DIFF
--- a/server/src/server.js
+++ b/server/src/server.js
@@ -809,6 +809,12 @@ app.post("/api/set-title/:id/:domain", function(req, res) {
       simpleResponse(res, "No such shot", 404);
       return null;
     }
+
+    if (!shot.isOwner) {
+      simpleResponse(res, "Attempt to set title on shot user doesn't own", 403);
+      return null;
+    }
+
     shot.userTitle = userTitle;
     if (shot.openGraph && shot.openGraph.title) {
         shot.openGraph.title = userTitle;
@@ -845,6 +851,10 @@ app.post("/api/save-edit", function(req, res) {
     if (!shot) {
       sendRavenMessage(req, "Attempt to edit shot that does not exist");
       simpleResponse(res, "No such shot", 404);
+      return null;
+    }
+    if (!shot.isOwner) {
+      simpleResponse(res, "Attempt to edit shot user doesn't own", 403);
       return null;
     }
     const name = shot.clipNames()[0];
--- a/server/src/servershot.js
+++ b/server/src/servershot.js
@@ -393,50 +393,30 @@ class ServerClip extends AbstractShot.prototype.Clip {
 
 Shot.prototype.Clip = ServerClip;
 
-Shot.get = function(backend, id, deviceId, accountId) {
-  return Shot.getRawValue(id, deviceId, accountId).then((rawValue) => {
-    if (!rawValue) {
-      return null;
-    }
-    const json = JSON.parse(rawValue.value);
-    const jsonTitle = json.userTitle || (json.openGraph && json.openGraph.title) || json.docTitle;
-    json.docTitle = jsonTitle || rawValue.title;
-    if (!json.url && rawValue.url) {
-      json.url = rawValue.url;
-    }
-    // FIXME: I am not at all confident about the logic here!
-    console.log("shot isOwner compute values", {userId: rawValue.userid, deviceId: deviceId, rawAccount: rawValue.accountId, accountId: accountId});
-    json.isOwner = rawValue.userid === deviceId || (accountId && rawValue.accountId === accountId);
-    const shot = new Shot(rawValue.userid, backend, id, json);
-    shot.urlIfDeleted = rawValue.url;
-    shot.accountId = rawValue.accountId;
-    shot.expireTime = rawValue.expireTime;
-    shot.deleted = rawValue.deleted;
-    shot.blockType = rawValue.blockType;
-    return shot;
-  });
-};
-
-// FIXME: What is the difference between Shot.getFullShot and Shot.get?
-Shot.getFullShot = function(backend, id) {
+Shot.get = async function(backend, id, deviceId, accountId) {
   if (!id) {
-    throw new Error("Empty id: " + id);
+    return null;
   }
-  return db.select(
-    `SELECT value, deviceid, devices.accountid FROM data, devices
-    WHERE data.deviceid = devices.id AND data.id = $1`,
-    [id]
-  ).then((rows) => {
-    if (!rows.length) {
-      return null;
-    }
-    const row = rows[0];
-    const json = JSON.parse(row.value);
-    // FIXME: needs ownerId set
-    const shot = new Shot(row.userid, backend, id, json);
-    shot.accountId = row.accountid;
-    return shot;
-  });
+
+  const rawValue = await Shot.getRawValue(id);
+  if (!rawValue) {
+    return null;
+  }
+  const json = JSON.parse(rawValue.value);
+  const jsonTitle = json.userTitle || (json.openGraph && json.openGraph.title) || json.docTitle;
+  json.docTitle = jsonTitle || rawValue.title;
+  if (!json.url && rawValue.url) {
+    json.url = rawValue.url;
+  }
+
+  const shot = new Shot(rawValue.userid, backend, id, json);
+  shot.isOwner = rawValue.userid === deviceId || (accountId && rawValue.accountId === accountId);
+  shot.urlIfDeleted = rawValue.url;
+  shot.accountId = rawValue.accountId;
+  shot.expireTime = rawValue.expireTime;
+  shot.deleted = rawValue.deleted;
+  shot.blockType = rawValue.blockType;
+  return shot;
 };
 
 Shot.getRawValue = function(id, deviceId, accountId) {
--- a/shared/shot.js
+++ b/shared/shot.js
@@ -203,7 +203,6 @@ class AbstractShot {
     this.thumbnail = attrs.thumbnail || null;
     this.abTests = attrs.abTests || null;
     this.firefoxChannel = attrs.firefoxChannel || null;
-    this.isOwner = attrs.isOwner === undefined ? undefined : !!attrs.isOwner;
     this._clips = {};
     if (attrs.clips) {
       for (const clipId in attrs.clips) {
@@ -334,9 +333,6 @@ class AbstractShot {
   }
 
   get isOwner() {
-    if (this._isOwner === undefined) {
-      throw new Error("isOwner has not been set");
-    }
     return this._isOwner;
   }
 
@@ -583,7 +579,7 @@ class AbstractShot {
 AbstractShot.prototype.REGULAR_ATTRS = (`
 origin fullUrl docTitle userTitle createdDate images
 siteName openGraph twitterCard documentSize
-thumbnail abTests firefoxChannel isOwner
+thumbnail abTests firefoxChannel
 `).split(/\s+/g);
 
 // Attributes that will be accepted in the constructor, but ignored/dropped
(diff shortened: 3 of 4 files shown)


punamdahiya:
@ianb submitting PR to fix page not found error in #5001 . This also updates logic to set shot isOwner property by checking if account is linked to device in devices table

server/src/servershot.js
```diff
@@ -393,8 +393,16 @@ class ServerClip extends AbstractShot.prototype.Clip {
...
+Shot.get = async function(backend, id, deviceId, accountId) {
+  try {
+    // Check for ownership before requesting rawValue to verify if account
+    // is still connected to device
+    const isOwner = await Shot.checkOwnership(id, deviceId, accountId);
+    if (!isOwner) {
+      accountId = null;
+    }
+
+    const rawValue = await Shot.getRawValue(id, deviceId, accountId);
```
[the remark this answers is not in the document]
punamdahiya: > But then these arguments could cause the load to fail. But... I think we just don't want to pass these arguments down. Admittedly that feels weird – what we mean by `Shot.get(backend, id, deviceId, accountId)` would then be very different than what we mean by `Shot.getRawValue(id, deviceId, accountId)`, and that's just very confusing.
> 
> BUT, the only caller that uses that is `Shot.get()`, so I suspect the right answer is to remove those arguments from `Shot.getRawValue()` entirely. AND we should change the meaning of `Shot.get()`. This is a case where it's a problem:
> 

One place where we are calling getRawValue directly outside of .get(
https://github.com/mozilla-services/screenshots/blob/master/server/src/server.js#L742

Will it be ok here to remove deviceId?

server/src/servershot.js
```diff
@@ -405,16 +413,19 @@ Shot.get = function(backend, id, deviceId, accountId) {
       json.url = rawValue.url;
     }
     // FIXME: I am not at all confident about the logic here!
-    console.log("shot isOwner compute values", {userId: rawValue.userid, deviceId: deviceId, rawAccount: rawValue.accountId, accountId: accountId});
-    json.isOwner = rawValue.userid === deviceId || (accountId && rawValue.accountId === accountId);
+    // Updating logic to use isOwner value from checkOwnership
+    json.isOwner = isOwner;
```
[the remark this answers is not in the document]
punamdahiya: That's correct, isOwner getting stored in database was causing the issue I mentioned in meeting today

shared/shot.js
```diff
@@ -203,7 +203,7 @@ class AbstractShot {
     this.thumbnail = attrs.thumbnail || null;
     this.abTests = attrs.abTests || null;
     this.firefoxChannel = attrs.firefoxChannel || null;
-    this.isOwner = attrs.isOwner === undefined ? undefined : !!attrs.isOwner;
+    this._isOwner = attrs.isOwner === undefined ? false : !!attrs.isOwner;
```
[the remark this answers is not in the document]
punamdahiya: If left undefined throwing error "isOwner not set" while taking shot from extension https://github.com/mozilla-services/screenshots/pull/4966/files#diff-a8b7d067b30e7dfc2ab7f8ba8f8b2b9dR338

shared/shot.js
```diff
@@ -334,9 +334,9 @@ class AbstractShot {
   }
 
   get isOwner() {
-    if (this._isOwner === undefined) {
+    /*  if (this._isOwner === undefined) {
```
punamdahiya: @ianb With this PR we are allowing isOwner to be undefined thast why commenting explicit throw of error here.

server/src/server.js
```diff
@@ -747,7 +747,7 @@ app.get("/data/:id/:domain", function(req, res) {
   }
   const shotId = `${req.params.id}/${req.params.domain}`;
   // FIXME: maybe we should allow for accountId here too:
-  Shot.getRawValue(shotId, req.deviceId).then((data) => {
+  Shot.getRawValue(shotId).then((data) => {
```
punamdahiya: Updating getRawValue to remove deviceId, is that ok?

punamdahiya:
@ianb Updated PR with feedback, if it looks good we can merge it in sketch-of-browser-only-login branch. Thanks!

server/src/server.js
6a68: How can you tell if the user owns the shot, if you aren't passing in the `deviceId` or the `accountId`?

server/src/server.js
```diff
@@ -813,7 +813,7 @@ app.post("/api/set-title/:id/:domain", function(req, res) {
     return;
   }
   Shot.get(req.backend, shotId, req.deviceId, req.accountId).then((shot) => {
-    if (!shot) {
+    if (!shot.isOwner) {
```
6a68: If I'm reading this code correctly, when `Shot.get()` is called, we don't know for sure if the shot actually exists--the `shotId` may be incorrect. If the shot isn't found, the returned `shot` will be null, so trying to access `shot.isOwner` will throw.

server/src/server.js
6a68: Hmm. At the top of this function (line 744) is "if not req.deviceId, return 404". I think that's not correct for the case where you are logging in from a non-Firefox browser, and have `accountId` but not `deviceId`, but I need to double-check where this is used.

At this particular line, I suspect we actually need to pass in `accountId` along with `deviceId` here, if you want to be able to call this from a non-firefox browser.

punamdahiya: This was the only place outside of Shot.get where we are using getRawValue , so wan't sure. I will update to bring check for deviceId and accountId inside getRawValue

server/src/servershot.js
```diff
@@ -205,6 +209,10 @@ class Shot extends AbstractShot {
         clipRewrites.rewriteShotUrls();
         const oks = clipRewrites.commands();
         const json = this.toJSON();
+        if (json.isOwner) {
+          delete json.isOwner;
```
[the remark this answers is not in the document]
punamdahiya: @6a68 I agree since we are no longer storing isOwner in database, we don't need isOwner as REGULAR_ATTR. [Here](https://github.com/mozilla-services/screenshots/pull/4966/files#diff-6db5e3f6a06e986e2aa9ab0f5c0170d1R23) it will anyways always return empty as we don't have json value in db. Will update to set it as property of shot itself, then we don't need to explicitly delete on update and insert

punamdahiya:
@6a68 open to suggestion on how to best abstract isOwner check.  Current PR resolves shot page throwing not found error in non-Firefox when disconnected from an accountId. We can use https://github.com/mozilla-services/screenshots/issues/4989 to discuss and implement req.user.owns(shot)

shared/shot.js
```diff
@@ -203,7 +203,7 @@ class AbstractShot {
     this.thumbnail = attrs.thumbnail || null;
     this.abTests = attrs.abTests || null;
     this.firefoxChannel = attrs.firefoxChannel || null;
-    this.isOwner = attrs.isOwner === undefined ? undefined : !!attrs.isOwner;
+    this._isOwner = attrs.isOwner === undefined ? undefined : !!attrs.isOwner;
```
6a68: Since the `isOwner` setter has the undefined check removed, I don't think we need to assign to `_isOwner` directly

shared/shot.js
6a68: Could you explain why this change is needed? Is there a particular error this works around?

server/src/servershot.js
```diff
@@ -404,20 +405,23 @@ Shot.get = function(backend, id, deviceId, accountId) {
     if (!json.url && rawValue.url) {
       json.url = rawValue.url;
     }
-    // FIXME: I am not at all confident about the logic here!
-    console.log("shot isOwner compute values", {userId: rawValue.userid, deviceId: deviceId, rawAccount: rawValue.accountId, accountId: accountId});
-    json.isOwner = rawValue.userid === deviceId || (accountId && rawValue.accountId === accountId);
+
     const shot = new Shot(rawValue.userid, backend, id, json);
+    shot.isOwner = rawValue.userid === deviceId || (accountId && rawValue.accountId === accountId);
```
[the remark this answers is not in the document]
punamdahiya: yes with this change, its bypassing constructor and no longer relying on JSON , which means I should remove [this](https://github.com/mozilla-services/screenshots/pull/4966/files#diff-a8b7d067b30e7dfc2ab7f8ba8f8b2b9dR206)

My understanding it makes sense to keep in JSON ( and create regular attribute) only if we want to save it in db

shared/shot.js
punamdahiya: This was failing in extension while taking shot

server/src/servershot.js
```diff
@@ -393,8 +393,9 @@ class ServerClip extends AbstractShot.prototype.Clip {
 
 Shot.prototype.Clip = ServerClip;
 
-Shot.get = function(backend, id, deviceId, accountId) {
-  return Shot.getRawValue(id, deviceId, accountId).then((rawValue) => {
+Shot.get = async function(backend, id, deviceId, accountId) {
+  try {
+    const rawValue = await Shot.getRawValue(id);
```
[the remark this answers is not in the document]
punamdahiya: > Oh, OK, I think I see what's going on here: by not passing the `deviceId` or `accountId` into `getRawValue`, the shot will always be retrieved. Then, in a second step, check ownership below (the `isOwner` stuff). Is that right?
> 
That's correct we are allowing getRawValue to return shot data irrespective of accountId or deviceId.
This is needed to show shots in correct owned or !owned state irrespective of accountId cookie in non-Firefox browsers

server/src/servershot.js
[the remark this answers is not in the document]
6a68: Oh, that's right! The options bundle is filtered by the `REGULAR_ATTRS`, that's what I was missing. Thanks

server/src/server.js
```diff
@@ -809,6 +809,12 @@ app.post("/api/set-title/:id/:domain", function(req, res) {
       simpleResponse(res, "No such shot", 404);
       return null;
     }
+
+    if (!shot.isOwner) {
+      simpleResponse(res, "Attempt to set title on shot user don't own", 403);
```
6a68: s/don't/doesn't/

server/src/server.js
```diff
@@ -847,6 +853,10 @@ app.post("/api/save-edit", function(req, res) {
       simpleResponse(res, "No such shot", 404);
       return null;
     }
+    if (!shot.isOwner) {
+      simpleResponse(res, "Attempt to edit shot user don't own", 403);
```
6a68: s/don't/doesn't

server/src/servershot.js
```diff
@@ -393,50 +393,30 @@ class ServerClip extends AbstractShot.prototype.Clip {
...
+  }
+  const json = JSON.parse(rawValue.value);
+  const jsonTitle = json.userTitle || (json.openGraph && json.openGraph.title) || json.docTitle;
+  json.docTitle = jsonTitle || rawValue.title;
+  if (!json.url && rawValue.url) {
+    json.url = rawValue.url;
+  }
+
+  const shot = new Shot(rawValue.userid, backend, id, json);
+  shot.isOwner = rawValue.userid === deviceId || (accountId && rawValue.accountId === accountId);
```
6a68: does seem a little nicer to do `shot.isOwner = shot.checkOwner(deviceId, accountId)`

6a68: eh, maybe inlining it here is best

PR #5004, opened 2018-10-05
Accessibility tests

kimberlythegeek:
running accessibility checks in python 2.7 and 3.5, against the home page of stage

Working on writing tests for the iframes--onboarding iframe, and possibly pre-selection and selection iframes. This requires a little more setup.

Currently, any failures will block the build. I need feedback on what your expectations are.
I will file issues for existing violations, and, if desired, mark those tests as expected failures to prevent blocking the build.

I'd appreciate any feedback you may have!


kimberlythegeek:
@6a68 @chenba @ianb @m8ttyB

kimberlythegeek:
Filed issues for existing violations.

kimberlythegeek:
FYI, this build will fail. I added new accessibility checks for the end-to-end workflow of taking a screenshot. I will file the issues after getting some feedback. 

(Also I'd like you guys to see the build and test output before xfailing anything)

The new tests run on the onboarding iframe, the preselection and selection iframes, and the screenshot page that follows.

I'm unsure if all of these checks are necessary--namely the preselection and selection iframes. I have not yet reviewed the violations that are being caught, I've been focusing on getting the tests working.

Also with this PR I've changed out xfails are configured. I modified `pytest-axe` so that xfails can be defined in a class-level dictionary `params`. This implementation could potentially change, but it removes the need for separate directories and `conftest.py` files for each test

kimberlythegeek:
Going to reset to an earlier commit while I work on the issues with circleci and the new tests.

kimberlythegeek:
Updated the locator for the "take screenshot" button. Now the failing tests correctly show current accessibility violations. Will check for duplicates, and add issues for new violations.

6a68:
@kimberlythegeek I think it'd be good to keep the addon (iframe) tests and server (homepage) tests in separate files. I'd be fine to land the server tests now, and the addon tests in a separate PR. Are the server tests ready for review?

kimberlythegeek:
@6a68 Sounds good. Yes, those tests are ready for review. Is the screenshot page considered a server test, too?

6a68:
@kimberlythegeek :+1: I'll take a look at the tests today. yeah, anything that tests a page on the screenshots website is a server test

6a68:
Actually, I think @ianb would be a better reviewer. He wrote the original python tests, and might have better feedback about how to integrate this into the existing test suite.

test/accessibility/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+def wait_for_element_to_load(driver, selector):
+    wait = WebDriverWait(driver, 10)
+    element = wait.until(EC.presence_of_element_located(selector))
+    return element
+
+
+@pytest.fixture(scope="class")
+def home_page():
+    """Launch Screenshots home page as guest."""
+    profile = webdriver.FirefoxProfile()
```
ianb: I'm pretty sure this is testing the version of Screenshots that packaged with Firefox, but won't test the development version. I.e., if you fix an error locally, this test won't see your fix.

We have code in our tests that handles this: https://github.com/mozilla-services/screenshots/blob/master/test/test.js#L54-L85 – if this was written in Node.js you could use that, though I suppose it could also be translated to Python.

.circleci/config.yml
```diff
@@ -78,13 +79,28 @@ jobs:
...
           command: |
             . "$NVM_DIR/nvm.sh"
             nvm use 8.10.0
             export PATH=~/firefox:$PATH
             FIREFOX_CHANNEL=NIGHTLY MOZ_HEADLESS=1 npm test
+      - run:
+          name: Run Accessibility Tests
+          command: |
+            pip install -U pip
+            pip install tox
```
ianb: These should be runnable locally, probably as `npm run test:accessibility` (`test:accessibility` is just a shell script listed in `package.json`)

It should use the same `make .venv` pattern that `test:server` uses. I suspect this will make the tox parts unnecessary. From there you just use things like `./.venv/bin/pytest ...` to access the specific tools you installed.

.circleci/config.yml
```diff
@@ -78,13 +79,28 @@ jobs:
             npm i -g get-firefox
             get-firefox --branch nightly --platform linux --extract --target ~/
             ~/firefox/firefox -v
+            wget -O /tmp/geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v$GECKODRIVER/geckodriver-v$GECKODRIVER-linux64.tar.gz
```
ianb: Is the version of geckodriver installed in `node_modules/.bin/geckodriver` usable for this? We're currently installed 0.21.0, but I don't know if that's intentional or if that was just the latest version when we set it up.

test/accessibility/test_server_accessibility.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+from pytest_axe.parameterize_tests import *  # NOQA
+
+
+class TestHomePageAccessibility(object):
+    params = {
```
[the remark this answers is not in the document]
kimberlythegeek: `params` are being accessed by the `pytest-axe` plugin. While it could be subject to change in the future, this is the current implementation that enables xfailing tests.

`pytest-axe` generates test functions at setup. The list of accessibility rules are fetched from aXe. This list is modified based on `params`, and used as input for `pytest_generate_tests`

https://github.com/mozilla-services/pytest-axe/blob/master/pytest_axe/parameterize_tests.py#L54-L62

test/accessibility/requirements.txt
```diff
@@ -0,0 +1,4 @@
+flake8==3.5.0
+pytest==3.8.2
+axe-selenium-python==2.1.1
+pytest-axe==1.1.3
```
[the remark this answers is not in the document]
kimberlythegeek: Alrighty 👌

test/accessibility/conftest.py
kimberlythegeek: Noted. I'll look at the code you shared.

.circleci/config.yml
kimberlythegeek: Alrighty. I've read through the code, but I wasn't exactly sure how to implement the `.venv` for my tests--perhaps we could schedule a meeting to go over some of changes you've suggested.

.circleci/config.yml
```diff
@@ -95,12 +91,10 @@ jobs:
           name: Run Accessibility Tests
           command: |
             pip install -U pip
-            pip install tox
+            pip install -r ./bin/require.pip
```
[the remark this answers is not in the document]
ianb: Unless... maybe this is relying on `pip` calling Python 3? Since it appears there's some Python 3 bits down there.

Technically switching the Python version is as simple as changing this line: https://github.com/mozilla-services/screenshots/blob/master/Makefile#L94

It's possible some server tests would break if that was done. Trying it out myself, I note that we need to upgrade to `sarge==0.1.5.post0`, then `rm -r .venv`, and then `npm run test:server`.

But there's a bunch of small issues, and now my server isn't running for some dumb unrelated reason. I put my changes in #5070, but I can't complete that as I have some other work that is going to take priority for the rest of this week.

package.json
```diff
@@ -101,6 +101,7 @@
     "test": "bin/test-helpers/make_addon_with_unavailable_server && mocha test/ test/server/unit/*",
     "test:selenium": "bin/test-helpers/make_addon_with_unavailable_server && mocha test/test.js",
     "test:server": "make .venv && make server && cd test/server && ../../.venv/bin/pytest",
+    "test:accessibility": "pytest test/accessibility --driver Firefox -srx --axe --verbose",
```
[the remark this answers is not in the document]
kimberlythegeek: I see in the `test:server` script it runs `make .venv` and `pytest`; I'll go with that approach for consistency

PR #5030, opened 2018-10-11
Fix and simplify Screenshots UI check test

6a68:
(This PR is based on advice from kmag, while the other recent PR was just importing upstream refactoring.)

It turns out there's no need to flip the pref and poll for the Screenshots page action to appear
or disappear. Instead, we can just directly enable or disable the addon via the
AddonManager. This won't conflict with the pref observing code, because
that code only reacts to pref changes, not to addon state changes.


6a68:
Rebased this one too. Hopefully the tests pass 🤞

PR #5032, opened 2018-10-11
Add new onboarding image to manifest

6a68:
The image shown in the new onboarding step (`onboarding-6.png`) should not be loaded by the webextension unless it's explicitly declared in the `web_accessible_resources` part of the manifest. I'm not sure why it's working, but it seems like an upstream bug; by adding it to the manifest, onboarding will continue to work when the bug's fixed.

### DIFF
--- a/webextension/manifest.json.template
+++ b/webextension/manifest.json.template
@@ -59,6 +59,7 @@
     "icons/onboarding-3.png",
     "icons/onboarding-4.png",
     "icons/onboarding-5.png",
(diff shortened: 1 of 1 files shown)


6a68:
Rebased. Hopefully the tests will now pass.

punamdahiya:
r+ You are right even without adding in manifest.json moz-extension://<extension-id>/icons/onboarding-6.png is loading correct.

PR #5036, opened 2018-10-13
A11y - Add lang attribte to <html> tag

### LINKED ISSUE #5015: <html> element does not have a lang attribute
[detailed info on html-has-lang rule](https://dequeuniversity.com/rules/axe/3.1/html-has-lang?application=axeAPI)
```
  Rule Violated:
  html-has-lang - Ensures every HTML document has a lang attribute
  	URL: https://dequeuniversity.com/rules/axe/3.1/html-has-lang?application=axeAPI
  	Impact Level: serious
  	Tags: cat.language wcag2a wcag311
  	Elements Affected:
  	1) Target: html
  		The <html> element does not have a lang attribute
```

yehudab:
Fixes #5015.


yehudab:
@flodolo, sorry you got marked as a reviewer. I think it was because I made a mistake and accidentally pushed translation changes as a result of a bad marge.
This was cleared last night, thanks to the help of @6a68.

To answer your question: Fluent is used to get the preferred language, the only problem is that it might return a list of languages, and the HTML standard allows just one to be declared as "the" language of the page. 
My code takes the first language in that list and assumes it is the main language.

ianb:
Thanks @yehudab !

Issue #5018, opened 2018-10-08
Maximum-scale on <meta> tag disables zooming on mobile devices


kimberlythegeek:
[detailed info on meta-viewport rule]( https://dequeuniversity.com/rules/axe/3.1/meta-viewport?application=axeAPI)
```
Rule Violated:
meta-viewport - Ensures <meta name="viewport"> does not disable text scaling and zooming
	URL: https://dequeuniversity.com/rules/axe/3.1/meta-viewport?application=axeAPI
	Impact Level: critical
	Tags: cat.sensory-and-visual-cues wcag2aa wcag144
	Elements Affected:
	1) Target: meta[name="viewport"]
		maximum-scale on <meta> tag disables zooming on mobile devices
```

yehudab:
Ready to take this, if there are no objections.

6a68:
Sure. I've sent you an invite to be a contributor on this repo. Once you accept, I can formally assign it to you (but I don't think anyone else will open a PR in the meantime)

yehudab:
Got the invite and accepted. Thanks!

PR #5046, opened 2018-10-17
Removed unneeded maximum-scale parameter.

yehudab:
DRY: Moved all meta-viewport declarations to a single location.
Fixes: #5018


yehudab:
Sure. Happy to help.

PR #5048, opened 2018-10-17
Fixes #5043 - Add download button to non-owner shot page

### LINKED ISSUE #5043: Re-add download button to non-owner page
In the redesign we've removed most controls for non-owner views (noted in #4979), but we'd like to reintroduce the download button for non-owners.

Since it's just one button it might look weird, so we're also seeking UX feedback.

punamdahiya:


6a68:
@punamdahiya Could you add a description of what this PR is supposed to do?

punamdahiya:
> @punamdahiya Could you add a description of what this PR is supposed to do?

@6a68 PR is bringing back download button on !owned shot page as shown here https://github.com/mozilla-services/screenshots/issues/5043#issuecomment-430080629

server/src/pages/shot/view.js
```diff
@@ -366,6 +366,14 @@ class Body extends React.Component {
...
     let copyButton = null;
 
+    downloadButton = <div className="download-shot-button">
+      <Localized id="shotPageDownloadShot" attrs={{title: true}}>
+        <button className={`button transparent nav-button`} onClick={this.onClickDownload.bind(this)}
+                title="Download the shot image">
+          <img src={this.props.staticLink("/static/img/icon-download.svg")} />
+        </button>
+      </Localized></div>;
+
```
punamdahiya: Moved downloadButton outside of ownership check, as want to display it for both owned and !owned shots

server/src/pages/shot/shotpage-header.js
```diff
@@ -107,7 +108,7 @@ exports.ShotPageHeader = class ShotPageHeader extends React.Component {
               hasFxa={this.props.isFxaAuthenticated}>
         { myShotsText }
         { shotInfo }
-        <div className="shot-alt-actions">
+        <div className={classnames({"shot-alt-actions": this.props.isOwner})}>
```
punamdahiya: For !owned shots, we want download button to show on the same row as shot title and don't need shot-alt-actions class that renders alt action buttons in separate row on medium and small window widths

punamdahiya:
@6a68 I am confused by suggested shotActions prop on ShotPageHeader, is it a boolean that sets if we want shotactions button (favorite, edit, copy, delete, settings) shown or not. 
Our buttons are defined and logic to show/hide is in view.js and gets rendered inside ShotPageHeader as props.children

punamdahiya:
@6a68 Taken a stab at keeping logic of owner only buttons and its styles together in view.js.  My understanding to bring logic of owner only palette inside ShotPageHeader  component, we need to bring in defining all buttons and its related methods inside the component. Let me know if I am missing a better approach here. Thanks

6a68:
> @6a68 I am confused by suggested shotActions prop on ShotPageHeader, is it a boolean that sets if we want shotactions button (favorite, edit, copy, delete, settings) shown or not.
Our buttons are defined and logic to show/hide is in view.js and gets rendered inside ShotPageHeader as props.children

Sorry, missed your question. We can keep the logic in view.js, and just have two different ways to pass rendered buttons into the ShotPageHeader: the `children` prop and another prop. I'll write a little code to explain what I had in mind

6a68:
@punamdahiya Here's a second attempt at an explanation: 

Instead of passing all the elements into the ShotPageHeader using the special `children` prop, have two props: one for buttons that behave differently at narrow widths (call it `shotActions`, maybe), and keep using the `children` prop for the buttons that don't wrap around (either the download button, or the sign in button).

Elaborating a little bit in code:

- Inside view.js, put the buttons for the owner-only actions in an array: 
```js
const shotActions = this.props.isOwner ?
  [favoriteShotButton, editButton, copyButton, downloadButton, trashOrFlagButton] : null
```
- Also inside view.js, pass `shotActions` into a new `shotActions` prop on the `ShotPageHeader`, and pass either the download button or the signin button as the `children`: 
```js
render() {
  return (
  <ShotPageHeader ... shotActions={shotActions}>
    {this.props.isOwner ? downloadButton : signInButton}
  </ShotPageHeader>
);
```
- Add the new `shotActions` prop to ShotPageHeader, and construct the wrapper div inside a helper method:
```js
  // inside ShotPageHeader.js 
  renderShotActions() {
    return (
      this.props.shotActions ?
        <div class="shot-alt-actions">{this.props.shotActions}</div> : null
    )
  }
```
- Inside the `render` method for `ShotPageHeader`, insert the `shotActions` in the appropriate spot, just before the `children`:
```js
  // inside the ShotPageHeader render method:
  const shotActions = this.renderShotActions();
  return (
       <Header shouldGetFirefox={this.props.shouldGetFirefox} isOwner={this.props.isOwner}
              hasFxa={this.props.isFxaAuthenticated}>
        { myShotsText }
        { shotInfo }
        { shotActions }
        { this.props.children }
      </Header>
```
This will give us a clear separation between elements that are rendered as a separate menu of buttons (the `shotActions`) and elements that are always rendered in the header (download / signin button).

I hope this clarifies what I was asking for earlier, but if it's still unclear, let me know :-)

server/src/pages/shot/shotpage-header.js
```diff
@@ -97,19 +97,26 @@ exports.ShotPageHeader = class ShotPageHeader extends React.Component {
...
     return (
       <Header shouldGetFirefox={this.props.shouldGetFirefox} isOwner={this.props.isOwner}
               hasFxa={this.props.isFxaAuthenticated}>
         { myShotsText }
         { shotInfo }
-        <div className="shot-alt-actions">
+        { shotActions }
         { this.props.children }
-        </div>
         { signin }
```
punamdahiya: Keeping signin  here as signin button is defined in shot page header component

server/src/pages/shot/view.js
```diff
@@ -366,6 +366,14 @@ class Body extends React.Component {
...
     if (this.props.isOwner) {
```
punamdahiya: This check isn't elegant but keeping it in favor of executing a block of code that's not needed for !owned shots

punamdahiya:
Thanks for catching console warnings! Updated PropTypes and key attributes in PR.

punamdahiya:
> Thanks for catching console warnings! Updated PropTypes and key attributes in PR.

keys docs shared above is informative. Good to know that key don't need to be unique globally. Updating with more meaningful key values

Issue #5054, opened 2018-10-19
"Sign in" text in upsell banner should be a link


6a68:
Seems like a missed opportunity to not linkify the "Sign in or sign up" text here:

<img width="585" alt="screen shot 2018-10-19 at 11 53 13 am" src="https://user-images.githubusercontent.com/96396/47238027-9e0c6380-d395-11e8-8ecd-964aca0cc858.png">

yehudab:
Thinking about it again, I guess a better URL would be: `/api/fxa-oauth/login/<current-url>`.

punamdahiya:
> /api/fxa-oauth/login

That's correct, its invokes [route handler here](https://github.com/mozilla-services/screenshots/blob/master/server/src/server.js#L1090)

PR #5056, opened 2018-10-20
Make the "sign in or sign up" clickable in banner

yehudab:
Fixes: #5054

### DIFF
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -15,7 +15,7 @@ buttonSignIn =
     .title = Sign In
 screenshotsLogo =
     .title = Screenshots Home
-bannerMessage = Sign in or sign up to access your shots across devices and save your favorites forever.
+bannerSignIn = <a>Sign in or sign up</a> to access your shots across devices and save your favorites forever.
 bannerUpsell = {gScreenshotsDescription} <a>Get Firefox now</a>
 
 ## Footer
--- a/server/src/ad-banner.js
+++ b/server/src/ad-banner.js
@@ -12,6 +12,10 @@ exports.AdBanner = class AdBanner extends React.Component {
     sendEvent("click-install-firefox-shot", {useBeacon: true});
   }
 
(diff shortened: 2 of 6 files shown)


locales/en-US/server.ftl
```diff
@@ -15,7 +15,7 @@ buttonSignIn =
     .title = Sign In
 screenshotsLogo =
     .title = Screenshots Home
-bannerMessage = Sign in or sign up to access your shots across devices and save your favorites forever.
+bannerMessage = <a>Sign in or sign up</a> to access your shots across devices and save your favorites forever.
```
flodolo: Sorry, this [kind of changes](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings) require a new string ID

6a68:
@yehudab Thanks for the PR. This bug hasn't been triaged yet (it's not in a milestone), so I'm not sure this change will actually land; we should get a product decision from John later this week.

In general, good-first-issue labels or stuff in the Stretch milestone is a safe bet if you're looking for ways to contribute :+1: (and I'll update our docs to make that clearer, sorry)

yehudab:
Thanks for the clarification. No worries if it doesn’t land soon. It was also a good practice for me to understand the l10n process and best practices.

server/src/ad-banner.js
```diff
@@ -25,9 +29,12 @@ exports.AdBanner = class AdBanner extends React.Component {
           </p>
         </Localized>;
     } else if (this.props.isOwner && !this.props.hasFxa) {
```
punamdahiya: With #4959 fix, condition here has changed to 

`} else if (!this.props.hasFxa) {`

yehudab:
@punamdahiya PR is now rebased and squashed

yehudab:
Documentation updated.

punamdahiya:
> Documentation updated.

Thank you! @yehudab

PR #5061, opened 2018-10-22
switch to audit-filter from npm package

g-k:
refs: https://github.com/mozilla-services/screenshots/pull/4948

NB: CI still fails because we haven't addressed some new advisories (they mostly look like lower priority ReDoSes).

r?


package.json
```diff
@@ -71,6 +72,7 @@
     "istanbul-middleware": "0.2.2",
     "mocha": "5.2.0",
     "node-sass": "4.9.4",
+    "npm": "^6.4.1",
```
g-k: npm (I believe from npm-run-all) is otherwise on 4.X, which doesn't support `npm audit --json`

package.json
```diff
@@ -53,6 +53,7 @@
   },
   "devDependencies": {
     "addons-linter": "1.3.6",
+    "audit-filter": "^0.3.0",
```
6a68: This should be the exact version, `"0.3.0"`

package.json
6a68: Hmm. I think the right way to do this would be to set the correct minimum npm version in the `engines` section, then maybe also add a `config` to set `engine-strict` to true? Might also want to bump the minimum node version in `engines` to the minimum for npm 6.4.1 to work; it's probably higher than 8.0.0, the current minimum.

6a68:
@g-k Thanks for the update, looks good. I'll deal with the package-lock issues

6a68:
Aha. Looks like the gigantic list of integrity updates from sha512 down to sha1 is a platform difference between linux and mac. Software is hard, I guess https://github.com/npm/npm/issues/17749

6a68:
Hmm, still seeing package-lock.json conflicts. I guess I'll blow away the entire file and try again for a second time.

ianb:
Note test failures are due to a regression in master

Issue #5034, opened 2018-10-12
Screenshots "Page not found" page is wrongly displayed on Safari and Chrome


SoftVision-CosminMuntean:
[Affected versions]:
- Chrome 69.0.3497.100
- Safari

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Open Chrome and navigate to a shot page that was previously deleted (eg: this shot).
2. Observe the "Page not found" page.

[Expected result]:
- The page is correctly displayed as in Firefox.

[Actual results]:
- The "Firefox Screenshots" title is displayed in the left center of the page and the specific image is very small and displayed at the bottom of the page.

[Notes]:
- The issue is not reproducible on Edge.
- Here is a screen shot of the issue:
![image](https://user-images.githubusercontent.com/19226534/46872327-73be2180-ce3c-11e8-9c42-80041b05139e.png)

ianb:
See also #5002

SoftVision-CosminMuntean:
I have verified this issue using the latest Screenshots (34.0.0) dev version and the issue is no longer reproducible. Tested on Chrome, Safari, Opera and Edge.

PR #5071, opened 2018-10-25
Fixes #5002 - UI fixes in non-Firefox

### LINKED ISSUE #5002: [FxA] Screenshots website UI fixes on non-Firefox browsers
Opening bug to track minor UI issues seen on non-Firefox

Chrome:
1) Page not found header height needs to be set
2) My Shots page - On hover over short thumbnail delete icon shows incorrect

punamdahiya:


punamdahiya:
Fixes #5034

server/src/pages/shot/text-tool.js
```diff
@@ -214,12 +214,19 @@ exports.TextTool = class TextTool extends React.Component {
     }
     const styles = window.getComputedStyle(this.textInput.current);
     const FONT_SIZE = parseInt(styles["font-size"], 10);
+    let line_height = styles["line-height"];
+
+    // Depending on the user agent computed styles
+    // line-height can have value as normal
+    if (line_height.toLowerCase() === "normal") {
+      line_height = 1.1 * FONT_SIZE;
```
punamdahiya: Using value 1.1[ (approximate default 1.2)](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#Values) to compute line-height for the fonts used in TextTool

punamdahiya:
Fixes #5002

punamdahiya:
Fixes #5106

punamdahiya:
Page not founc issue got fixed in  master with https://github.com/mozilla-services/screenshots/issues/5034, I will rebase PR on top of latest master, that should fix the issue

punamdahiya:
@6a68 Updated PR with master that should fix page not found, thanks for catching the stray tick updated to remove that in shot index view.

server/src/pages/shot/text-tool.js
6a68: Hmm. This special case doesn't cover other weird stuff, like ems or percentages, that might be set as a user agent's default. Can we just set the line-height in the CSS directly?

punamdahiya:
@6a68 AFAIK It’s hard to determine line-height for a font-size and set in CSS as it [varies across platform.](https://github.com/mozilla-services/screenshots/blob/master/server/src/pages/shot/text-tool.js#L218)

PR #5103, opened 2018-10-29
Fixes #4986 - Hide edit button on mobile

### LINKED ISSUE #4986: [FxA] Disable mobile annotations on prod/stage
See https://docs.google.com/document/d/1wxWNqHlQLUoeDrKqI5mUOSrhW-p2SN4_xwB2hRPk02k/edit?usp=sharing

On Mobile, Shot page will be shown without Edit icon

punamdahiya:


6a68:
@punamdahiya I'm not sure about this. Rather than just disable a feature, how about we have QA do some testing and see what the major issues are (if any)? Then we'll have a list of blockers, and an understanding of what work is needed to provide a working experience on mobile.

punamdahiya:
> @punamdahiya I'm not sure about this. Rather than just disable a feature, how about we have QA do some testing and see what the major issues are (if any)? Then we'll have a list of blockers, and an understanding of what work is needed to provide a working experience on mobile.

+1 it will be good to test and have a bug with results, updated meeting notes to request Cosmin edit flow QA on mobile

6a68:
Looks like we should indeed disable mobile editing until the issues collected under #5118 have been fixed. Merging

PR #5104, opened 2018-10-29
Fixes #5102 - Bad Oauth State with FxA login

### LINKED ISSUE #5102: [FxA] Bad Oauth state when login with both FxAState and user cookie
On Firefox, for new profiles we don't create deviceId  (user cookie)  till first screenshot upload. If a new user 
a) Goes on home page
b) Click SignIn button , we create FxAState cookie
c) Take a screenshot ( creates user cookie)
d) Click SignIn and try to complete login , get Bad Oauth state

punamdahiya:


punamdahiya:
Fixes #5102

server/src/server.js
```diff
@@ -1091,16 +1091,22 @@ app.get("/api/fxa-oauth/login/*", async function(req, res, next) {
   try {
     const stateBytes = await randomBytes(32);
     let state = stateBytes.toString("hex");
-    let stateId = null;
 
-    if (req.deviceId) {
+    const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("fxa-oauth")});
+    let stateId = cookies.get("fxaState");
+    if (stateId && (stateId.search(/^[a-zA-Z0-9_-]{1,255}$/) === -1)) {
```
ianb: Arguably it's clearer to say `/^[a-zA-Z0-9_-]{1,255}$/.test(stateId)`, but that's just a minor nit.

server/src/server.js
```diff
@@ -1091,16 +1091,22 @@ app.get("/api/fxa-oauth/login/*", async function(req, res, next) {
...
+
+    if (!stateId && req.deviceId) {
       stateId = req.deviceId;
     } else {
       const uuidPromise = util.promisify(genUuid.generate);
       const uuid = await uuidPromise(genUuid.V_RANDOM);
       stateId = uuid.toString();
 
-      const cookies = new Cookies(req, res, {keys: dbschema.getKeygrip("fxa-oauth")});
       cookies.set("fxaState", stateId, {signed: true});
```
ianb: So... this doesn't really make sense to me. I think the logic is:

* Get stateId from the cookie, make sure it's valid
* If there was no cookie stateId, then use deviceId
* If there is no deviceId, then throw away stateId and regenerate it

But we never actually *use* the stateId cookie we load. It always gets clobbered. The only thing we're doing is making sure *not* to use deviceId if they used stateId in the past... but we don't actually keep the stateId then.

server/src/server.js
6a68: @punamdahiya I'm not sure this is the right regexp to verify a UUID is well-formed

server/src/server.js
punamdahiya: That's correct, logic here checks for fxaState cookie and if found clobbers and regenerates fxaState not using the previous value. As discussed in standup, will update to delete fxaState cookie if deviceId is found and continue to use deviceId as stateId.

punamdahiya:
@ianb Updated PR to remove fxaState cookie when deviceId exists. Thanks!

PR #5119, opened 2018-10-31
Show expiration info on all screen sizes. Fixes: #5068

### LINKED ISSUE #5068: Expiration info should always be visible
Seems like this data should be visible at all widths, but it's hidden at 2 out of the 4:

image

image

image

image

yehudab:
Fixes #5068

The fix includes showing full expiration info on all sizes but `small`.
For `small`, the `time-diff` part is removed.

Example:
screenshot_2018-10-31 firefox privacy notice

### DIFF
--- a/static/css/frame.scss
+++ b/static/css/frame.scss
@@ -71,13 +71,9 @@
   .expire-info {
     display: block;
     @include respond-to("small") {
-      display: none;
-    }
-    @include respond-to("large") {
-      display: none;
-    }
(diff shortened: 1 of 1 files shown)


6a68:
@youwenliang What do you think: does it make sense to just show expiration, but not creation date, at the smallest widths?

youwenliang:
Looks good to me, I agree that expiration is more important than creation date.

PR #5125, opened 2018-11-02
Fixes #4790 - NonFxA Shot takers should be briefly onboarded

### LINKED ISSUE #4790: [FxA] Non-FxA shot takers should be briefly onboarded
This is underdefined in my spec at the moment. The first time a shooter sees the new shot UI, we should indicate that shots expire by default in two weeks, and that they can sign in to shave them forever or favorite them.

punamdahiya:


punamdahiya:
Updated with changes and ready for review. Thanks!

server/src/pages/shot/controller.js
```diff
@@ -21,7 +21,11 @@ function shouldHighlightEditIcon(model) {
 }
 
 function shouldShowPromo(model) {
```
6a68: This should really live in the view, not the controller, I think

punamdahiya:
@6a68 by page view , do you mean respective view.js for shot, shot index and homepage. fxA-onboarding-dialog is shown on all the pages where we have signin button and uses [signinbutton component render method ](https://github.com/mozilla-services/screenshots/pull/5125/files#diff-8e2054fb3aa47a68d032b565f47719dbR38)to invoke it. 

So if we are moving display logic in respective page views, we will need to pass the state from view to signin button using props, is that correct?

6a68:
@punamdahiya I had something a little different in mind. Let me get a patch together quickly and I'll open a PR into your PR

6a68:
Just reviewing what we discussed after the internal meeting today:

- This looks great so far, let's also have the onboarding manager contain the logic for when to show the edit tools onboarding card, and the logic for when to show the different AdBanner states
- To keep the PR from growing in size, just keep sending the existing props into existing components, like AdBanner, and replace conditional checks with a call to the onboarding manager 
- The onboarding manager ideally will contain all of the logic related to which promo pieces are shown, and when they are shown. By keeping the logic in one place, we'll be able to easily change the onboarding / promo sequence to try to tweak retention and feature usage.

The changes in this PR will centralize logic, but not simplify the number of spots where we check the logic. In a future step (maybe next time we touch this code), we can centralize where the onboarding manager is called. For instance, it would be good to stop passing hasFxa and isOwner up to the AdBanner, and instead just have the page view tell the page header which banner to render, if any. (Probably will need to split the AdBanner into a FirefoxUpsellBanner and an FxaUpsellBanner as part of this.)

A [higher-order component](https://reactjs.org/docs/higher-order-components.html) is another approach we could use, where the rendered page is passed into an HOC that tacks on any promo UI and returns the promo-equipped page, but then we would lose the conceptual simplicity of "the page view assembles the page," where you just read the page view to see what's what. And anyway, an HOC is just as global a component as the onboarding / promo manager, so things aren't better encapsulated; and it would add significant conceptual bulk. Seems like extra complexity without much gain, but we can reconsider it as our requirements change :+1:

punamdahiya:
@6a68 Updated PR with the feedback, trying to figure one last issue where click of dismiss doesn't respond probably because of how react handles getDerivedStateFromProps, looking into that. Thanks

server/src/signin-button.js
```diff
@@ -2,44 +2,64 @@ const React = require("react");
...
     this.state = {
       displaySettings: props.isFxaAuthenticated,
+      showFxaOnboardingDialog: !!props.showFxaOnboardingDialog,
     };
   }
 
   static getDerivedStateFromProps(nextProps, prevState) {
-    return { displaySettings: nextProps.isFxaAuthenticated };
+    return { displaySettings: nextProps.isFxaAuthenticated,
+             showFxaOnboardingDialog: nextProps.showFxaOnboardingDialog};
```
punamdahiya: This is needed to load correct state for myshots where firstRun doesn't load the onboardingDialog state, however this is interfering with click of dismiss logic for home and shot page

server/src/promotion-manager.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+  const hasSeen = localStorage.hasSeenEditButton;
+  if (!hasSeen && enableAnnotations) {
+    localStorage.hasSeenEditButton = "1";
+  }
+  return !hasSeen;
+};
+
+/* Display signin or upsell ad-banner inside header */
+exports.whichAdBanner = function(shouldGetFirefox, isOwner, hasFxa) {
```
6a68: Split this into two methods, `shouldShowFirefoxBanner` and `shouldShowFxaBanner`. Call the methods inline instead of comparing the return value. Making this change will give all methods in this class the same boolean return type and keep the function names more uniform.

server/src/signin-button.js
```diff
@@ -2,44 +2,69 @@ const React = require("react");
...
+    // set counter to max to stop showing promo again
+    localStorage.hasSeenOnboardingDialog = 3;
   }
 };
 
 exports.SignInButton.propTypes = {
   initialPage: PropTypes.string,
   isFxaAuthenticated: PropTypes.bool,
   staticLink: PropTypes.func,
+  showFxaOnboardingDialog: PropTypes.bool,
```
6a68: I'd avoid using `showFoo` to mean the boolean state that decides whether we show Foo, because it seems like `showFoo` would be a show/hide method, not a boolean. This should be `shouldShowFxaOnboardingDialog`. It's long and clumsy, but is the least surprising name for what it does

1
```diff
@@ -0,0 +1,21 @@
...
+# x, exec = run command (the rest of the line) using shell
+# d, drop = remove commit
+#
+# These lines can be re-ordered; they are executed from top to bottom.
+#
+# If you remove a line here THAT COMMIT WILL BE LOST.
+#
+# However, if you remove everything, the rebase will be aborted.
+#
+# Note that empty commits are commented out
```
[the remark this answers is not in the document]
punamdahiya: Oops, removed

server/src/promotion-manager.js
```diff
@@ -0,0 +1,58 @@
+/* Display Firefox Account onboarding dialog in header below signin button */
+exports.shouldShowOnboardingDialog = function(isOwner, isShotPage) {
```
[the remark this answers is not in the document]
punamdahiya: Updated to PromotionStrategy class

server/src/header.js
```diff
@@ -25,4 +31,5 @@ exports.Header.propTypes = {
   isOwner: PropTypes.bool,
   hasFxa: PropTypes.bool,
   shouldGetFirefox: PropTypes.bool,
+  isFxaOnboardingDialog: PropTypes.bool,
```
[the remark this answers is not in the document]
punamdahiya: Updated boolean value to a consistent name hasFxaOnboardingDialog

server/src/pages/homepage/controller.js
```diff
@@ -15,6 +16,7 @@ exports.launch = function(m) {
       });
     }
   }
+  model.fxaOnboardingDialog = shouldShowOnboardingDialog();
```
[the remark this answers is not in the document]
punamdahiya: Updated to hasFxaOnboardingDialog to keep it consistent

server/src/promotion-manager.js
punamdahiya: Done

server/src/signin-button.js
punamdahiya: Updated to hasFxaOnboardingDialog

punamdahiya:
@6a68 Updated PR with feedback, agreed consistent boolean name hasFxaOnboardingDialog is much more easier to follow. 

@flodolo Can you please review localization part of the PR. Thanks!

locales/en-US/server.ftl
```diff
@@ -18,6 +18,17 @@ screenshotsLogo =
...
 
+# Text used in Firefox Account onboarding promo shown below
+# Sign in button in header
+onboardingPromoTitle = What’s new with Firefox Screenshots?
+onboardingPromoMessage = Now, sign in to Screenshots with a Firefox Account and do more:
+onboardingPromoMessageListItem1 = Access your library on all of your devices
+onboardingPromoMessageListItem2 = Store your favorite shots forever
+onboardingPromoDismissButton = Dismiss
+    .title = Dismiss
+onboardingPromoSigninButton = Sign in
```
flodolo: Not a blocker for localization, but should this use Title Case ("Sign In") or Sentence case ("Sign in")? From the rest of the file, title case seems to be used in buttons.

punamdahiya: Thanks for catching it, updated to use title case

server/src/signin-button.js
```diff
@@ -2,44 +2,69 @@ const React = require("react");
...
-        </a>
-      </Localized>;
+    return <div className="fxa-signin">
+        <Localized id="buttonSignIn" attrs={{title: true}}>
+          <a className="transparent nav-button icon-settings" tabIndex="0" href={logInURI} title="SignIn"
+             onClick={this.clickHandler.bind(this)}>
+            <img src={this.props.staticLink("/static/img/icon-settings.svg")} />
+          </a>
+        </Localized>
+        <FxaOnboardingDialog logInURI={logInURI} display={this.state.hasFxaOnboardingDialog}
```
6a68: Remove the `display` prop from the FxaOnboardingDialog and just conditionally render it here

punamdahiya: +1, Updated with feedback. Thanks!

punamdahiya:
As per review feedback, landing in master. Thanks!

PR #5127, opened 2018-11-02
Fixes #5126 - RTL undo/redo icons and shots index

### LINKED ISSUE #5126: Improve RTL support
In #4261 Yehuda added full RTL support to the entire web Screenshots interface, yet there are 2 minors RTL issues remaining to fix:
1. The Redo and Undo button should be mirrored (the arrow pointing to the right should be on the right side and its action should be Undo, and the left should be on the left doing Redo). Current implementation on the dev platform:
default

2. The screenshotted items on the main Screenshots page should be aligned to the right, and their order of appearance should be reversed (the newer shots on the right, the older on the left). Current implementation on the dev platform:
default

yehudab:
Undo/Redo icons switched direction using CSS.
Shots index required passing the correct layout option to the Masonry component.


server/src/pages/shotindex/view.js
```diff
@@ -81,7 +84,7 @@ class Body extends React.Component {
     } else {
```
punamdahiya: nit: we can move masonryOptions definition from line 63 in else block where its used.

punamdahiya:
@yehudab Thanks for fixing #5126, with the nit taken care, PR looks good to merge in master.

server/src/pages/shotindex/view.js
yehudab: Good point. Fixed.

PR #5129, opened 2018-11-02
Update 'async' keyword for sarge 0.1.5

6a68:
Rename keyword arg `async` to `async_`, as mentioned in the sarge 0.1.5 [changelog](https://sarge.readthedocs.io/en/latest/overview.html#change-log).

This gets the tests passing for me locally.


6a68:
Looks like there's some other breakage too

PR #5133, opened 2018-11-02
[WIP] Fix tests?

6a68:
Hopefully the tests will all pass with these two changes.

PR #5147, opened 2018-11-05
Fix Python test regression

ianb:
Due to a update in sarge for Python 3 compatibility, we have to use async_= instead of async=


ianb:
Oops, fixed in #5133

PR #5152, opened 2018-11-08
Fixes #5067 - add custom dimension for login type to GA

### LINKED ISSUE #5067: Add GA custom dimension for browser-only logins
We're going to have logins that are browser-only (with no extension support), and we want to track how those users engage. We already have all the relevant pings, but we'd like to be able to sort those pings between extension users and non-extension users. I propose we add a new custom dimension (`cdN`) that indicates how the user logged in. This will only effect web-based pings. 

From my reading of `ga-activation.js` we don't have great support for globally setting custom dimensions. Maybe the best way is changing ga-activation.js to have an additional flag, that indicates this new custom dimension. This is somewhat similar to how `__USER_ID__` works in that file. Essentially that file is custom built for each user, and this would be another parameter for how it is built.

ianb:


punamdahiya:
@ianb did a basic verification using console logs that [script here](https://github.com/mozilla-services/screenshots/pull/5152/files#diff-5acba341e8aec6ff75f4e5048efb1254R159) is getting a correct login type. I have updated deviceId documentation with accountId details.

Including @6a68 for feedback/comments. Thanks!

docs/METRICS.md
```diff
@@ -83,6 +85,15 @@ Event label: exactly what control invoked the action, such as toolbar-button.  T
 * `selection`: anything that happens during the selection phase, that doesn't happen in the topbar
 * `keyboard`: any keyboard shortcut used
 
+#### Login status custom dimension
+
+The `cd9` dimension holds the "login type". The values can be:
+
+* Empty: visitor is not logged in
```
6a68: LGTM, as long as the dashboard allows filtering on empty values of custom dimensions

punamdahiya: Good catch! Updated to pass explicit value 'non-login' for not logged in case

PR #5153, opened 2018-11-08
Fixes #5146 - Copy string on myshots shows blank

### LINKED ISSUE #5146: The Copy string is not localized in supported languages and on certain cases the string is not displayed at all
[Affected versions]:
Nightly 65.0a1
Screenshots 34.0.0 dev

[Affected Platforms]:
Windows 10

[Prerequisites]:
- Have the latest Screenshots 34.0.0 dev version installed from here.
- Have several selections saved in the profile.
- Have the browser set to Spanish (Spain)

[Steps to reproduce]:
1. Open the Firefox browser with the profile from prerequisites.
2. Navigate to any website and click the "Take a Screenshot" from "Page Actions" menu.
3. Save a selection.
4. On the selection page, click on "All shots" button.
5. In "All shots" page, hover a shot, then click on "Share" button
6. See how "Copy" text is displayed on the button.

[Expected result]:
- The Copy string is localized in supported languages and  is displayed inside the button.

[Actual results]:
- The Copy string is not localized in supported languages and on certain cases the string is not displayed at all.

[Additional Notes]:
screenshots - chrome - copy button text
copy button spanish - english text

punamdahiya:


server/src/share-buttons.js
```diff
@@ -155,10 +155,10 @@ class ShareButtonPanel extends React.Component {
           value={ this.props.shot.viewUrl }
           onClick={ this.onClickInputField.bind(this) }
           onChange={ function() {} /* react gives a warning otherwise */ } />
-        <Localized id={ this.state.copy === "copy" ? "shotPageCopy" : "shotPageCopied" }>
+        <Localized id={ this.state.copy === "copy" ? "shotPageCopyText" : "shotPageCopied" }>
```
flodolo: I'm confused. What was "shotPageCopy" before this PR?

locales/en-US/server.ftl
```diff
@@ -100,6 +100,7 @@ shotPageShareButton =
     .title = Share
 shotPageCopyButton =
     .title = Copy image to clipboard
+shotPageCopyText = Copy
```
flodolo: English is unclear (noun or verb?), I'd prefer a different ID to avoid confusion, e.g. shotPageCopyActionLabel (since `shotPageCopyButton` is already in use)

server/src/share-buttons.js
punamdahiya: @flodolo that was for exact same text "Copy" but got removed in a PR that landed for new copy button on shots. 
Bringing it back with new id as we still need it on my shots page or can we still use the same id in this case?

locales/en-US/server.ftl
punamdahiya: It's a verb, will update with suggested ID

PR #5173, opened 2018-11-13
Fixes #5110 - Favorite shots from myshots page in Edge and Safari

### LINKED ISSUE #5110: [FxA] Can't favorite shots in Safari on My Shots page
This is relevant because it'll probably also break in mobile safari...but it's not a mobile bug, it's a safari bug ;_;

STR:

1. Create a shot in a local copy of screenshots (in firefox)
2. Sign into FxA in local firefox
3. Sign into FxA locally in Safari
4. Go to the my shots page
5. Click the 'favorite' icon

Expected: favorite icon turns red

Actual: 403 error

screen shot 2018-10-30 at 2 08 47 pm

punamdahiya:


punamdahiya:
fetch polyfill is failing in [Edge with error](https://gist.github.com/punamdahiya/1a72d37125759cf86cb9a4838a2f478d) 'Failed to execute 'fetch()' on 'Window': Invalid argument'

Updating toggleFavoriteShot to use XHR to fix setting favorite from myshots page on Edge

6a68:
@punamdahiya I wonder if you're running into [this issue](https://stackoverflow.com/questions/49109878/typeerror-failed-to-execute-fetch-on-window-invalid-value), maybe?

6a68:
Also, the fetch polyfill we're using is for server-side fetch, like fetching oauth credentials. The issue here is related to doing a client-side fetch from edge or old safari, is that right?

punamdahiya:
> @punamdahiya I wonder if you're running into [this issue](https://stackoverflow.com/questions/49109878/typeerror-failed-to-execute-fetch-on-window-invalid-value), maybe?
 
Error says invalid argument instead of invalid value and looks like for this user it's working with POST

punamdahiya:
> Also, the fetch polyfill we're using is for server-side fetch, like fetching oauth credentials. The issue here is related to doing a client-side fetch from edge or old safari, is that right?

It's Edge specific , for Safari I was able to fix by passing credentials: 'include' . And you are correct don't see explicit include of require("node-fetch") in shot index controller

6a68:
> It's Edge specific , for Safari I was able to fix by passing credentials: 'include' .

I wonder if the issue isn't `fetch`, but [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Could you try constructing the query string manually and see if that fixes it?

punamdahiya:
> > It's Edge specific , for Safari I was able to fix by passing credentials: 'include' .
> 
> I wonder if the issue isn't `fetch`, but [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Could you try constructing the query string manually and see if that fixes it?

+1, getting rid of URLSearchParams fixes it for Edge :)

server/src/pages/shotindex/controller.js
```diff
@@ -213,14 +216,6 @@ exports.toggleFavoriteShot = async function(shot) {
...
   } catch (err) {
     console.warn("Error updating expiration:", err);
-
-    // The error could be one from the server or a network level error from
-    // fetch; we display a generic message either way as it's (probably?)
-    // friendlier than percolating a fetch error message up to the user.
-    const errorMessage = document.getElementById("shotIndexAlertErrorFavoriteShot").textContent;
-    if (errorMessage) {
-      window.alert(errorMessage);
-    }
```
punamdahiya: Code above is carried over from shot page and is redundant on myshots

punamdahiya:
@kimberlythegeek TestHomePageAccessibility  is failing for image tag on home page. The images on homepage are background images , I tried to [add role="none"  to make test pass](https://dequeuniversity.com/rules/axe/3.1/image-alt?application=axeAPI) but that didn't help. Can you please recommend the fix to help pass the test. Thanks!

6a68:
> @kimberlythegeek TestHomePageAccessibility is failing for image tag on home page. The images on homepage are background images , I tried to add role="none" to make test pass but that didn't help. Can you please recommend the fix to help pass the test. Thanks!

@punamdahiya I think the fix would be to add alt text to the image. Is it the gear image at the top right corner? If so, you could go with something like "Settings icon"

server/src/pages/shotindex/controller.js
6a68: @punamdahiya I'm not sure this is redundant. It's good to be consistent in how error messages are presented to the user, and the shot page shows a window.alert error if changing expiration fails, so I think it makes sense to also show a window.alert with an error message here.

It does seem that there is a ton of duplication between this controller's `toggleFavoriteShot` and the shot index controller's `changeShotExpiration` (which uses xhr, but seems otherwise pretty much identical). If you have the time and inclination, I think it would be great to extract the shared code.

6a68:
@kimberlythegeek Would it be possible for the axe test to include the id or xpath of the element that threw the error?

punamdahiya:
> > @kimberlythegeek TestHomePageAccessibility is failing for image tag on home page. The images on homepage are background images , I tried to add role="none" to make test pass but that didn't help. Can you please recommend the fix to help pass the test. Thanks!
> 
> @punamdahiya I think the fix would be to add alt text to the image. Is it the gear image at the top right corner? If so, you could go with something like "Settings icon"

Adding alt text to my shots and settings img tags is still failing homepage accessibility tests.

Running home page with Voice over screen reader (Safari) and NVDA( firefox on windows) is reading My shots and Settings button as "visited link, my shots" and "visited link, settings" without alt text on img tags as as we have title attribute set on anchor tags surrounding respective image tags.

server/src/pages/shotindex/controller.js
punamdahiya: 👍 what should be the best place to bring in this common code from controller, a new util class or may be shared/shot

6a68:
@punamdahiya There is an axe accessibility extension that you can run in the page, rather than trying to figure this out via screen reader: https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/

That said, I do think we should just land the fix and fix the tests async

6a68:
I'll try running the axe devtools addon now and see if I can quickly spot the error

6a68:
I'll try retriggering the tests.

With the alt text added to the settings icon, I'm not seeing any alt text errors in the extension:

screen shot 2018-11-14 at 9 57 46 am

punamdahiya:
> @punamdahiya There is an axe accessibility extension that you can run in the page, rather than trying to figure this out via screen reader: https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/
> 
> That said, I do think we should just land the fix and fix the tests async

Thanks for sharing, in dev icon-shots and icon-settings are two places it errors for img alt text

server/src/pages/shotindex/controller.js
6a68: Hmm. I think a separate util is the way to go. shared/shot is shared between server and client, so we have to be super careful about what code goes in there.

server/src/pages/shot/view.js
```diff
@@ -296,9 +296,9 @@ class Body extends React.Component {
     if (typeof expireTime !== "number") {
       expireTime = expireTime.getTime();
     }
-    const deleteTime = new Date(expireTime + this.props.retentionTime);
+
     let restoreWidget;
-    const expirationTimeDiff = <TimeDiff date={deleteTime} />;
+    const expirationTimeDiff = <TimeDiff date={this.props.expireTime} />;
```
punamdahiya: This fixes below error in console for expired pages
error {"msg":"Warning: Failed prop type: Invalid prop `date` of type `date` supplied to `TimeDiff`, expected `number`.\n    in TimeDiff\n    in Body"}

punamdahiya:
@6a68 Updated PR to use shot-utils to toggle favorite on shot and my shots pages. In future we can use shot util to move shared methods such as delete shot. Thanks!

server/src/pages/shotindex/view.js
```diff
@@ -170,6 +170,9 @@ class Body extends React.Component {
         <Localized id="shotIndexPageConfirmShotDelete">
           <div id="shotIndexPageConfirmShotDelete" hidden></div>
         </Localized>
+        <Localized id="shotPageAlertErrorUpdatingExpirationTime">
```
6a68: Let's get rid of this

server/src/pages/shot/view.js
6a68: It seems like you're losing the `expireTime + retentionTime` logic here

server/src/pages/shotindex/view.js
punamdahiya: Quick clarification, this means we are getting rid of error alert on shotindex. Is that correct?

server/src/pages/shot/view.js
punamdahiya: Didn't realise, we show restore button for 14 days on expired shots, will fix

punamdahiya: Need to account for restore logic which isn't toggle favorite but extend expiration date for retention time (14 days)

punamdahiya:
@6a68 Updated PR with review feedback. Updates account for recover logic for expired pages and changes toggleFavorite method name in shot-utils to changeShotExpiration. Thanks!

server/src/lib/shot-utils.js
```diff
@@ -0,0 +1,40 @@
...
+    const url = backend + "/api/set-expiration";
+    if (!newExpiration) {
+      newExpiration = shot.isFavorite ? defaultExpirationMs : 0;
+    }
+    const data = {
+      "id": shot.id,
+      "expiration": newExpiration,
+      "_csrf": csrfToken,
+    };
+    const resp = await fetch(url, {
```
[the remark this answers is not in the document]
punamdahiya: Done, just wrapping await fetch in a try/catch doesn't make lint happy, complains of resp as undefined. Updated to wrap fetch and resp check in try catch block

server/src/pages/shot/controller.js
```diff
@@ -88,32 +89,23 @@ exports.launch = function(data) {
...
+    await shotUtils.changeShotExpiration(shot, model.backend, model.csrfToken,
+                                         model.defaultExpiration, expiration);
+    model.shot.expireTime = model.expireTime = shot.expireTime;
+    if (wasExpired) {
+      location.reload();
+    }
+    render();
+  } catch (err) {
+    const errorMessage = document.getElementById("shotPageAlertErrorUpdatingExpirationTime").textContent;
+    if (errorMessage) {
```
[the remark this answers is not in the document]
punamdahiya: +1, we should check for element instead of message. It's going to error on click of recover as we don't have element there ( that could be a fix in separate issue)

server/src/pages/shot/view.js
```diff
@@ -296,7 +296,9 @@ class Body extends React.Component {
     if (typeof expireTime !== "number") {
       expireTime = expireTime.getTime();
     }
-    const deleteTime = new Date(expireTime + this.props.retentionTime);
+
+    const deleteTime = expireTime + this.props.retentionTime;
+
```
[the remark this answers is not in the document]
punamdahiya: Done

punamdahiya:
@6a68 Updated with review feedback, I am good with you landing after a final pass. Thank you!

PR #5180, opened 2018-11-14
Fixes #5179 - Alt text for homepage img tags

### LINKED ISSUE #5179: Fix home page to pass accessibility tests

punamdahiya:


punamdahiya:
Hold as we need to localize alt texts

punamdahiya:
@6a68 PR fixes accessibility tests on homepage. Thanks!

locales/en-US/server.ftl
```diff
@@ -9,6 +9,13 @@ gNoShots
 gScreenshotsDescription = Screenshots made simple. Take, save, and share screenshots without leaving Firefox.
 
 ## Header
+imgSettings =
+  .alt = Settings Icon
+imgSignIn =
+  .alt = Sign In Icon
```
flodolo: "Sign In" is a verb, "sign-in" is the noun, and I believe you want the latter in this context

johngruen: I think we should have empty alt strings here anyway since we have it on a the parent anchor

punamdahiya: Created https://github.com/mozilla-services/screenshots/pull/5181 to use empty alt text

PR #5186, opened 2018-11-16
Fixes #5184 - My shots fail if accessed with zero shots and no deviceIds

### LINKED ISSUE #5184: [FxA] if you login without ever having uploaded a shot, My Shots fails
For instance, log into https://screenshots.stage.mozaws.net (where you probably haven't uploaded anything) and go to My Shots, and it'll fail. In the logs it will show a parse error (related to #5183)

This is because you've never registered any device with that accountId, and so the query for all related devices returns [].

I'm not sure if we should return [] from `Shot.getShotsForDevice` in this case (which would be reasonable), or if we should raise an exception. We might want different messaging for this case, in which case an exception would let us distinguish it.

punamdahiya:


server/src/servershot.js
```diff
@@ -542,6 +543,9 @@ Shot.getShotsForDevice = function(backend, deviceId, accountId, searchQuery, pag
   }).then(() => {
     const offset = (pageNumber - 1) * SHOTS_PER_PAGE;
     let sql, args, idNums;
+    if (!deviceIds.length) {
+      return [];
+    }
```
punamdahiya: @ianb Issue here is safeguard check is missing before query on Line 567 and 552. All other places where we execute sqls needing data.deviceid IN checks we have this safeguard code

server/src/servershot.js
```diff
@@ -499,10 +499,11 @@ Shot.getShotsForDevice = function(backend, deviceId, accountId, searchQuery, pag
     [deviceId, accountId]
   ).then(rows => {
     deviceIds = rows.map(x => x.id);
+
   }).then(() => {
     if (!deviceIds.length) {
       shotsPage.totalShots = 0;
-      return [{totalShots: 0}];
+      return [{totalshots: 0}];
```
punamdahiya: Updated to bring in synch with property name used in Line 541

punamdahiya:
@ianb Please review. This is an existing issue , probably hidden away well because we never had a use case  where there will be no deviceIds before  Shot.getShotsForDevice call. 
Fixing not create deviceIds until the upload and allowing My shots access with accountIds started code to reach this query where we should have an explicit check for zero deviceIds. Thanks!

PR #5188, opened 2018-11-16
Advise using push -f to release to server-prod, I've made too many mistakes with merging

ianb:
In anticipation I've already turned off the protected branch.

The `ours` merge strategy only makes things worse IMHO, as it lets things proceed dangerously. OTOH, conflict resolution is also always the wrong thing. `push -f` is genuinely the right thing to do.

### DIFF
--- a/docs/server-release.md
+++ b/docs/server-release.md
@@ -9,14 +9,14 @@ To release the server:
 - [ ] Tag with `git tag VERSION`
 - [ ] Push version and tags, `git push && git push --tags`
 - [ ] Merge to `server-prod`:
-  - `git checkout server-prod && git merge -s ours master && git push`
+  - `git push -f origin master:server-prod`
   - This will trigger a deploy to stage. A deploy will take about 30 minutes.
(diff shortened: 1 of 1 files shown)


6a68:
If we can't keep a long-running server-prod branch updated cleanly, why do we have such a branch? Why not just tag master and release from the tag?

ianb:
Tagging would be better, yes, this just documents what works now. We need some scheme to indicate a candidate server tag, and then update the circleci config to trigger the release based on a tag that matches that scheme.

ianb:
It's a little obtuse, but I added instructions to disable then enable branch protection.

hwine:
@ianb Can you describe, or point to, what challenges or errors you ran into with the "ours" strategy? 

If I can't "fix" things, I'd at least like to [warn others](https://wiki.mozilla.org/GitHub/Repository_Security/Problems_and_Options#Problem:_Our_workflow_uses_force_pushes_to_production_to_ensure_we_know_what_will_be_deployed.)

As always, there are tradeoffs -- our main goal is to ensure we get the "best fit" for each project.

PR #5191, opened 2018-11-17
Add missing "=" for gNoShots in en-US server.ftl file

### LINKED ISSUE #5189: Possibly invalid syntax in gNoShots string in locales/en-US/server.ftl file?
Was trying to randomly parse *.ftl files using **fluent-syntax** and was getting a node type of `Junk` which I wasn't expecting...

https://github.com/mozilla-services/screenshots/blob/f61ce20204b521af80b6e87d5ca7841643c73a5f/locales/en-US/server.ftl#L7-L8

Note that the en-CA translation has a trailing `=` after `gNoShots` and doesn't seem to throw an error:

https://github.com/mozilla-services/screenshots/blob/f61ce20204b521af80b6e87d5ca7841643c73a5f/locales/en-CA/server.ftl#L9-L10

pdehaan:
/cc @flodolo @Pike 

I wasn't seeing this as an error/warning on Pontoon or when trying to run <kbd>$ compare-locales --validate l10n.toml .</kbd> locally, but maybe it's a special case because it's in the reference "en-US" locale.

Fixes #5189

### DIFF
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -4,7 +4,7 @@
 ## Global phrases shared across pages, prefixed with 'g'
 gMyShots = My Shots
 gHomeLink = Home
-gNoShots
+gNoShots =
     .alt = No shots found
 gScreenshotsDescription = Screenshots made simple. Take, save, and share screenshots without leaving Firefox.
 


Pike:
Yes, this is one of the caveats of the python parser and it's support for the full evolution of syntaxes.

I've proposed a solution to @stasm in https://github.com/projectfluent/python-fluent/pull/77

PR #5204, opened 2018-11-18
Fix #5150, disable Library button if upload-disabled pref is set

### LINKED ISSUE #5150: Disable Library button if upload-disabled pref is set
See https://bugzilla.mozilla.org/show_bug.cgi?id=1503057

testeaxeax:
Fix #5150

### DIFF
--- a/webextension/experiments/screenshots/api.js
+++ b/webextension/experiments/screenshots/api.js
@@ -18,6 +18,7 @@ const LibraryButton = {
 
   init(webExtension) {
     this._initialized = true;
+    this.visible = !Services.prefs.getBoolPref("extensions.screenshots.upload-disabled", false);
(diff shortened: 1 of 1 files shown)


6a68:
Hey @testeaxeax, thanks for the pull request!

I was planning to use @mythmon's new pref-flipping library to manage this pref for Screenshots, but I'm not sure if it's ready yet. @mythmon, any updates on getting that library into the tree?

mythmon:
That library is currently in review in [bug 1479127](https://bugzilla.mozilla.org/show_bug.cgi?id=1479127). I expect it will be ready to land next week, but that depends on what happens with review. I think it would be safe to base patches to mozilla-central off of the in-review patch, if you're comfortable with that.

PR #5206, opened 2018-11-20
Fixes #5193 - Settings page shows incorrect signin page after login w…

### LINKED ISSUE #5193: [FxA] Settings page show incorrect signin page with a new firefox account
Settings page shows incorrect signin page when logged in with Firefox Account that doesn't have a deviceId associated to it.

In confirm-login, we are saving profile data in account only when [deviceId exists ](https://github.com/mozilla-services/screenshots/blob/master/server/src/server.js#L1182) causing empty accountInfo on settings page

firefox-not-signed-in

punamdahiya:
…ith new firefox account

### DIFF
--- a/server/src/pages/settings/model.js
+++ b/server/src/pages/settings/model.js
@@ -8,7 +8,7 @@ exports.createModel = function(req) {
   };
 
   return db.select(`
-    SELECT accounts.avatarurl, accounts.nickname, accounts.email FROM accounts, devices
+    SELECT accounts.avatarurl, accounts.nickname, accounts.email FROM accounts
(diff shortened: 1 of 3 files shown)


punamdahiya:
For new Firefox accounts, after successful sign in on non-Firefox browsers, click of settings icon shows an incorrect signin page because of null accountinfo

We have a deviceId check in confirm-login route-handler that inserts accountid in accounts table only if deviceid exists, resulting in null accountinfo on settings page.

Fix involves , adding accountid in accounts table after every successful firefox account login. Thanks!

PR #5207, opened 2018-11-20
Fixes #5185 - Access settings page directly from url

### LINKED ISSUE #5185: [FxA] Settings page UI when accessed directly using /settings
Scope of this bug is to discuss /settings page when accessed directly from url in firefox and non-firefox browsers
a) **on non-Firefox browsers**, 
i) if not signed into Firefox Account, [page with message "You must have Screenshots installed"](https://github.com/mozilla-services/screenshots/blob/master/server/src/pages/settings/server.js#L10)
non-firefox-not-signedin

ii) if signed into Firefox Account,
non-firefox-signedin
 

b) **On Firefox browser**

i) If not signed in Firefox Account
firefox-not-signed-in

ii) if signed into Firefox Account
firefox-signedin

punamdahiya:


punamdahiya:
As discussed https://github.com/mozilla-services/screenshots/issues/5185#issuecomment-440391055 , for now if a user access /settings from url without signin, we are redirecting user to homepage.
Fix involves 
a) removing explicit check for deviceId and accountId before showing settings page
b) redirect user to home page if user is not signed into FxA for both firefox and non-firefox browsers
c) Leaving accountInfo null handling by showing below signin page on settings view.js after removing sync messaging.
settings-minimal-signin

test/server/test_responses.py
```diff
@@ -45,7 +45,7 @@ def test_settings_page_requires_auth():
     user = ScreenshotsClient()
 
     resp = requests.get(urljoin(user.backend, "/settings"))
-    assert resp.status_code == 403
+    assert resp.status_code == 200
```
ianb: We should change this test to confirm that the response was redirected (otherwise the test no longer really tests anything, i.e., doesn't test authentication logic). This is a [response object](http://docs.python-requests.org/en/master/api/#requests.Response) and an be checked for `resp.is_redirect` or probably better would be to check `resp.url == user.backend + "/"`

server/src/pages/settings/view.js
```diff
@@ -75,20 +75,12 @@ class Body extends React.Component {
             <Localized id="settingsGuestAccountMessage">
               <p className="title">Guest Account</p>
             </Localized>
-            <Localized id="settingsSignInInvite">
-              <p className="info">Sign in to sync across devices</p>
-            </Localized>
```
ianb: It doesn't seem right to me that we are removing these...? Don't we still want to display these messages to people who aren't logged in via FxA?

punamdahiya: We are redirecting user to homepage when a user doesn't have accountId cookie set. 

This [block of code](https://github.com/mozilla-services/screenshots/blob/55a4a35b79860f7c066882f7199879d1d82a1c32/server/src/pages/settings/view.js#L71) shows only if accountInfo is null that is querying db for the accountInfo doesn't return any record. 
If needed for this edge case, we can keep the strings with word sync removed from messaging.

punamdahiya:
Including @flodolo to review string changes. Thanks!

PR #5226, opened 2018-11-29
Update the cached-path-relative transitive dependency to get rid of an npm audit warning

ianb:
The issue is minor: https://npmjs.com/advisories/739

This should fix the builds


punamdahiya:
@ianb looks safe and good to land thanks!

PR #5240, opened 2018-12-11
Reduce metrics

6a68:
Here's a low-effort way to shut off the GA pings:
- set the server GA blob to always assume do not track is set to 'true', so that the sendEvent calls will always be stubbed out
- make `sendEvent` a no-op on the addon side, leaving the `incrementCount` function in place to actually send the telemetry data

I think this should work. R?

### DIFF
--- a/server/src/ga-activation.js
+++ b/server/src/ga-activation.js
@@ -12,38 +12,8 @@ function _dntEnabled(dnt, ua) {
 
     'use strict';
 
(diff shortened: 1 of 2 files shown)


6a68:
Ah, looks like I need to fix some linter errors around deleted code

6a68:
This was a PR to get some discussion going. Closing for now.

PR #5242, opened 2018-12-11
Allow use of Arrow keys to Cycle between Images

### LINKED ISSUE #4968: Use arrow keys to move between images
[From Discourse:](https://discourse.mozilla.org/t/feature-req-use-arrow-keys-to-move-between-images/32142)

Would be great to be able to use the left/right arrow keys to move between images.

Michael-Overall:
closes #4968 

## Method:
- Retrieve list of URLS to all images from database
- Find the URLS adjacent to the current image being loaded, then retrieve them with a shot's RawValue, and add them to the Shot's props
- Use these props to render clickable "next" and "previous" links in the .../shot/view.js view
- Add an event listener to the view which looks for the "next" and "previous" DOM link objects by their id attribute, then runs their .click() event to cycle between shots

## Rationale:
- Simple links were used instead of loading a carousel of images to compensate for if the user has a ton of images (loading all of the user's shots to a single view would be bad)
- Made the event listener reference links on the page, because it was added to the React component during ComponentDidMount(), so prop values with the actual URLs to redirect to were not available at the time.

## To Test:
- Start up the Screenshots development server (./bin/run-server), and the addon (./bin/run-addon)
- Take some screenshots with the add-on, and hit the "save" button so that they get saved to the local database
- Saving a screenshot to the could will take you to the server's view for that specific shot (if you're not here, choose a shot from the localhost:10080/shots index view). Here, you should see new "previous" and "next" links in this view. Try hitting the left or right arrow keys on your keyboard, and the page should cycle to your next or previous screenshot.


ianb:
Hi Michael. Unfortunately we're shutting down the server this summer, and trying to move people off the server as a result. There was some ongoing internal discussion about this that led to this PR being in limbo for a while. Thank you for your contribution, I'm sorry we were able to include it.

PR #5245, opened 2018-12-18
Fixes #4667 - [a11y] Add alt text to images on shot and myshots page

### LINKED ISSUE #4667: Add alt text to images
Add alt text to images, both on the shot page, and the my shots page.

Really easy solution: "Screenshot image"

Nicer solution: insert text visible in the image, whether via OCR, or via DOM capture at creation time

punamdahiya:


ianb:
Since we're not continuing with the server I think we can ignore this issue until the shutdown.

PR #5272, opened 2019-01-02
Fix #5254 and fix #5253, remove FxA marketing

### LINKED ISSUE #5253: Turn off notification about Settings page
We have a call to action pointing to the settings page. We should remove/disable that.

Code removal is not necessary, we just need to stop it from displaying.

### LINKED ISSUE #5254: Turn off FxA marketing
This include #5253, but also the top bar promotion of FxA.

ianb:
This simply says to never show the promotions (but does not remove the promotion code).


6a68:
Fair enough

PR #5280, opened 2019-01-04
Fix #5277, turn down logging if sendEvent fails

### LINKED ISSUE #5277: Ensure graceful failure for GA if server is gone
We want to be sure if the server is gone, that attempts to send information to GA fails gracefully.

ianb:
This also turns off sending events for the session if it ever gets a 410 Gone response

### DIFF
--- a/webextension/background/analytics.js
+++ b/webextension/background/analytics.js
@@ -7,6 +7,9 @@ this.analytics = (function() {
 
   let telemetryPrefKnown = false;
   let telemetryEnabled;
+  // If we ever get a 410 Gone response from the server, we'll stop trying to send events for the rest
+  // of the session
+  let hasReturnedGone = false;
 
   const EVENT_BATCH_DURATION = 1000; // ms for setTimeout
   let pendingEvents = [];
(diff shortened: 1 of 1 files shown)


webextension/background/analytics.js
```diff
@@ -307,15 +314,17 @@ this.analytics = (function() {
...
+    request.then(response => {
+      if (response.status === 410) { // Gone
+        hasReturnedGone = true;
+        pendingEvents = [];
+      }
+      if (!response.ok) {
+        log.debug(`Error code in event response: ${response.status} ${response.statusText}`);
+      }
+    }).catch(error => {
+      log.debug(`Error event in response: ${error}`);
```
6a68: Some users may not upgrade from 66 to 67 before the server is shut down in May, and `fetch` will throw if the server can't be found, so I think it'd be good to set `hasReturnedGone` inside this `catch`.

ianb: My hope is that we can put in a placeholder site before the site goes down completely. Unfortunately we can't tell the difference between the site being offline, and the computer being offline. I've made a followup in  #5293 to discuss this.

PR #5282, opened 2019-01-04
[Hold] Fix #5235, enable pageAction on reader mode

### LINKED ISSUE #5235: Import upstream reader mode fixes
Directly import https://phabricator.services.mozilla.com/D13908, and maybe also need to update manifest.json to change the match pattern for page action to work

ianb:
This PR will fail, but it's necessary to enable the pageAction startup method. I think the Firefox version on CI doesn't support the `about:reader*` manifest.json action.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -835,7 +835,7 @@
         },
         "string_decoder": {
           "version": "1.1.1",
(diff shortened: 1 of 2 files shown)


6a68:
It's probably some pref we aren't setting inside the profile when running on CI, but yeah, we'll definitely need this for the export.

PR #5294, opened 2019-01-07
Fix #5244, use incognito: spanning in manifest

### LINKED ISSUE #5244: Explicitly opt-in Screenshots for private browsing mode
Mozilla will be defaulting webextensions to disabled in private browsing mode in 66. See the list posting for more details and history: https://mail.mozilla.org/pipermail/firefox-dev/2018-December/006917.html

We already did some work to detect PBM and disable uploads in that case. We just need to twiddle some manifest keys to keep Screenshots working in PBM.

ianb:
Soon Firefox will disable add-ons without this, or require people to opt-in

I believe this follows @mixedpuppy's recommendation

### DIFF
--- a/webextension/manifest.json.template
+++ b/webextension/manifest.json.template
@@ -5,6 +5,7 @@
   "description": "__MSG_addonDescription__",
   "author": "__MSG_addonAuthorsList__",
   "homepage_url": "https://github.com/mozilla-services/screenshots",
+  "incognito": "spanning",
   "applications": {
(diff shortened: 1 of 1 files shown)


mixedpuppy:
> Soon Firefox will disable add-ons without this, or require people to opt-in
> 
> I believe this follows @mixedpuppy's recommendation

It doesn't hurt, but it wasn't necessary.  System addons will continue to run as-is and still have private browsing access.

Issue #5262, opened 2018-12-20
Notify server-using users of server changes


ianb:
We want to have passive notification of the server status for users who have used the server (recently enough that they have unexpired shots). The user status is in #5260 

These users should get links to the #5247 page, somewhere in the interface.

johngruen:
Closing per discussion in stand up 1-10

Issue #5256, opened 2018-12-20
Disable upload


ianb:
We should disable upload for all users in the extension.

At this point we do not need to remove related code.

ianb:
This is done in #5301

adigourdi:
Just noticed that uploading was disabled in latest nightly, but I can't find why this feature got removed

PR #5297, opened 2019-01-09
Fix #5260, query user status from server

### LINKED ISSUE #5260: Query user status from server
We want to know if a user needs notification about the server status. We should query the status (using #5251) and decide if the user:

1. Is using the server, and has shots with indefinite expiration
2. Is using the server, but all shots would normally expire
3. Has not used the server, or all shots are expired

### LINKED ISSUE #5261: Pop open server status tab on indefinite shot users
We'll identify users from #5260. Users who have indefinite shots should get a new tab opened up that goes to the page in #5247 

This should happen once. After the first time, those users will be treated like other server-using users, who get passive notification but not active notification.

### LINKED ISSUE #5263: Remove My Shots links for non-server users
Users who have not used the server (per #5260) should no longer see the My Shots button, or see Screenshots in the Library.

Note if the user somehow manages to save a shot after they use Firefox 66 with no shots, we don't need to worry about them. (This could possibly happen if they change Firefox versions, but seems like an obscure concern.)

This should be enough so that users who haven't used the server will not realize the server existed.

ianb:
Also fix #5263, remove My Shots links for non-server-users

Fix #5261, pop open server page for people with indefinite shots

This was a bit of an annoying patch, as anything involving state, several conditionals, timing, and failure conditions of all the above will be. This doesn't include a fix for #5262, but that will be a fairly simple addition to this change.

### DIFF
--- a/server/src/middleware/csrf.js
+++ b/server/src/middleware/csrf.js
@@ -37,8 +37,8 @@ function csrfHeadersValid(req) {
   const referer = req.headers.referer;
 
   if (isAuthPath(req.path)) {
-    // web ext background scripts don't send headers
-    return origin === undefined && referer === undefined;
+    // web ext background scripts don't send headers, or send a moz-extension origin
+    return (origin === undefined || origin.startsWith("moz-extension:")) && referer === undefined;
   }
 
   return true;
--- a/webextension/background/startBackground.js
+++ b/webextension/background/startBackground.js
@@ -1,4 +1,4 @@
-/* globals browser, main, communication */
+/* globals browser, main, communication, manifest */
 /* This file handles:
      clicks on the WebExtension page action
      browser.contextMenus.onClicked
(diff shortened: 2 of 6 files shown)


webextension/background/startBackground.js
```diff
@@ -10,6 +10,16 @@ const startTime = Date.now();
 
 this.startBackground = (function() {
   const exports = {startTime};
+  // Wait until this many milliseconds to check the server for shots (for the purpose of migration warning):
+  const CHECK_SERVER_TIME = 10000; // 10 seconds
+  // If we want to pop open the tab showing the server status, wait this many milliseconds to open it:
+  const OPEN_SERVER_TAB_TIME = 10 * 60 * 1000; // 10 minutes
```
ianb: #5298 is a followup to determine the best time value.

server/src/middleware/csrf.js
```diff
@@ -37,8 +37,8 @@ function csrfHeadersValid(req) {
   const referer = req.headers.referer;
 
   if (isAuthPath(req.path)) {
-    // web ext background scripts don't send headers
-    return origin === undefined && referer === undefined;
+    // web ext background scripts don't send headers, or send a moz-extension origin
+    return (origin === undefined || origin.startsWith("moz-extension:")) && referer === undefined;
```
ianb: I don't know why this was different from other uses of /api/login (XHR vs fetch?), but for whatever reason I started seeing a moz-extension origin.

ianb:
We'll discuss the popup question in #5298 

I'm landing this despite the test failure (which I can't quite figure out – in *one* situation the autoselection isn't working, even though it's working elsewhere). These tests will go away with #5256, which will happen very soon, so doing lots of work to fix them when they are going to be deleted or changed doesn't seem useful.

PR #5313, opened 2019-01-15
Fix #5299, disable shutdown header

### LINKED ISSUE #5299: Make notification bar of server shutdown configurable
We don't want to push the messaging on everyone yet, so we should make the topbar message configurable, and only turn it on once we have the messaging better in place.

ianb:
I wanted to make this configurable, but passing that information via the backend, into React, into the client during hydration, started to touch lots of files. Hardcoding it keeps it simple.


punamdahiya:
LGTM, thanks!

PR #5316, opened 2019-01-16
Fix #5298, change export-needed tab opening notification time to 10 seconds

### LINKED ISSUE #5298: Change tab opening timing
In #5297 there's values for how long after startup to check the server (which doesn't matter too much), and how long after that to open the tab. I've put in 10 minutes, as I *think* that's better than just opening it immediately. But that's an arbitrary number, and should be reviewed.

ianb:


ianb:
I think the new tab should be focused... I suppose there's opportunities for something else to take the focus, but in those same cases I suspect that explicitly focusing the new tab could also be overloaded.

For SUMO, I think we just have to do our best to get all this updated – we have too many intersecting pieces to get them all updated at exactly the same time.

PR #5324, opened 2019-01-17
Add reviewed language for the site banner

ianb:
(I believe it's the right thing to do this as two strings to account for the `<b>`, but I'm not 100% sure)

### DIFF
--- a/locales/en-US/server.ftl
+++ b/locales/en-US/server.ftl
@@ -17,9 +17,7 @@ screenshotsLogo =
     .title = Screenshots Home
 bannerSignIn = <a>Sign in or sign up</a> to access your shots across devices and save your favorites forever.
 bannerUpsell = {gScreenshotsDescription} <a>Get Firefox now</a>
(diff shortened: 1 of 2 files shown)


locales/en-US/server.ftl
```diff
@@ -17,6 +17,9 @@ screenshotsLogo =
     .title = Screenshots Home
 bannerSignIn = <a>Sign in or sign up</a> to access your shots across devices and save your favorites forever.
 bannerUpsell = {gScreenshotsDescription} <a>Get Firefox now</a>
+# These two messages are presented in a banner on all pages, like: <b>shutdownWarningIntro</b> shutdownWarningBody
+shutdownWarningIntro = Saved screenshots are expiring soon.
```
flodolo: Is there a reason for this not being a single string? Technically it should work.

```
shutdownWarningIntro = 
    <b>Saved screenshots are expiring soon.</b> Starting in June, Screenshots
    will no longer offer online storage. Want to keep shots from your library?
    <a>Download them to your computer.</a>
```

PR #5346, opened 2019-01-29
Fix #5332, cache startup server check

### LINKED ISSUE #5332: Cache startup server check
Once we get a positive response from the server we don't need to check again, we can cache indefinitely

ianb:
This applies to the migration code that runs on startup

### DIFF
--- a/webextension/background/startBackground.js
+++ b/webextension/background/startBackground.js
@@ -122,12 +122,17 @@ this.startBackground = (function() {
         break;
       }
     }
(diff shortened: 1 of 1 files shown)


6a68:
@ianb not sure what's up with the tests. feel free to self-merge

ianb:
Something is going on with CI, as all tests are failing for other non-addon PRs too. I guess I'll just go ahead...

PR #5353, opened 2019-02-08
Fix #5241, don't ship files in test/ with Firefox

### LINKED ISSUE #5241: Screenshots is shipping browser test files as part of the Firefox build
In the generated section of screenshot's moz.build, various test files are being set as FINAL_TARGET_FILES. This means they're being including in the xpi that is shipped with Firefox itself.

These shouldn't be necessary as the build system automatically bundles up test files separately to Firefox itself.

Additionally, the .eslintrc.yml isn't actually being shipped, I suspect because it is a dot file the build system is ignoring it somehow.

https://searchfox.org/mozilla-central/rev/adcc169dcf58c2e45ba65c4ed5661d666fc3ac74/browser/extensions/screenshots/moz.build#463-467

```
FINAL_TARGET_FILES.features['screenshots@mozilla.org']["test"]["browser"] += [
  'test/browser/.eslintrc.yml',
  'test/browser/browser.ini',
  'test/browser/browser_screenshots_ui_check.js'
]
```

ianb:
Also changes the base path from ./webextension->webextension, to be ./->, which makes the copying work; I'm not sure why this was necessary or why it worked before, but it made it work this time.

### DIFF
--- a/bin/update_mozbuild.py
+++ b/bin/update_mozbuild.py
@@ -6,20 +6,22 @@ import io
 import re
 
 skipFiles = [
-    "manifest.json.template"
+    "manifest.json.template",
+    "test",
 ]
 
 
 def getFullFileList(outputLoc, dirName):
     result = {dirName: []}
     for entry in os.listdir(outputLoc):
+        if entry in skipFiles:
+            continue
         if os.path.isdir(os.path.join(outputLoc, entry)):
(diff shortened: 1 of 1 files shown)


ianb:
I made a followup in #5357 – it's slightly confusing because we have two config files

PR #5386, opened 2019-07-10
Rename bn-BD as bn, remove bn-IN

flodolo:
We're unifying bn-IN and bn-BD under a single locale (bn). Once this is merged, we'll move the project in Pontoon.
mozilla-l10n/pm-projects#169

Note that bn-IN files contain only comments.


flodolo:
I'd go ahead and merge, but I'm not sure if someone is actually still looking at this repository at this point.

6a68:
@flodolo hey, you're fine to land this. it looks like the tests failed due to the known intermittent issue where webdriver doesn't connect to the remote browser correctly. I restarted them, you're fine to wait or just land this patch.

I'm not sure when the next export to firefox / deploy will happen, but we can discuss on slack if there are time constraints around making this change in tree.

6a68:
@flodolo Now that I think about it, if you need this to land in m-c any time soon, the best approach would be to open a separate phabricator request against the code in tree. fxa work is keeping me and @ianb very busy, and doing a screenshots export from github into m-c is unlikely to happen any time soon. sorry :-\

flodolo:
> Now that I think about it, if you need this to land in m-c any time soon, the best approach would be to open a separate phabricator request against the code in tree

Is the procedure documented somewhere?

PR #5395, opened 2019-10-01
Static homepage

ianb:
This makes the homepage static, and removes dynamic endpoints from the server.

It doesn't remove code, just stubs out the entry points into all that code.


ianb:
And of course tests fail because there's no longer a server...

